PR #50, opened 2022-12-26
feat: Search hooks

JoaquinBattilana:
API Documentation: https://docs.lens.xyz/docs/search-profiles-and-publications

Adds `useSearchProfiles` and `useSearchPublications` hooks with examples and tests


examples/web-wagmi/src/components/Examples.tsx
```diff
@@ -1,49 +0,0 @@
-import { Link } from 'react-router-dom';
```
JoaquinBattilana, line 1: This file was not being used

krzysu:
code looks good, but when testing using `web-wagmi` example app the results are different than on Lenster, is that expected? ie. search for "web4" profile

regarding searching for Publications, the results are also off, i.e. search for "web3" publications using the example app, I'm getting multiple post from "@0x5181.burned" profile which looks like some spam bot.
Also there is a web dev console error: 
> Warning: Encountered two children with the same key, `0x5181-0x15`

Is that ok or shall we make sure that search results are of good quality, at least the same as on Lenster?

krzysu:
one more thing, but that might not be related to this PR:

the `hasMore` flag of `useInfiniteScroll` seems not to be correct when there are no more items, which results in the text "Loading more..." never disappearing.

Screenshot 2022-12-27 at 16 00 36

JoaquinBattilana:
> the `hasMore` flag of `useInfiniteScroll` seems not to be correct when there are no more items, which results in the text "Loading more..." never disappearing.

Yes, I found that too, we're checking that if `!cursor` then `hasMore=false` and it seems that sometimes in the API response `cursor` is defined even if there're not more items. We should check if this is an API bug or is expected, then we can change the `hasMore` flag to check against `totalITems` (maybe?)

JoaquinBattilana:
> code looks good, but when testing using `web-wagmi` example app the results are different than on Lenster, is that expected? ie. search for "web4" profile
> 
> regarding searching for Publications, the results are also off, i.e. search for "web3" publications using the example app, I'm getting multiple post from "@0x5181.burned" profile which looks like some spam bot. Also there is a web dev console error:
> 
> > Warning: Encountered two children with the same key, `0x5181-0x15`
> 
> Is that ok or shall we make sure that search results are of good quality, at least the same as on Lenster?

I'll check this, probably some difference in the query parameters, i can set up the default parameters to be the same as Lenster (do we want that?)

joshstevens19:
Paging issue has been reported before if you find the exact query please raise full case on > https://github.com/lens-protocol/api-examples/issues and we fix it

cesarenaldi:
re: default parameters

@JoaquinBattilana can you clarify what parameters? We need to draw a line between providing the best baseline UX to all apps using Lens SDK and specific social media take on search experience for their audience/needs.

Maybe if you can identify the difference between our code and Lenster we can understand what is "baseline good UX params" and what is Lenster specific.

krzysu:
I created tickets for issues with pagination for search queries
https://github.com/lens-protocol/api-examples/issues/114
https://github.com/lens-protocol/api-examples/issues/115

cesarenaldi:
@JoaquinBattilana can you please update the [`TODO.md`](https://github.com/lens-protocol/lens-sdk/blob/main/TODO.md) as part of this PR?

cesarenaldi:
Can we please include changeset config to this PR?

JoaquinBattilana:
> code looks good, but when testing using `web-wagmi` example app the results are different than on Lenster, is that expected? ie. search for "web4" profile
> 
> regarding searching for Publications, the results are also off, i.e. search for "web3" publications using the example app, I'm getting multiple post from "@0x5181.burned" profile which looks like some spam bot. Also there is a web dev console error:
> 
> > Warning: Encountered two children with the same key, `0x5181-0x15`
> 
> Is that ok or shall we make sure that search results are of good quality, at least the same as on Lenster?

For the ids problem i created a ticket in the API repo
https://github.com/lens-protocol/api-examples/issues/116

The difference with lenster is that they don't use `sources` in their search query and they also use `customFilters` (which i added on the query parameters on the last push)

packages/react/src/publication/useSearchPublications.ts
```diff
@@ -0,0 +1,39 @@
...
+  limit?: number;
+  observerId?: string;
+  customFilters?: Array<CustomFiltersTypes>;
+}>;
+
+export function useSearchPublications({
+  query,
+  limit = DEFAULT_PAGINATED_QUERY_LIMIT,
+  observerId,
+  customFilters,
```
cesarenaldi, line 23: What's the implication of exposing `customFilters`? If we know them and it's a finite set let's formalize them with clear types, if not, better hold for now until we clarify what use cases we need to support.

Let's make sure we don't expose too low level details of the API which would be detrimental to the purpose of this SDK (simplify interactions with the Lens Protocol).

packages/api-bindings/src/graphql/search.graphql
```diff
@@ -0,0 +1,63 @@
...
+) {
+  result: search(
+    request: {
+      query: $query
+      type: PUBLICATION
+      limit: $limit
+      cursor: $cursor
+      sources: $sources
+      customFilters: $customFilters
+    }
```
cesarenaldi, line 17: Let's reduce the amount of translations by exposing just the `request` object to our hook code.

JoaquinBattilana: I think this is not possible here without some hacky solution, if we just expose `request` then we need to make a generic `search` query for both `SearchPublications` and  `SearchProfiles`, because we can't set up `type: PUBLICATION`  or `type:PROFILE` directly in the query:

```graphql
query SearchPublications(
  $request: SearchRequestQuery
  $observerId: ProfileId
) {
  result: search( request: $request) {
    ... on PublicationSearchResult {
      __typename
      items {
        ... on Post {
          ...Post
        }

        ... on Comment {
          ...Comment
        }
      }
      pageInfo {
        ...CommonPaginatedResultInfo
      }
    }
  }
  
      ... on ProfileSearchResult {
      __typename
      items {
        ...ProfileFields
      }
      pageInfo {
        ...CommonPaginatedResultInfo
      }
    }
}
```

and as we know this throws an error because the type from `id` is different in both `Profile` and `Publication`
I tried to find some way to do a spread or setup a default parameter in a graphql query but i found nothing, do you have any suggestion?

JoaquinBattilana: Well, we can probably set the `type` on the react hook, but then the query defined in `search.graphql` as `SearchPublications` will have no sense outside of the react hook context

packages/react/src/publication/useSearchPublications.ts
JoaquinBattilana: At the moment `customFilters` is an enum (this may or not change in the future) so we know the values and they're defined in the generated codegen file. With this PR we are exporting it in the `react` package.
For now users will be able to use the only defined filter `CustomFiltersTypes.Gardeners` which filter some spam. I think this one is pretty useful for any ui to delete spam from their explore/search pages.

JoaquinBattilana: I opted for deleting the `customFilters` in my last commit, i don't want to stale this PR anymore, we can discuss in the future if it's worth adding.

examples/web-wagmi/src/components/Breadcrumbs.tsx
```diff
@@ -10,20 +11,18 @@ export function Breadcrumbs() {
 
   return (
     <div style={{ display: 'flex', flexDirection: 'row', gap: '.5rem' }}>
-      <Link to="/">
-        <a>Home</a>
-      </Link>
```
JoaquinBattilana, line 15: this is because a <a>  inside <a> console error

examples/web-wagmi/src/components/Breadcrumbs.tsx
```diff
@@ -10,20 +11,18 @@ export function Breadcrumbs() {
...
+      <Link to="/">Home</Link>
 
       {paths.map((path, index) => {
         const to = `/${paths.slice(0, index + 1).join('/')}`;
 
         return (
-          <>
+          <React.Fragment key={to}>
```
JoaquinBattilana, line 20: I think this is the only way to set a key to a fragment

krzysu (approved):
👍

PR #51, opened 2022-12-27
Bugfix/NextJS support

cesarenaldi:
There was a misconfiguration with the `@lens-protocol/react` `exports` which was causing some bundler to not find the `@lens-protocol/react/web` entry point. NextJS SWC configuration was one of these.

This PR fixes the issue and adds a NextJS example app which will be included into the CI build step preventing similar issues to happen in the future.


packages/react/package.json
```diff
@@ -21,12 +21,12 @@
...
-        "types": "./dist/esm/web.d.ts",
-        "default": "./dist/esm/web.js"
+        "types": "./dist/esm/web/index.d.ts",
+        "default": "./dist/esm/web/index.js"
       },
       "require": {
-        "types": "./dist/cjs/web.d.ts",
-        "default": "./dist/cjs/web.js"
+        "types": "./dist/cjs/web/index.d.ts",
+        "default": "./dist/cjs/web/index.js"
```
krzysu, line 29: so this is the actual fix only?

krzysu:
we don't need files from `examples/nextjs/public` folder, the rest LGTM!

krzysu (approved):

packages/react/package.json
cesarenaldi: Fixed here: https://github.com/lens-protocol/lens-sdk/pull/51/commits/42307333ca23343f5dd642871a1c12c5f3cc1d8f

cesarenaldi:
> we don't need files from `examples/nextjs/public` folder, the rest LGTM!

Fixed here: https://github.com/lens-protocol/lens-sdk/pull/51/commits/42307333ca23343f5dd642871a1c12c5f3cc1d8f

PR #52, opened 2022-12-28
feat: useExplorePublicationsHook

JoaquinBattilana:
Added:

- `ExplorePublications` graphql query and mocks
- `useExplorePublications` react hook based on `ExplorePublications` query
- `useExplorePublications` basic test and example
- Created `LinkPublicationCard` to use across the examples


krzysu:
code looks good, but there is a pagination issue with `ExplorePublications` query, I investigated and created this issue in the API repository https://github.com/lens-protocol/api-examples/issues/113

krzysu (approved):
code LGTM but it doesn't really work because of API pagination issue, it stops loading more after 2 or 3 pages. This shouldn't really ever happen for exploration that we run out of publications.

packages/react/src/utils.ts
```diff
@@ -15,3 +16,5 @@ export function assertRequiredSigner(signer: Signer): asserts signer is Required
 }
 
 export const DEFAULT_PAGINATED_QUERY_LIMIT = 10;
+
+export const DEFAULT_PUBLICATION_SORT_CRITERIA = PublicationSortCriteria.Latest;
```
cesarenaldi, line 20: Is this used in more than one place or planned to?

cesarenaldi: What is the aim with these consts? Are these exported by the package at all?

packages/react/src/publication/useExplorePublications.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export function useExplorePublications({
+  observerId,
+  limit = DEFAULT_PAGINATED_QUERY_LIMIT,
+  sortCriteria = DEFAULT_PUBLICATION_SORT_CRITERIA,
+  timestamp,
+  noRandomize,
+  publicationTypes,
+  excludeProfileIds,
+}: UseExplorePublicationsArgs = {}): PaginatedReadResult<Array<Publication>> {
```
cesarenaldi, line 29: Can we please clarify what is the mindset with making these specific set of parameters available?

Not sure some of these are intelligible (e.g. `noRandomize`) as they are.

joshstevens19:
@krzysu FYI explore will only show the last week worth of content. On testnet this can come to the end fast! On polygon ofc it should rarely happen but it can happen!

packages/react/src/publication/__tests__/useExplorePublications.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+import { renderHookWithMocks } from '../../__helpers__/testing-library';
+import { DEFAULT_PAGINATED_QUERY_LIMIT, DEFAULT_PUBLICATION_SORT_CRITERIA } from '../../utils';
+import { Publication } from '../types';
+import { useExplorePublications } from '../useExplorePublications';
+
+describe(`Given the ${useExplorePublications.name} hook`, () => {
+  const mockPublications: Publication[] = [mockPost()];
+
+  describe('when the query returns data successfully', () => {
+    it('it should return publications that match the explore default parameters', async () => {
```
cesarenaldi, line 17:
```suggestion
    it('should return publications that match the explore default parameters', async () => {
```

packages/react/src/publication/useExplorePublications.ts
```diff
@@ -0,0 +1,47 @@
...
+  timestamp?: number;
+  noRandomize?: boolean;
+  publicationTypes?: Array<PublicationTypes>;
+  excludeProfileIds?: Array<string>;
+}>;
+
+export function useExplorePublications({
+  observerId,
+  limit = DEFAULT_PAGINATED_QUERY_LIMIT,
+  sortCriteria = DEFAULT_PUBLICATION_SORT_CRITERIA,
```
cesarenaldi, line 24: Let's inline `PublicationSortCriteria.Latest` for now. When we will need to share this we can then think about a proper const. The unit test seems to not explore with/without default params... not that I am advocating it. Maybe for the unit test you can use a value for `PublicationSortCriteria` that is not `PublicationSortCriteria.Latest` and by configuring the query mock it will implicitly check it passes to the query variables.

packages/react/src/publication/index.ts
```diff
@@ -6,12 +6,21 @@ import {
...
 
 export type { Publication } from './types';
 export type { PostFragment, CommentFragment, MirrorFragment, CommentWithFirstCommentFragment };
-export { isPostPublication, isCommentPublication, isMirrorPublication };
+export {
+  isPostPublication,
+  isCommentPublication,
+  isMirrorPublication,
+  PublicationSortCriteria,
+  PublicationTypes,
```
cesarenaldi, line 25: Cool, I was looking for these exports.

packages/api-bindings/src/graphql/publications.graphql
```diff
@@ -48,3 +48,45 @@ query Publications(
...
+  result: explorePublications(
+    request: {
+      limit: $limit
+      cursor: $cursor
+      timestamp: $timestamp
+      sortCriteria: $sortCriteria
+      sources: $sources
+      publicationTypes: $publicationTypes
+      noRandomize: $noRandomize
+      excludeProfileIds: $excludeProfileIds
```
cesarenaldi, line 72: To minimize TS type def and GQL definitions we can simply pass the entire request object and have the wrapping hook be the sole place of indirection between outer facing parameters and GQL variables.

examples/web-wagmi/src/components/feed/Feed.tsx
```diff
@@ -44,7 +25,7 @@ export function Feed() {
       {infiniteScroll.data
         .filter((i) => isPostPublication(i.root))
         .map((item, i) => (
-          <Publication key={`${item.root.id}-${i}`} feedItem={item} />
+          <LinkedPublicationCard key={`${item.root.id}-${i}`} publication={item.root} />
```
cesarenaldi, line 28: Let's use a non-link card here. @reecejohnson is reworking these examples to be more of a showcase of all the hooks rather than browseable app. I know I advocated for making these links, but it turns out it does not worth the effort and introduces issues with nested `<a>` which we don't need to deal with.

cesarenaldi:
Can we please include changeset config to this PR?

cesarenaldi:
Can we please update `TODO.md` as well?

packages/react/src/utils.ts
JoaquinBattilana: To be honest is not worth it because the `sortCriteria` is only used in `ExplorePublications` query, i'll move it to the inline version that you suggested in the other comment

packages/react/src/publication/useExplorePublications.ts
JoaquinBattilana: I added a test with custom parameters

packages/react/src/publication/useExplorePublications.ts
JoaquinBattilana: I feel like we should give the user the ability to choose if they want to use a parameter or not, even if it's an obscure one. Putting that matter aside i think `noRandomize` and `customFilters` (added in my last commits) are the only ones that are not self-explanatory. `customFilters` is being used in some big UIs like Lenster so I think it should be added.

packages/react/src/index.tsx
```diff
@@ -33,3 +33,4 @@ export type {
   FiatAmount,
   Matic,
 } from '@lens-protocol/shared-kernel';
+export { CustomFiltersTypes } from '@lens-protocol/api-bindings';
```
JoaquinBattilana, line 36: Does this belongs here? It's used in both publication and profile so it doesn't make sense to be inside one of those folders.

packages/react/src/publication/useExplorePublications.ts
JoaquinBattilana: I opted for deleting those parameters, we can think in the future if it's worth adding

PR #54, opened 2022-12-28
feat: improve `web-wagmi` structure, navigation and styling 

reecejohnson:
- Adds better navigation to `web-wagmi` with header and base pages per section
- Improves basic styling for `web-wagmi`
- Adds missing hook examples


krzysu (approved):
🔥

PR #56, opened 2022-12-29
feat: implement `useReaction` hook (T-294)

krzysu:
Add `useReaction` hook with example and tests

Documentation https://docs.lens.xyz/docs/tricky-cherries-happen-usereaction


packages/react/src/publication/adapters/ReactionGateway.ts
```diff
@@ -0,0 +1,94 @@
...
+          publicationId,
+          profileId,
+          reaction: reactionTypeMap[reactionType],
+        },
+      });
+
+      if (!data) {
+        return failure(
+          new NetworkError(
+            new Error(`Failed to add reaction to a publication with id "${publicationId}"`),
```
krzysu, line 48: @cesarenaldi what would be the best way to handle errors here? I see that `NetworkError` is deprecated, but as we are very specific with error types across the codebase, what would you suggest to return here?
Possible errors are:
- server/network connection
- add/remove reaction which already exists
- wrong variables
- ??

cesarenaldi: With hindsight `NetworkError` is not a domain failure mode.
It's a generic wrapper for unexpected errors that might happen during any operation.

The decision we made when adopting a code style that make use of `Result<T, E>` (which is getting widely adopted as primitive in many languages like Kotlin with his parametrized enums, Swift with it's equivalent of `Result`, but also Rust) is to make a clear distinction between "known failure modes" (i.e. normal way of operating that leads to a not successful result) and actual unexpected failures (which we usually refer to as traditional exceptions).

Known failure modes are piped back into the program's control flow via the `Result<T, E>` primitive (`Failure<T, E>` to be accurate). These becomes explicit and forces consumers of a given function/method to be explicit in the way these are handled.

Exceptions are instead thrown and bubbles up the call stack "as usual".

When using a 3rd party lib or a system API it's not uncommon to have to catch low level exception which are thrown, detect if a failure mode has been encountered and return a `Failure<T, E>` with a domain modelled error type.

To me, Server/Network connection errors for the class of problems we are solving falls into the "exception categories". Throw them, let them bubble up the call stack. In a React app the  call stack might lead to a render cycle method and they can be handled with a traditional React Error Boundary (*)

Add/remove reaction if can be programmatically cause in a normal way of operating seems to fall into known failure modes (so `failure` seems appropriate). If instead they can only be caused by a programming mistake I would consider a cheaper invariant error.

Wrong variables without more context seems to fall into a coding mistake... can it be done in a normal operating situation?

I guess IMO there is no one size fits all. It depends on the problem at hand and what are the actual failure modes (if there are any).

(*) we should/could consider to add an Error Boundary in the `LensProvider` at some point as catch-all in case the developer didn't manage they own application error boundaries.

krzysu: thank you for the explanation. I found out that in the case of `useReaction` hook, we have a presenter that has to know about possible failures of the API call to revert the optimistic reaction. that forces us to return a possible error in the `error` field. or shall we throw from the hook if an error is returned from the gateway?

right now it looks like here, but according to what you wrote, that is a coding mistake to allow to submit the same reaction multiple times.
Screenshot 2023-01-02 at 15 37 58

packages/react/src/publication/adapters/useReactionController.ts
```diff
@@ -0,0 +1,26 @@
...
+import { useSharedDependencies } from '../../shared';
+import { ReactionGateway } from './ReactionGateway';
+import { ReactionPresenter } from './ReactionPresenter';
+
+export function useReactionController() {
+  const { apolloClient } = useSharedDependencies();
+
+  const presenter = new ReactionPresenter(apolloClient.cache);
+  const gateway = new ReactionGateway(apolloClient);
+  const reaction = new Reaction(gateway, presenter);
```
cesarenaldi, line 12: This approach creates new adapters at every render cycle which by React design can happen any time.

Although it's not wrong per-se implies that the controller knows that these adapters are probably stateless or the only state they care about is the dependencies provided.

I feel this approach could be error prone as someone might make a change to one of these adapters, for example adding an internal state (think of an in-memory cache for a gateway, or a reference to some UI state for a presenter) and because such state gets replaced with a new one at every re-render they might introduce a malfunction or even a memory leak depending if the such state is able to keep a reference to the adapters n-th instance.

I am not advocating for one over the other but 2 different approaches might be better:
- memoize the creation of these adapters (this could also inform the reader if their state is tied to the hook lifecycle)
- create instances of these adapters as part of the `add` and `remove` methods. This could be helpful to communicate to the reader the presence of an ephemeral state that could be tied to the operation lifecycle (add or remove a reaction).

This last option could be helpful to address your very valid comment here: https://github.com/lens-protocol/lens-sdk/pull/56#discussion_r1060072482

Maybe we can bounce some ideas on this topic.

krzysu: I understand it now, my bad. I saw the 2nd approach in other Controllers (create instances of these adapters as part of the add and remove methods) so I will just do the same.

package.json
```diff
@@ -11,7 +11,8 @@
     "test": "turbo run test --no-cache",
     "dev": "turbo run dev --parallel",
     "lint": "turbo run lint --parallel",
-    "lint:fix": "turbo run lint:fix --parallel"
+    "lint:fix": "turbo run lint:fix --parallel",
+    "changeset": "changeset"
```
desfero, line 15: wondering if we need that as far as i know then your run `pnpm <something>` it also looks into the `node_modules/bin` folder.
https://pnpm.io/using-changesets

krzysu: you are right! I will remove this one, but leave it in the Readme

examples/web-wagmi/src/publications/UseReaction.tsx
```diff
@@ -0,0 +1,98 @@
...
+    profileId,
+  });
+
+  const hasReactionType = hasReaction({
+    reactionType,
+    publication,
+  });
+
+  const toggleReaction = async () => {
+    if (!publication) return;
```
reecejohnson, line 31: nit: this check is redundant, profile is always defined

reecejohnson (approved):
Looks good mate! 👏 Left one small comment

PR #57, opened 2023-01-02
feat: add `useCreateComment` 

desfero:
- Adds `useCreateComment` hook inspired by the way `useCreatePost` works
- Adds follow UI example in `web-wagmi`
- Adds new docs for review https://dash.readme.com/project/lensprotocol/v1.1/docs/use-create-comment


examples/web-wagmi/src/components/Breadcrumbs.tsx
```diff
@@ -1,3 +1,4 @@
+import { Fragment } from 'React';
```
desfero, line 1: small fixes in this file to avoid having console errors

examples/web-wagmi/src/discovery/UseFeed.tsx
```diff
@@ -33,23 +34,26 @@ export function Feed() {
...
-      {infiniteScroll.data
-        .filter((i) => isPostPublication(i.root))
-        .map((item, i) => (
-          <Publication key={`${item.root.id}-${i}`} feedItem={item} />
-        ))}
-
-      {infiniteScroll.hasMore && <p ref={infiniteScroll.observeRef}>Loading more...</p>}
+      <WhenLoggedIn>{({ profile }) => <CreatePost activeProfile={profile} />}</WhenLoggedIn>
+
+      {infiniteScroll.data.length > 0 ? (
```
desfero, line 45: This change would allow to create a new post when the user has an empty feed.

packages/react/src/transactions/adapters/PublicationCallGateway.ts
```diff
@@ -237,7 +325,21 @@ export class PublicationCallGateway implements ICreatePostCallGateway {
...
+      contentURI: await this.uploadPublicationMetadata(request),
+      collectModule: resolveCollectModule(request),
+      referenceModule: resolveReferenceModule(request),
+    };
+  }
+
+  private async uploadPublicationMetadata(
+    request: CreatePostRequest | CreateCommentRequest,
+  ): Promise<string> {
     return this.upload(createPublicationMetadata(request));
```
desfero, line 343: Should we model some known error when the upload failed? 
Given that `upload` is externally provided by the user it may throw some exceptions that would be good to forward back to the `useCreatePost/Comment` hook in place of `error` so the user may decide to handle those in a way we can't foresee.
Any thoughts?

packages/react/src/transactions/adapters/PublicationCallGateway.ts
```diff
@@ -226,6 +290,30 @@ export class PublicationCallGateway implements ICreatePostCallGateway {
     );
   }
 
+  private async createCommentCall<T extends CreateCommentRequest>(
```
cesarenaldi, line 293: I know in the original codebase I modelled all these as methods of the same class... I think I violated Open/Closed Principle in doing so and the code seems complicated. Is it thinkable to create an abstract class that a Post Gateway and a Comment Gateway extends by filling the gaps? Will it be easier to manage? Also "create mirror" beyond creating a Lens Publication does have little overlap with the settings of Post and Comment.

Thoughts?

packages/react/src/transactions/adapters/PublicationCallGateway.ts
cesarenaldi: I've done some work here: https://github.com/lens-protocol/lens-sdk/pull/61

Worth taking into account my other comment: https://github.com/lens-protocol/lens-sdk/pull/57#discussion_r1060600649

There is an opportunity to reduce the complexity of this code. Maybe with a generic or an abstract class?

packages/react/src/transactions/adapters/__tests__/PublicationCallGateway.spec.ts
```diff
@@ -82,6 +90,42 @@ function mockCreatePostViaDispatcherMutationMockedResponse({
...
+      variables,
+    },
+    result: {
+      data,
+    },
+  };
+}
+
+function mockCreateCommentViaDispatcherMutationMockedResponse({
+  variables,
```
krzysu, line 112: these two mock functions, is it fine to keep them here or they should be places in `@lens-protocol/api-bindings/mocks` as all other mocks?

krzysu (approved):
I wasn't able to test it in web-wagmi app as I don't have Polygon Matic in my test wallet, but the rest looks good 👍

packages/react/src/transactions/adapters/PublicationCallGateway.ts
desfero: Seems there were no use cases for an abstract class as post/comments are both following separate code paths. 
They share some common utils (which are already separate functions) and upload handler logic (which for now is very simple, once #58 is merged we would need to create a new `UploadHandler` class that handles catching and rehrowing upload errors) so I refactored them as 2 separate classes.

cesarenaldi:
> I wasn't able to test it in web-wagmi app as I don't have Polygon Matic in my test wallet, but the rest looks good 👍

@krzysu in general, did you try the instructions in the alert? it suggests how to get Mumbai MATIC...
then the example upload function should guide you to fund your Bunldr wallet to upload to Aerwave.

packages/react/src/transactions/adapters/PublicationCallGateway.ts
cesarenaldi: Composition over inheritance feels a better choice, I totally agree.

krzysu:
> > I wasn't able to test it in web-wagmi app as I don't have Polygon Matic in my test wallet, but the rest looks good 👍
> 
> @krzysu in general, did you try the instructions in the alert? it suggests how to get Mumbai MATIC... then the example upload function should guide you to fund your Bunldr wallet to upload to Aerwave.

oh, I was sure I needed Polygon MATIC to upload to aerwave. I have Mumbai MATIC and I've just checked that creating comments works from the example web-wagmi app. 👍

packages/react/src/transactions/adapters/useCreateCommentController.ts
```diff
@@ -0,0 +1,56 @@
...
+    );
+
+    const createComment = new CreateComment(
+      signedCreateComment,
+      gateway,
+      transactionQueue,
+      presenter,
+    );
+
+    void createComment.execute(request);
```
krzysu, line 52: maybe we should await here?
When using the example web-wagmi app, the text input is cleared even before signing the first message in the wallet. If the user rejects the message, the comment is lost.

examples/web-wagmi/src/publications/UseCreateComment.tsx
```diff
@@ -0,0 +1,90 @@
...
+      collect: {
+        type: CollectPolicyType.NO_COLLECT,
+      },
+      reference: ReferencePolicy.ANYBODY,
+    });
+
+    setContent('');
+  };
+
+  return (
```
krzysu, line 44: can we show somewhere a notification/toast/message in the example component when the comment was created successfully?

packages/react/src/transactions/adapters/useCreateCommentController.ts
cesarenaldi: I agree. When @desfero pointed out the issue with upload publication metadata errors and I looked into it, I ended up awaiting for this promise so to let possible exceptions during the upload bubble up to the `useCreatePost` hook. I think it should happen the same here and in `useCreateComment`.

packages/react/src/transactions/adapters/useCreateCommentController.ts
```diff
@@ -0,0 +1,56 @@
...
+    apolloClient,
+    transactionFactory,
+    activeWallet,
+    transactionGateway,
+    protocolCallRelayer,
+    transactionQueue,
+  } = useSharedDependencies();
+
+  return async (request: CreateCommentRequest) => {
+    const gateway = new CommentCallGateway(apolloClient, transactionFactory, upload);
```
cesarenaldi, line 29: Totally agree in deferring the creation of gateway now that is no longer a shared resource.

examples/web-wagmi/src/publications/UseCreateComment.tsx
cesarenaldi: This is an interesting topic. I created a separate ticket to discuss how to move this proposal: https://avara.height.app/lens-sdk/T-2943

packages/react/src/transactions/adapters/__tests__/PublicationCallGateway.spec.ts
desfero: moved

packages/react/src/transactions/adapters/useCreateCommentController.ts
desfero: Right, was waiting for #61 to be merged to rebase and fix also catching the `uploadHandler` error.

desfero:
@cesarenaldi @krzysu should be ready for review. 
I'm working on a follow-up PR that would add a `PendingComment` type to the `publications` query so we can have a similar UI that the `useCreatePost` example has where the post dynamically appears after posting.

krzysu (approved):
LGTM!

cesarenaldi (approved):

PR #58, opened 2023-01-03
feat: T-660 create post optimistic update

cesarenaldi:


packages/api-bindings/src/graphql/common.graphql
```diff
@@ -116,6 +116,19 @@ fragment Wallet on Wallet {
   }
 }
 
+fragment Media on Media {
```
cesarenaldi, line 119: I also dropped the `Fields` suffix while I was moving this definition from profile GQL to common GQL definitions file.

packages/react/src/feed/useFeed.ts
```diff
@@ -15,15 +15,30 @@ export function useFeed({
...
+      data: undefined,
+      ...rest,
+    };
+  }
+
+  return {
+    loading: false,
+    data,
+    ...rest,
+  };
```
cesarenaldi, line 43: This change is not needed, reverting in following PR.

packages/react/src/helpers.ts
```diff
@@ -58,10 +58,14 @@ export type PaginatedReadResult<T> = ReadResult<T> & {
...
 
+type PaginatedQueryResult<K> = {
+  result: { pageInfo: CommonPaginatedResultInfoFragment; items: K };
+};
+
 export function usePaginatedReadResult<
   K,
-  T extends { result: { pageInfo: CommonPaginatedResultInfoFragment; items: K } },
   V,
+  T extends PaginatedQueryResult<K> = PaginatedQueryResult<K>,
```
cesarenaldi, line 68: Default value was needed for a previous attempt. But find this increased readability so I kept `PaginatedQueryResult` and its usage.

packages/react/src/transactions/adapters/responders/CreatePostResponder.ts
```diff
@@ -0,0 +1,162 @@
...
+
+function feedItem(rootRef: Reference): FeedItemFragmentEntry {
+  return {
+    __typename: 'FeedItem',
+    root: rootRef,
+    comments: [],
+  };
+}
+
+export class CreatePostResponder implements ITransactionResponder<CreatePostRequest> {
```
cesarenaldi, line 71: Missing rollback logic coming in following PR.

packages/api-bindings/src/graphql/common.graphql
reecejohnson: I like it!

packages/api-bindings/src/apollo/transforms.ts
```diff
@@ -0,0 +1,26 @@
...
+      enter: (node) => {
+        if (node.typeCondition?.name.value === fragmentName) {
+          return null;
+        }
+        return;
+      },
+    },
+    FragmentSpread: { enter },
+    FragmentDefinition: { enter },
+  }) as DocumentNode; // 😞
```
reecejohnson, line 25: 💔

reecejohnson (approved):
Looks good!

examples/web-wagmi/src/components/header/Header.tsx
```diff
@@ -9,28 +9,32 @@ export function Header() {
...
         style={{
           display: 'flex',
           alignItems: 'center',
           justifyContent: 'space-between',
-          gap: '1rem',
-          position: 'absolute',
-          top: '1rem',
-          right: '1rem',
-          width: 'auto',
+          width: '100%',
```
desfero, line 17: do we need `width: 100%` ?

packages/react/src/helpers.ts
desfero: Do we overall need that third generic parameter? Any benefits of keeping it over just inlining?

packages/react/src/shared.tsx
```diff
@@ -115,7 +116,7 @@ export function createSharedDependencies(config: LensConfig, { onLogout, onError
     [TransactionKind.APPROVE_MODULE]: new NoopResponder(),
     [TransactionKind.COLLECT_PUBLICATION]: new NoopResponder(),
     [TransactionKind.CREATE_COMMENT]: new NoopResponder(),
-    [TransactionKind.CREATE_POST]: new NoopResponder(),
+    [TransactionKind.CREATE_POST]: new CreatePostResponder(apolloClient),
```
desfero, line 119: Should we follow the same flow for comments to add comments to the feed?

cesarenaldi: I don't think so, maybe we can consider adding comments to relevant comments query... although matching the cached edge could be tricky without assuming some restriction to the `publication` query. Maybe the fact we have a `useComments` dedicated hook which encapsulate the query request details can help in that sense. Thoughts?

packages/react/src/transactions/adapters/responders/CreatePostResponder.ts
```diff
@@ -0,0 +1,162 @@
...
+  BroadcastedTransactionData,
+  ITransactionResponder,
+  TransactionData,
+} from '@lens-protocol/domain/use-cases/transactions';
+import { invariant, never } from '@lens-protocol/shared-kernel';
+
+type FeedItemFragmentEntry = {
+  __typename: 'FeedItem';
+  root: Reference;
+  comments: Maybe<Array<Reference>>;
```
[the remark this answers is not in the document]
cesarenaldi: Don't know... just used it TBH. Opinions?

examples/web-wagmi/src/components/header/Header.tsx
cesarenaldi: I can double check... this was before xmas.

packages/react/src/helpers.ts
cesarenaldi: I guess we technically don't... without will make the argument type slightly heavy to read (not strong about it)...

```ts
{ error, data, loading, fetchMore }: QueryResult<PaginatedQueryResult<K>, V>
```
Does this make any difference?

```ts
export function usePaginatedReadResult<
  K,
  V
  T = PaginatedQueryResult<K>
>(...) {...}
```

packages/react/src/transactions/adapters/responders/CreatePostResponder.ts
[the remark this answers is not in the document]
cesarenaldi: What is the most common approach so far `Array<T>` or `T[]` in our codebase? What do you reckon?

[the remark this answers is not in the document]
desfero: `Array<T>` is used in just a couple of places. 
What I recon now is that with `T[]` it's hard to type an array of unions as you either have to use parenthesis or introduce a new type whereas with `Array` it's just `Array<One | Two>`.

packages/react/src/shared.tsx
desfero: not sure about touching yet relevant comments query, let's see later from the consumer's side if they see it's required there.
What we can do right now is to just update `profile` comments list.
Does that make sense?

packages/react/src/helpers.ts
desfero: no difference from what i can tell, just was curious why it's there.
personally, I just try to keep the amount of generic to what is really required as it makes the `hover over popup` easier to consume but not big deal.

packages/react/src/shared.tsx
cesarenaldi: I don't think have a dedicated hook for a profile comments.... in the end it was a way to call the `publications` query... is that you are referring to with?
> update profile comments list

desfero: I see, now I get the problem. 
Not sure if the dedicated hook would change the complexity as it would not affect how you need to manually match the edge in the cache. 
Anyway we have a custom `keyArgs` configuration for the `publication` query which can be used to properly match the cached edge, right?

packages/react/src/helpers.ts
cesarenaldi: I can relate to it. Addressed here: https://github.com/lens-protocol/lens-sdk/pull/61/commits/7c92d07195fc6f8be041351d0bf7c0785cc93d3b

examples/web-wagmi/src/components/header/Header.tsx
cesarenaldi: Removed here: https://github.com/lens-protocol/lens-sdk/pull/61/commits/7c92d07195fc6f8be041351d0bf7c0785cc93d3b

packages/react/src/transactions/adapters/responders/CreatePostResponder.ts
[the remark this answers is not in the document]
cesarenaldi: Changed here: https://github.com/lens-protocol/lens-sdk/pull/61/commits/7c92d07195fc6f8be041351d0bf7c0785cc93d3b

desfero (approved):

packages/react/src/helpers.ts
cesarenaldi: @desfero I couldn't pin point why but if the result shape is not part of the generic TS struggles to narrow down the response of a generate query in these circumstances:
```ts
export function useSearchPublications({
  query,
  limit = DEFAULT_PAGINATED_QUERY_LIMIT,
  observerId,
}: UseSearchPublicationsArgs): PaginatedReadResult<(PostFragment | CommentFragment)[]> {
  const { sources, apolloClient } = useSharedDependencies();

  return usePaginatedReadResult(
    useSearchPublicationsQuery({
      variables: {
        query,
        sources,
        limit,
        observerId,
      },
      client: apolloClient,
    }),
  );
}

export function useSearchProfiles({
  query,
  limit = DEFAULT_PAGINATED_QUERY_LIMIT,
  observerId,
}: UseSearchProfilesArgs): PaginatedReadResult<ProfileFieldsFragment[]> {
  const { apolloClient } = useSharedDependencies();

  return usePaginatedReadResult(
    useSearchProfilesQuery({
      variables: {
        query,
        limit,
        observerId,
      },
      client: apolloClient,
    }),
  );
}
```

I don't think it's the reason why it was in the generic in first instance... just found out after rebasing my changes on top of recent `main`.

PR #59, opened 2023-01-04
feat: add unfollow profile

reecejohnson:


packages/react/src/transactions/adapters/useUnfollowController.ts
```diff
@@ -0,0 +1,38 @@
...
+    const unfollowProfiles = new UnfollowProfile(
+      activeWallet,
+      transactionGateway,
+      unfollowProfileCallGateway,
+      protocolCallRelayer,
+      transactionQueue,
+      presenter,
+    );
+
+    void unfollowProfiles.execute(request);
```
desfero, line 34: seems we can just `await` here

examples/web-wagmi/src/profiles/UseFollowAndUnfollow.tsx
```diff
@@ -10,13 +15,19 @@ type ProfileFollowProps = {
...
 
 function FollowButton({ profile }: ProfileFollowProps) {
-  const { follow, isPending } = useFollow({ profile });
+  const { follow, isPending: isFollowing } = useFollow({ profile });
+  const { unfollow, isPending: isUnfollowing } = useUnfollow({ profile });
 
-  if (profile.isFollowedByMe || profile.isOptimisticFollowedByMe) return <p>Following</p>;
+  if (profile.isFollowedByMe || profile.isOptimisticFollowedByMe)
+    return (
+      <button onClick={unfollow} disabled={isUnfollowing || profile.isOptimisticFollowedByMe}>
```
desfero, line 23: is that by design that `Unfollow` button is in `disabled` state until you refrersh the page?
That's caused by `useUnfollowController` being `true` even after we already updated `isFollowedByMe` to be `true`. 
Can we revert to `false` `isFollowedByMe` in `FollowResponser` `commit` phase to unblock unfollowing?

desfero (changes requested):

packages/react/src/transactions/adapters/useUnfollowController.ts
cesarenaldi: I think this is true for most operations... the release of the current call stack for any operation that ends up involving the `TransactionQueue` is done within the `TransactionQueue` itself. Awaiting this promise is a great opportunity to let invariant errors bubble up to the call site, so the developer can act on coding mistakes or bugs involving their setup.

reecejohnson: thanks yeah, this is good find. added in the `await` 👍

packages/react/src/transactions/useUnfollow.ts
```diff
@@ -0,0 +1,45 @@
...
+import { useState } from 'react';
+
+import { TransactionState, useHasPendingTransaction } from './adapters/TransactionQueuePresenter';
+import { useUnfollowController } from './adapters/useUnfollowController';
+
+export type UseUnfollowArgs = {
+  profile: ProfileFieldsFragment;
+};
+
+export function useUnfollow({ profile }: UseUnfollowArgs) {
```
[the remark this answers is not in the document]
reecejohnson: added 👍

desfero (approved):

packages/react/src/transactions/adapters/useUnfollowController.ts
krzysu: did you? I still can see `void` there 🤔

PR #60, opened 2023-01-04
feat: implement `useWhoReacted` hook (T-293)

krzysu:
implement `useWhoReacted` hook with an example and test

documentation https://docs.lens.xyz/docs/cyan-plums-approve-usewhoreacted


reecejohnson (approved):
Looks great ser 🫡

PR #61, opened 2023-01-04
feat: improvements on `useCreatePost` optimistic update logic

cesarenaldi:
Changes
- Removes `txHash` from `TransactionError`. This was unused, occasionally `undefined` and source of confusion. The failure is anyway presented with the original Request Model object and if broadcasted the `txHash` so we have plenty of data to report to the user if need to.
- Adds rollback logic in case the tx fails
- Manages scenarios in which the Publication Metadata upload fails. This returns (wrapped in a `FailedUploadError`) back to the `useCreatePost` as `error` the developer can use to diagnose the issue and/or inform the user.

### DIFF
--- a/packages/domain/src/entities/Transactions.ts
+++ b/packages/domain/src/entities/Transactions.ts
@@ -119,7 +119,7 @@ export enum TransactionErrorReason {
 export class TransactionError extends Error {
   name = 'TransactionError' as const;
 
-  constructor(readonly reason: TransactionErrorReason, readonly txHash?: string) {
-    super(`Transaction "${txHash as string}" failed due to: ${reason}`);
+  constructor(readonly reason: TransactionErrorReason) {
+    super(`Transaction failed due to: ${reason}`);
   }
 }
--- a/packages/react/src/feed/useFeed.ts
+++ b/packages/react/src/feed/useFeed.ts
@@ -15,30 +15,15 @@ export function useFeed({
 }: UseFeedArgs): PaginatedReadResult<FeedItemFragment[]> {
   const { apolloClient, sources } = useSharedDependencies();
 
-  const { data, loading, ...rest }: PaginatedReadResult<FeedItemFragment[]> =
-    usePaginatedReadResult(
-      useFeedQuery({
-        variables: {
-          profileId,
-          observerId,
-          sources,
-          limit: limit ?? 10,
-        },
-        client: apolloClient,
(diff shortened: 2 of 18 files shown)


packages/react/src/feed/useFeed.ts
```diff
@@ -15,30 +15,15 @@ export function useFeed({
...
-      ...rest,
-    };
-  }
-
-  return {
-    loading: false,
-    data,
-    ...rest,
-  };
+  return usePaginatedReadResult(
```
reecejohnson, line 18: 👍

reecejohnson (approved):
LGTM

packages/domain/src/entities/Transactions.ts
```diff
@@ -119,7 +119,7 @@ export enum TransactionErrorReason {
 export class TransactionError extends Error {
   name = 'TransactionError' as const;
 
-  constructor(readonly reason: TransactionErrorReason, readonly txHash?: string) {
-    super(`Transaction "${txHash as string}" failed due to: ${reason}`);
+  constructor(readonly reason: TransactionErrorReason) {
```
desfero, line 122: agree, that was quite confusing to have `txHash` as optionsl

packages/react/src/transactions/adapters/PublicationCallGateway.ts
```diff
@@ -168,7 +179,10 @@ export class PublicationCallGateway implements ICreatePostCallGateway {
       chainType: ChainType.POLYGON,
       id: v4(),
       request,
-      asyncRelayReceipt: this.initiatePostCreation(request),
+      asyncRelayReceipt: this.initiatePostCreation(
```
desfero, line 182: nit: could we just leave the code as it was in `resolveCreatePostRequestArg` and just call that function over here? 
```
this.initiatePostCreation(await this.resolveCreatePostRequestArg(request))
```

cesarenaldi: ah! that looks neater, I like it! Will have a look at such solution.

cesarenaldi: Much neater! Good suggestion. See here: https://github.com/lens-protocol/lens-sdk/pull/61/commits/7c92d07195fc6f8be041351d0bf7c0785cc93d3b

PR #64, opened 2023-01-05
feat: bundle packages to be Ecmascript modules (T-186)

desfero:
The current support of ESM modules seems to be inconsistent and some of the libraries still don’t provide proper support.
All major bundlers support ESM but also on top of that provide good patching when a module is not proper ESM modules.
The most strict implementation has the Node.js which is the cause reason why next.js example fails.
(For .e.g when no `package.json#exports` is present it will switch back and use `package.json#module` which is the case with apollo client (https://github.com/apollographql/apollo-feature-requests/issues/287 and https://github.com/apollographql/apollo-client/issues/9890) but node.js doesn't support `package.json#modules` and therefore converts module to CommonJS which is different than what bundlers doo (see https://nodejs.org/api/esm.html#commonjs-namespaces))

I had also a look if we can get rid of `folder packages` (the one with just `package.json` files) but unfortunately, we have at least 2 blockers:
- Typescript won’t use `package.json#exports` until we switch our packages to use `moduleResolution: `node next` which on the other hand would require us to add `.js` to all relative imports (see https://github.com/microsoft/TypeScript/issues/51862#issuecomment-1358049778)
- Metro is not supporting `package.json#exports` (see https://github.com/facebook/metro/issues/670)

I tried to get rid also of `mocks` folders as they are importing `devDependencies`	which is something `tsup` does bundle by default (and we want to still keep `import` and use the single version of `jest`)
I have considered using `project references` or importing directly `src` files but without any quick success:
- Project references are not working good with `ts-jest` (see https://github.com/kulshekhar/ts-jest/issues/1648)
- Direct `src` imports won’t work cause `ts-jest` is using `package.json.exports` where `mocks` that point directly to `src` should not be present as we omit `src` folder when publishing package
So for now we just make sure `tsup` doesn't bundle `devDependencies` by overriding `external` config.

The good news is that we can drop `CommonJS` support as all bundlers seem to properly handle ES imports/exports.
- There is a quirk with `jest` as it only supports `imports/exports` when running in `ESM modules` mode (https://jestjs.io/docs/ecmascript-modules). We can’t enable ESM for jest until apollo fixes https://github.com/apollographql/apollo-client/issues/9890. So for now I’m just transpiling `@lens-protocol` packages by using `transformIgnorePatterns` in `jest.config.js`.

So to summarise:
- We only now ship ESM Modules
- Until the `@apollo/client` fixes the ESM modules support Next.js example won’t work without either
    - transpiling `@lens-protocol` package (the easiest and recommended) 
    - making sure that any of the @lens-protocol imports don’t get imported by `node.js` (which basically means no @lens-protocol imports inside files that are imported during SSR


desfero:
Link T-186

examples/web-wagmi/src/profiles/components/ProfileCard.tsx
```diff
@@ -10,9 +9,7 @@ type ProfileCardProps = {
 export function ProfileCard({ profile }: ProfileCardProps) {
   return (
     <article>
-      <Link to={`/profile/handle/${profile.handle}`}>
-        <ProfilePicture picture={profile.picture} />
-      </Link>
+      <ProfilePicture picture={profile.picture} />
```
desfero, line 12: There is no support, for now, to provide route `param` with the handle so it was pointing to a nonexisting route.

cesarenaldi:
Thank you for the exhaustive explanation! I'll provide a proper code review shortly.

desfero:
There is one thing I forgot to include, the generation of `d.ts.map` files, which means DX gonna be bad as typescript won't nicely navigate to implementation between packages. Let's review that PR and will do the followup with DX improvements this week.

examples/web-wagmi/tsconfig.json
```diff
@@ -3,7 +3,7 @@
     "target": "ESNext",
     "useDefineForClassFields": true,
     "lib": ["DOM", "DOM.Iterable", "ESNext"],
-    "allowJs": false,
+    "allowJs": true,
```
cesarenaldi, line 6: Is this intentional?

desfero:
@cesarenaldi the `tsconfig.base.json` is left intentionally until we get a better understanding of how to solve local development issues that we have right now (a need to rebuild everytime, so bad auto imports).

packages/shared-kernel/jest.config.js
```diff
@@ -1,5 +1,6 @@
 export default {
-  preset: 'ts-jest',
+  preset: 'ts-jest/presets/js-with-ts',
   testEnvironment: 'node',
   testPathIgnorePatterns: ['/node_modules/', '/dist/'],
+  transformIgnorePatterns: [`/node_modules/(?!@lens-protocol/*)`],
```
cesarenaldi, line 5: I assume this was previously working cause it was loading the CJS files.

Jest support for ESM confuses me all the times I look into it. What would happen if we were to keep `ts-jest` preset and add `NODE_OPTIONS=--experimental-vm-modules` to the Jest script?

packages/react/src/transactions/adapters/PendingTransactionGateway/PendingTransactionGateway.ts
```diff
@@ -12,7 +12,7 @@ import {
 } from '@lens-protocol/domain/use-cases/transactions';
 import { assertNever, invariant } from '@lens-protocol/shared-kernel';
 import { IStorage } from '@lens-protocol/storage';
-import differenceBy from 'lodash/differenceBy';
+import differenceBy from 'lodash/differenceBy.js';
```
cesarenaldi, line 15: Out of curiosity, are you aware of any plan for Lodash to have proper `exports` configuration?

cesarenaldi:
Great work! Thank you for looking into this annoying issue.

cesarenaldi (approved):

packages/react/src/transactions/adapters/PendingTransactionGateway/PendingTransactionGateway.ts
desfero: There is `lodash-es` that's under the hood using `es` exports, maybe it's worth to migrate at some point (or completely remove the use of lodash given it's used in just a couple of places)

examples/web-wagmi/tsconfig.json
desfero: All changes to the `package.json` in `src/packages` are intentional, this one is just making everything in sync across all `tsconfig`.

cesarenaldi (approved):

PR #65, opened 2023-01-06
feat: amend `useCreateProfile` and add example

reecejohnson:
- Adds `useHasPendingTransaction` to `useCreateProfile` so consumer has ability to know when profile creation tx has been indexed
- Adds internal state in `useCreateProfile` to manage pending status and handle
- Adds basic create profile example with validation, error reporting and viewing newly created profile
- Introduces `useWaitUntilTransactionSettled` for use in scenarios where we need to know if an action has been confirmed on-chain


packages/react/src/profile/useCreateProfile.ts
```diff
@@ -1,19 +1,36 @@
...
+      transaction.request.kind === TransactionKind.CREATE_PROFILE &&
+      transaction.request.handle === handle,
+  );
 
+  return {
+    create: async (_handle: string) => {
       try {
-        const result = await createProfile(handle);
+        setHandle(_handle);
+        setIsPending(true);
```
reecejohnson, line 30: We need this extra state here to achieve the "instant loading" feel in the UI.

Without this, you click the `Create Profile` button and the `isPending` flag remains as `false` for a few ms/s, giving the sense to the user that nothing is happening.

packages/react/src/profile/useCreateProfile.ts
```diff
@@ -1,19 +1,36 @@
...
-    create: async (handle: string) => {
-      setIsPending(true);
+  const hasPendingTx = useHasPendingTransaction(
+    (transaction): transaction is TransactionState<CreateProfileRequest> =>
+      transaction.request.kind === TransactionKind.CREATE_PROFILE &&
+      transaction.request.handle === handle,
+  );
 
+  return {
+    create: async (_handle: string) => {
```
reecejohnson, line 27: Experimented with passing the handle directly into `useCreateProfile` similar to how we do for `useFollow`.

This made things very difficult in the markup side of things, not being able to call the `useCreateProfile` before having a defined handle. Doesn't lend itself well to forms.

examples/web-wagmi/src/profiles/UseCreateProfile.tsx
```diff
@@ -0,0 +1,76 @@
...
+            }}
+          />
+          <button type="submit" disabled={!handle || isPending}>
+            {isPending ? 'Creating...' : 'Create profile'}
+          </button>
+        </div>
+      </form>
+
+      <div>
+        {handle && !isPending && submitted && !error && !isViewCreatedProfileOpen && (
```
krzysu, line 55: is there any more dev friendly way to know if the profile was created? maybe just use `useProfile` directly in UseCreateProfile component to show error when handle doesn't exists?
just this chain of conditions in this line doesn't look nice.

examples/web-wagmi/src/profiles/test.tsx
```diff
@@ -0,0 +1,19 @@
+import { ProfileFieldsFragment, useFollow } from '@lens-protocol/react';
+
+type ProfileFollowProps = {
+  profile: ProfileFieldsFragment;
+};
+
+export function FollowProfile({ profile }: ProfileFollowProps) {
```
krzysu, line 7: was this file intended to be added to the repo? (test.tsx)

examples/web-wagmi/src/profiles/UseCreateProfile.tsx
reecejohnson: Agreed it doesn't look nice, but we need to guard against calling the API after each change to the handle input

examples/web-wagmi/src/profiles/test.tsx
reecejohnson: thanks, removed

examples/web-wagmi/src/profiles/UseCreateProfile.tsx
```diff
@@ -0,0 +1,70 @@
...
+          <button type="submit" disabled={!handle || isPending}>
+            {isPending ? 'Creating...' : 'Create profile'}
+          </button>
+        </div>
+      </form>
+
+      <div>
+        {error && <p>{error.message}</p>}
+
+        {handle && submitted && error === null && <ShowProfile handle={handle} />}
```
reecejohnson, line 57: @krzysu this has been streamlined to check only if we have a handle and have submitted without an error

packages/react/src/profile/useCreateProfile.ts
```diff
@@ -1,23 +1,46 @@
...
+        setIsPending(true);
+        setError(null);
+
         const result = await createProfile(handle);
 
         if (result.isFailure()) {
           setError(result.error);
         }
+
+        await waitUntilTransactionIsSettled(
```
reecejohnson, line 34: @krzysu this allows us to wait until the tx has reached settled status so that we can automatically show the created profile in the UI

packages/domain/src/entities/Transactions.ts
```diff
@@ -119,7 +119,7 @@ export enum TransactionErrorReason {
 export class TransactionError extends Error {
   name = 'TransactionError' as const;
 
-  constructor(readonly reason: TransactionErrorReason) {
-    super(`Transaction failed due to: ${reason}`);
+  constructor(readonly reason: TransactionErrorReason, readonly txHash?: string) {
```
krzysu, line 122: I thought Cesare removed that in some other PR, I guess it's coming from a merge conflict.

krzysu (approved):
I like it! it looks and feels good now. 👍

packages/domain/src/entities/Transactions.ts
reecejohnson: cool, removed it here again too

PR #70, opened 2023-01-11
test: adds missing adapter tests

reecejohnson:
Adds missing tests for recently introduced adapters.


packages/api-bindings/src/mocks.ts
```diff
@@ -1,2 +1,3 @@
 export * from './graphql/__helpers__';
 export * from './apollo/__helpers__/mocks';
+export { createProfileTypePolicy } from './apollo/createProfileTypePolicy';
```
reecejohnson, line 3: We need to access this in `FollowProfilesResponder.spec.ts` as part of the Apollo client setup.

Although not strictly a mock, we shouldn't be exporting as part of the main package

packages/react/src/transactions/adapters/responders/__tests__/FollowProfilesResponder.spec.ts
```diff
@@ -0,0 +1,137 @@
...
+
+import { FollowProfilesResponder } from '../FollowProfilesResponder';
+
+function setupTestScenario({ existingProfile }: { existingProfile: ProfileFieldsFragment }) {
+  const apolloCache = new InMemoryCache({
+    addTypename: true,
+    typePolicies: {
+      Profile: createProfileTypePolicy(),
+    },
+  });
```
cesarenaldi, line 20: I created the `createMockApolloCache` exactly for this purpose. It gives you an `InMemory` cache with all the type policies and possible types wired in.

packages/api-bindings/src/mocks.ts
cesarenaldi: We can avoid this by using the `createMockApolloCache` helper. See my other comment.

desfero (approved):
Some of the tests would fail when we merge #71, right @reecejohnson ?

PR #71, opened 2023-01-11
feat: optimistic updates and defence for follow/unfollow

reecejohnson:
- Adds optimistic update of UI for follow
- Adds defence in `useUnfollow` hook is a profile is only optimistically following another and attempts to unfollow

### DIFF
--- a/packages/domain/src/use-cases/profile/UnfollowProfile.ts
+++ b/packages/domain/src/use-cases/profile/UnfollowProfile.ts
@@ -10,6 +10,12 @@ export type UnfollowRequest = {
   kind: TransactionKind.UNFOLLOW_PROFILE;
 };
 
+export class OptimisticUnfollowError extends Error {
+  constructor(message: string) {
+    super(message);
+  }
+}
+
 export type IUnfollowProfileCallGateway = IUnsignedProtocolCallGateway<UnfollowRequest>;
 
 export type IUnfollowProfilePresenter = IProtocolCallPresenter;
--- a/packages/domain/src/use-cases/profile/UnfollowProfile.ts
+++ b/packages/domain/src/use-cases/profile/UnfollowProfile.ts
@@ -10,12 +10,6 @@ export type UnfollowRequest = {
   kind: TransactionKind.UNFOLLOW_PROFILE;
 };
 
-export class OptimisticUnfollowError extends Error {
-  constructor(message: string) {
-    super(message);
-  }
-}
-
 export type IUnfollowProfileCallGateway = IUnsignedProtocolCallGateway<UnfollowRequest>;
 
 export type IUnfollowProfilePresenter = IProtocolCallPresenter;
--- a/packages/react/src/transactions/useFollow.ts
+++ b/packages/react/src/transactions/useFollow.ts
@@ -1,10 +1,10 @@
 import {
-  ProfileFieldsFragmentWithSupportedFollowModule,
-  ProfileFieldsFragment,
-  isProfileFieldsFragmentWithFeeFollowModule,
-  getFollowPolicyTypeFromProfileFieldsFragment,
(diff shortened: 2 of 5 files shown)


packages/react/src/transactions/useUnfollow.ts
```diff
@@ -30,6 +35,15 @@ export function useUnfollow({ profile }: UseUnfollowArgs) {
...
   return {
     unfollow: async () => {
+      if (!profile.isFollowedByMe && profile.isOptimisticFollowedByMe) {
+        setError(
+          new OptimisticUnfollowError(
+            `Your follow request for ${profile.handle} is still pending.`,
+          ),
+        );
+        return;
+      }
```
cesarenaldi, line 45: Neat!

packages/domain/src/use-cases/profile/UnfollowProfile.ts
```diff
@@ -10,6 +10,12 @@ export type UnfollowRequest = {
   kind: TransactionKind.UNFOLLOW_PROFILE;
 };
 
+export class OptimisticUnfollowError extends Error {
+  constructor(message: string) {
+    super(message);
+  }
+}
```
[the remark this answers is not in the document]
reecejohnson: okay yeah, makes sense.

i like `PrematureUnfollow`

packages/react/src/transactions/useFollow.ts
```diff
@@ -97,35 +92,36 @@ export function useFollow({ profile }: UseFollowArgs) {
     | WalletConnectionError
     | null
   >(null);
+  const [isPending, setIsPending] = useState<boolean>(false);
   const follow = useFollowController();
   const activeWallet = useActiveWallet();
   const { data: activeProfile } = useActiveProfile();
```
desfero, line 98: interesting observation that `useActiveWallet` and `useActiveProfile` do not follow the same approach of returning `wallet/profile`

reecejohnson: It is interesting, both are using reactive vars but active profile has this logic:

```typescript
if (state === ApplicationsState.LOADING) {
    return {
      loading: true,
      data: undefined,
    };
  }
  ```
  
  do we know the reason for the discrepancy?  @cesarenaldi @desfero

krzysu (approved):
👍

packages/react/src/transactions/useFollow.ts
desfero: it appears as a precaution to not start loading queries until we know that the app signaled readiness.
I can see that when the code evolves here (https://github.com/lens-protocol/lens-sdk/blob/3556929122e89389bf9d4640afd03810a0516065/packages/domain/src/use-cases/lifecycle/Bootstrap.ts#L70-L77) we can end up with some race condition between a call to `activeProfile.loadActiveProfileByOwnerAddress` and `applicationPresenter.signalReady.

reecejohnson: after line 73 in the code above, we know for sure that we have a an active profile, so it looks like the checks are fine?

desfero: i guess soo, maybe @cesarenaldi knows some edge cases?

cesarenaldi: It has to do with avoiding flash of logged-out content. To know if we have done-done and if we have a profile or not the only reliable way was to wait for the ready signal.

Arguably we can shift the logged-in/out responsibility to the `useActiveWallet` and let the `useActiveProfile` just propagate the content of the Reactive Var, thoughts?

reecejohnson: Continuing this conversation here: https://avara.height.app/T-3548 🫡

PR #72, opened 2023-01-11
feat: `useUpdateProfileImage` hook (T-1547)

krzysu:
add `useUpdateProfileImage` hook

documentation https://docs.lens.xyz/docs/pink-rats-pump-useupdateprofileimage


packages/react/src/profile/useUpdateProfileImage.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  const updateImage = useUpdateProfileImageController();
+
+  const hasPendingTx = useHasPendingTransaction(
+    (tx): tx is TransactionState<UpdateProfileImageRequest> =>
+      tx.request.kind === TransactionKind.UPDATE_PROFILE_IMAGE &&
+      isUpdateOffChainProfileImageRequest(tx.request),
+  );
+
+  const update = useCallback(
```
desfero, line 43: Any specific reason to use `useCallback`?

examples/web-wagmi/src/upload.ts
```diff
@@ -1,41 +1,78 @@
...
+    throw new Error('User cancelled');
   }
 
-  const serialized = JSON.stringify(data);
-  const tx = await bundlr.upload(serialized, {
-    tags: [{ name: 'Content-Type', value: 'application/json' }],
+  const bundlr = await getBundlr();
+
+  const webStream = file.stream();
+  const nodeStream = new ReadableWebToNodeStream(webStream);
```
desfero, line 72: just curious, any specific reason we are not using the guide from bundler docs? https://docs.bundlr.network/tutorials/file-uploader
I see they are using a package called `filereader-stream`.

packages/react/src/transactions/adapters/responders/__tests__/UpdateProfileImageResponder.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+        url: newImageUrl,
+      }),
+    });
+
+    it(`should update profile picture to the latest one`, async () => {
+      const scenario = setupTestScenario({ profile });
+
+      await scenario.responder.commit(transactionData);
+
+      expect(scenario.profileFromCache).toEqual(profile);
```
desfero, line 57: it appears this test is a little bit broken.
`commit` should modify the profile `picture`, then how it's possible that `scenario.profileFromCache === profile` after updating the cache with a new picture? 

it appears that the `arrange` part of the unit test is wrong as we don't set up a different profile image than the one generated by `faker.image.imageUrl()`.
I'm surprised that `faker.image.imageUrl` returns the same URL every time, that's why the test is passing.
When creating a profile on line `40` we do need to provide some fake `url` which we would then try to overwrite with a new fake url (i suggest to not use `faker.image.imageUrl` at all given it doesn't create random url)

desfero (changes requested):

examples/web-wagmi/src/profiles/UseUpdateProfileImage.tsx
```diff
@@ -0,0 +1,156 @@
...
+        setIsUploading(false);
+        if (e instanceof Error) {
+          setUploadError(`Upload error: ${e.message}`);
+        }
+      }
+    }
+  };
+
+  const handleUpdateProfileImage = async () => {
+    const url = await uploadImageCandidate();
```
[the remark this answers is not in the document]
krzysu: I don't see how I can make `uploadImageCandidate` always return a string (unless empty string) as `uploadImage` can fail in many ways, but I added invariant checks instead of `if` checks.

examples/web-wagmi/src/upload.ts
krzysu: first of all `filereader-stream` doesn't provide typescript types, it's old and depends on even older modules. I didn't want to include it in our example.
I've spent a few hours trying to understand what it is doing and simplify it for our use case. Unfortunately `WebBundlr` implementation is heavily based on node/common version and works with node.js types, so the problem was how to make it work with web stream or buffer.
Cesare helped me here with finding ReadableWebToNodeStream module that glued it all together.
In general `WebBundlr` is not really web-friendly 😄

packages/react/src/transactions/adapters/responders/__tests__/UpdateProfileImageResponder.spec.ts
krzysu: wow, good catch, I didn't realize that faker is always returning the same image URL. I will fix that.

examples/web-wagmi/src/profiles/UseUpdateProfileImage.tsx
```diff
@@ -0,0 +1,156 @@
...
+      </button>
+      {error && <p>{error.message}</p>}
+      {isPending && <p>Updating your Lens profile's image...</p>}
+    </div>
+  );
+}
+
+function UploadAndUpdateNewImage({ profile }: { profile: ProfileFieldsFragment }) {
+  const [candidateFile, setCandidateFile] = useState<ILocalFile<ImageType> | null>(null);
+  const [uploadError, setUploadError] = useState<string>();
```
reecejohnson, line 46: nit: imo it's better to have this typed as the the Error class that can be thrown or null, so we can better distinguish between error or no error. Or maybe even just `string | null`

then we can avoid setting to an empty string to signal there is no error

packages/react/src/profile/useUpdateProfileImage.ts
```diff
@@ -0,0 +1,70 @@
...
+        setIsPending(false);
+      }
+    },
+    [updateImage, profileId],
+  );
+
+  return {
+    update,
+    error,
+    isPending: isPending || hasPendingTx,
```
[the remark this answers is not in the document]
krzysu: yes, that makes sense if I add an optimistic update in the `prepare` phase, I will look into that.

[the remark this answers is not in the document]
reecejohnson: yeah we changed this back in the follow/unfollow hooks too, where `isPending` only represents the promise and not the tx

packages/react/src/transactions/adapters/responders/UpdateProfileImageResponder.ts
```diff
@@ -0,0 +1,45 @@
...
+): request is UpdateOffChainProfileImageRequest {
+  return 'url' in request;
+}
+
+export class UpdateProfileImageResponder
+  implements ITransactionResponder<UpdateProfileImageRequest>
+{
+  constructor(private apolloCache: ApolloCache<NormalizedCacheObject>) {}
+
+  async commit({ request }: BroadcastedTransactionData<UpdateProfileImageRequest>) {
```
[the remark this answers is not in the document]
reecejohnson: +1 for optimistic update here and toast notification if error

packages/react/src/transactions/adapters/responders/__tests__/UpdateProfileImageResponder.spec.ts
krzysu: so the 4th argument is a randomize flag :) I fixed the test
```
faker.image.imageUrl(1234, 2345, 'cat', true) // 'https://loremflickr.com/1234/2345/cat?lock=6849'
```

krzysu:
I added an optimistic update with Responders `prepare` and `rollback` and fixed everything else that was mentioned in the reviews. Please @desfero @reecejohnson  take a look again!
Do you know how to create a condition that would trigger a rollback in the example app?

packages/react/src/transactions/adapters/responders/UpdateProfileImageResponder.ts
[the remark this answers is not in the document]
cesarenaldi: @krzysu you can also use the state of the responder instance, it's an object in the end.

Although it's not about rollback we do something akin to using infos from previous calls in the `UpdateProfileMetadataResponder` (see my last PR: https://github.com/lens-protocol/lens-sdk/pull/78

packages/react/src/transactions/adapters/responders/__tests__/UpdateProfileImageResponder.spec.ts
cesarenaldi: wat? Faker returns the same image URL?

packages/react/src/transactions/adapters/responders/UpdateProfileImageResponder.ts
```diff
@@ -0,0 +1,67 @@
...
+              },
+            };
+          },
+        },
+      });
+    }
+  }
+
+  async commit(_: BroadcastedTransactionData<UpdateProfileImageRequest>) {
+    //
```
cesarenaldi, line 52: In other places we do perform a profile query with `network-only` fetch policy so to update the cache. API might return different URLs to optimised media. The bigger argument is for publication media so cannot predict if and when this will be extended to profile image.

packages/react/src/transactions/adapters/responders/UpdateProfileImageResponder.ts
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  async rollback({ request }: BroadcastedTransactionData<UpdateProfileImageRequest>) {
+    // we don't know the previous profile image url, so just query the profile again
+    if (isUpdateOffChainProfileImageRequest(request)) {
+      await this.client.query<GetProfileQuery, GetProfileQueryVariables>({
+        query: GetProfileDocument,
+        variables: {
+          request: { profileId: request.profileId },
```
cesarenaldi, line 61: Should we include the `observerId` so relevant edges are tailored to the current profile?

Thinking out loud: is it safe to use `request.profileId` or should we access the active profile id?

packages/react/src/transactions/adapters/responders/__tests__/UpdateProfileImageResponder.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+          },
+        },
+      });
+    });
+  });
+
+  describe(`when "${UpdateProfileImageResponder.prototype.rollback.name}" method is invoked`, () => {
+    it(`should restore the original profile picture from the server`, async () => {
+      const scenario = setupTestScenario({ profile });
+      await scenario.responder.rollback(transactionData);
```
cesarenaldi, line 78: I believe this test might stay green even if we remove the exercise line.

I reckon we should use the `prepare` call as test setup.
```ts
const scenario = setupTestScenario({ profile });
await scenario.responder.prepare(transactionData);

await scenario.responder.rollback(transactionData);

expect(scenario.profileFromCache).toMatchObject(profile);
```

```

Comment (krzysu) [reply packages/react/src/transactions/adapters/responders/__tests__/UpdateProfileImageResponder.spec.ts:80]: you are right, fixed

File: packages/react/src/transactions/adapters/responders/UpdateProfileImageResponder.ts (hunk 9 above)
Comment (krzysu) [reply packages/react/src/transactions/adapters/responders/UpdateProfileImageResponder.ts:61]: I could also use the `Map<ProfileId, Url>` approach and just revert the URL in the cache, which one do you think it's better? I guess cache-only has a lower impact.

File: TODO.md [hunk 11]
```diff
@@ -25,8 +25,8 @@
   - [ ] signed, gas-less
   - [ ] signed, paid gas (fallback)
 - [ ] `useUpdateProfileImage` - upload
-  - [ ] sign-less, gas-less (via dispatcher)
-  - [ ] signed, gas-less
+  - [x] sign-less, gas-less (via dispatcher)
```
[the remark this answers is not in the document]
krzysu: Right, I created subtasks for this ticket

packages/react/src/transactions/adapters/responders/UpdateProfileImageResponder.ts
krzysu: in the end, it was way easier to implement a fresh query on commit and rollback than to store the URL value in the Map. Will that be enough?

cesarenaldi: I think it's fine. I was thinking about scenarios where there are several updates related to a profile in the queue (e.g. display name, cover image, profile image)... but couldn't think of scenario that yields the sequencing of the `prepare/commit|rollback` in the wrong order.

In this comment I was asking a different thing though. I was entertaining the idea if we should include `observerId` in the query request var.

krzysu: oh yes, I added `observerId` too to the query (I just haven't pushed that yet). 
One has to be an activeUser to update their profile image so one will be also the observer. Could adding the observer harm any other use case?

krzysu: pushed now https://github.com/lens-protocol/lens-sdk/pull/72/commits/2037406cfc565783a868e4890a7dfa0b05c55701

cesarenaldi: I think there is a chance that a user makes an update to the profile picture of one of the profiles they own... then switch the current active profile (T-1368 I just started) before the tx settles. By the time the commit/rollback happens the observer will be technically wrong.

The thing is that nobody might notice cause it affects fields that one is unlikely to care about when looking at their own profiles (e.g. `isFollowing(profileId: $observerId)`).

Let's not worry about this for now.

PR #73, opened 2023-01-12
feat: useUpdateProfileDetails integration branch

cesarenaldi:
This is now the integration branch for all T-1558 related PRs.

### Previously

Profile Attributes can now be read as following. Let's say your app reads an attribute called `location` that you know being a string.

```
const location = profile.attributes.location?.toString() ?? null // string | null
```

Let's say instead you have a DoB field you know contains a `Date`
```
const dob = profile.attributes.dob?.toDate() // Date | null
```

The changes here gives helper to access Number, Date, String, Boolean attributes.


packages/api-bindings/src/graphql/profile.graphql
```diff
@@ -141,3 +145,49 @@ query MutualFollowersProfiles(
...
+
+    ... on RelayerResult {
+      ...RelayerResult
+    }
+
+    ... on RelayError {
+      ...RelayError
+    }
+  }
+}
```
cesarenaldi, line 193: These 2 mutations are strictly not related to the changes in this PR but part of the same feature. They will be used in the follow PR.

packages/domain/src/use-cases/profile/UpdateProfileDetails.ts
```diff
@@ -5,12 +5,17 @@ import {
 } from '../transactions/DelegableProtocolCallUseCase';
 import { IProtocolCallPresenter } from '../transactions/ProtocolCallUseCase';
 
+export type ProfileAttributeValue = Date | string | number | null;
+
+export type PartialAttributesUpdate = Record<string, ProfileAttributeValue>;
+
 export type ProfileDetails = {
-  name: string;
+  // attributes: PartialAttributesUpdate;
```
cesarenaldi, line 13: Although not used in this PR, this was useful for me to reason around the update path.

This PR implementation is all about the read path. `Profile#attributes` is this object that contains all custom attributes.

In this file you can have a glimpse of what changes are going to happen in the update path to complement the read.

```ts
export type ProfileDetails = {
  attributes: PartialAttributesUpdate;
  name: string;
  bio: string | null;
}
```

reecejohnson (approved):
Nice, this is a big DX improvement.

Works great

packages/api-bindings/src/graphql/profile.graphql
```diff
@@ -72,6 +70,12 @@ fragment ProfileFields on Profile {
       ...RevertFollowModuleSettings
     }
   }
+
+  __attributes: attributes {
```
desfero, line 74: i like the `__` prefix

desfero (approved):

PR #82, opened 2023-01-16
fix: remove redundant method

reecejohnson:


krzysu (approved):

packages/blockchain-bindings/src/errors.ts
```diff
@@ -1,24 +0,0 @@
-// eslint-disable-next-line no-restricted-imports
-import { errorCodes } from 'eth-rpc-errors';
```
desfero, line 2: @reecejohnson should we also remove the `eth-rpc-errors` package?

PR #85, opened 2023-01-16
feat: update follow policy

reecejohnson:
- Adds `useUpdateFollowPolicy` hook
- Adds example usage in `web-wagmi` 
- create client/local field `followPolicy` to act as a more intuitive version of older `followModule` for easier reading/updating of follow module

### Screenshot 
image


packages/react/src/transactions/index.ts
```diff
@@ -21,3 +22,9 @@ export type {
   NoCollectPolicy,
   SupportedPublicationMediaType,
 } from '@lens-protocol/domain/use-cases/publications';
+export type {
+  ChargeFollowPolicy,
+  NoFeeFollowPolicy,
+} from '@lens-protocol/domain/use-cases/profile';
+export { FollowPolicyType } from '@lens-protocol/domain/use-cases/profile';
+export { getFollowPolicyTypeFromProfileFieldsFragment } from '@lens-protocol/api-bindings';
```
cesarenaldi, line 30: It's a bit of a mouthful. Should we provide a local-only field or something instead?

packages/react/src/transactions/useUpdateFollowPolicy.ts
```diff
@@ -0,0 +1,62 @@
...
+
+        if (result.isFailure()) {
+          setError(result.error);
+        }
+
+        await waitUntilTransactionIsSettled(
+          (transaction): transaction is TransactionState<UpdateFollowPolicyRequest> =>
+            transaction.request.kind === TransactionKind.UPDATE_FOLLOW_POLICY &&
+            transaction.request.profileId === profileId,
+        );
```
cesarenaldi, line 54: Do we need to wait? I don't think we necessarily need to have the user waiting, in the end it's about someone else following them. Compared to setting the dispatcher for example I cannot something that the user can do wrong if they we "let them go" when this is not mined/indexed yet.

examples/web-wagmi/src/profiles/UseUpdateFollowPolicy.tsx
```diff
@@ -0,0 +1,213 @@
...
+  );
+
+  const { data: currencies, loading } = useCurrencies();
+  const { updateFollowPolicy, isPending, error } = useUpdateFollowPolicy();
+
+  async function handleSubmit(event: React.FormEvent<HTMLFormElement>) {
+    event.preventDefault();
+    if (!followPolicyTypeToUpdate || !currencies) return;
+
+    const formData = new FormData(event.currentTarget);
```
cesarenaldi, line 85: Thanks you for highlight this approach in my PR https://github.com/lens-protocol/lens-sdk/pull/77. Why not extend this to `followPolicyTypeToUpdate` as well? It seems something that can also be part of form data.

examples/web-wagmi/src/profiles/UseUpdateFollowPolicy.tsx
```diff
@@ -0,0 +1,213 @@
...
+  const [followPolicyTypeToUpdate, setFollowPolicyTypeToUpdate] = useState<FollowPolicyType | null>(
+    currentFollowModule,
+  );
+
+  const { data: currencies, loading } = useCurrencies();
+  const { updateFollowPolicy, isPending, error } = useUpdateFollowPolicy();
+
+  async function handleSubmit(event: React.FormEvent<HTMLFormElement>) {
+    event.preventDefault();
+    if (!followPolicyTypeToUpdate || !currencies) return;
```
cesarenaldi, line 83: nit: If it's possible to include `followPolicyTypeToUpdate` in the "FormData approach" we can also reduce the noise of currencies by either deferring the rendering of the form until currencies are fetched (requires an intermediate component sure but it will have more linear code.

reecejohnson: Thanks, I like this approach, amended the form to be in it's own component where all props are required - much more linear

packages/react/src/transactions/useFollow.ts
```diff
@@ -33,36 +32,42 @@ type FollowProfilesFlowRequest = {
...
         isProfileFieldsFragmentWithFeeFollowModule(profile),
-        'Profile is not with a fee follow module',
+        'Profile must have fee follow module',
       );
 
       return {
         ...baseRequest,
-        fee: createFollowRequestFee(profile.followModule),
+        fee: createFollowRequestFee(
+          profile.followPolicy as ChargeFollowPolicy,
```
cesarenaldi, line 68: I believe the way that `followPolicyType` is computed confused TS so it's not able to narrow down the union.

Maybe worth do an eager return if `profile.followPolicy === null` and then evaluate `profile.followPolicy.type` with the switch.

cesarenaldi: Thinking out loud... any reason to have `profile.followPolicy` nullable at all?

reecejohnson: great idea, we now model each case within `FollowPolicy` so no longer need to support null - this simplifies things massively

examples/web-wagmi/src/App.tsx
```diff
@@ -92,7 +91,6 @@ export function App() {
                 <Route path="/publications/useCreatePost" element={<UseCreatePost />} />
                 <Route path="/publications/useCreateComment" element={<UseCreateComment />} />
                 <Route path="/publications/useReaction" element={<UseReaction />} />
-                <Route path="/publications/UseCreateMirror" element={<UseCreateMirror />} />
```
cesarenaldi, line 95: Is this intentional?

examples/web-wagmi/src/App.tsx
```diff
@@ -112,7 +110,7 @@ export function App() {
                 <Route path="/profiles/useProfilesToFollow" element={<ProfilesToFollow />} />
                 <Route path="/profiles/useFollow" element={<UseFollowAndUnfollow />} />
                 <Route path="/profiles/useUpdateProfileImage" element={<UseUpdateProfileImage />} />
-                <Route path="/profiles/useProfilesOwnedBy" element={<UseProfilesOwnedBy />} />
```
cesarenaldi, line 115: Is the replacement of `UseProfilesOwnedBy` route a merge issue?

examples/web-wagmi/src/profiles/ProfilesPage.tsx
```diff
@@ -37,9 +37,14 @@ const profileHooks = [
     path: '/profiles/useUpdateProfileImage',
   },
   {
-    label: 'useProfilesOwnedBy',
```
cesarenaldi, line 40: I think this was lost in the merge.

packages/api-bindings/codegen-api.yml
```diff
@@ -50,10 +51,10 @@ generates:
       - 'typescript':
           nonOptionalTypename: true
       - add:
-          content:
-            [
+          content: [
               "import type { ClientErc20Amount } from './ClientErc20Amount';",
-              "import type { ProfileAttributes } from './ProfileAttributes';",
+              "import type { ProfileAttributes } from './ProfileAttributes';
+              import type { FollowPolicy } from '@lens-protocol/domain/use-cases/profile';",
```
cesarenaldi, line 57: nit: can you please use another array entry?

packages/api-bindings/src/apollo/createProfileTypePolicy.ts
```diff
@@ -1,9 +1,47 @@
...
+  followModule,
+}: {
+  followModule: FollowModule | null;
+}): FollowPolicy {
+  const followPolicyType = getFollowPolicyTypeFromProfileFieldsFragment(followModule);
+
+  if (followPolicyType === FollowPolicyType.CHARGE) {
+    invariant(isFeeFollowModule(followModule), 'Profile must have the fee follow module');
+
+    const erc20Value = erc20({
```
cesarenaldi, line 26: Any reason to not use `erc20Amount` helper?

packages/api-bindings/src/graphql/__helpers__/fragments.ts
```diff
@@ -84,7 +84,10 @@ export function mockProfileFieldsFragment(
     },
 
     dispatcher: null,
-    followModule: null,
+
+    __followModule: null,
+    followPolicy: mockFreeFollowPolicy(),
```
cesarenaldi, line 89: Neat choice for the mock default data.

packages/api-bindings/src/apollo/createProfileTypePolicy.ts
```diff
@@ -1,9 +1,47 @@
...
 
+function resolveFollowPolicy({
+  followModule,
+}: {
+  followModule: FollowModule | null;
+}): FollowPolicy {
+  const followPolicyType = getFollowPolicyTypeFromProfileFieldsFragment(followModule);
+
+  if (followPolicyType === FollowPolicyType.CHARGE) {
+    invariant(isFeeFollowModule(followModule), 'Profile must have the fee follow module');
```
cesarenaldi, line 24: Is this invariant needed?

If we were to check the `followModule.__typename` the discriminated union should be narrow down so this invariant and maybe the new `isFeeFollowModule` (if not used anywhere else) are not needed.

Maybe we are using the `getFollowPolicyTypeFromProfileFieldsFragment` too early. It seems is getting in the way.

packages/domain/src/use-cases/profile/UpdateFollowPolicy.ts
```diff
@@ -24,6 +25,12 @@ export type NoFeeFollowPolicy = {
   type: FollowPolicyType.ANYONE | FollowPolicyType.ONLY_PROFILE_OWNERS | FollowPolicyType.NO_ONE;
 };
 
+export type UnsupportedFollowPolicy = {
+  type: FollowPolicyType.UNKNOWN;
+};
+
+export type FollowPolicy = ChargeFollowPolicy | NoFeeFollowPolicy | UnsupportedFollowPolicy;
```
cesarenaldi, line 32: ah ok, this is the source of confusion regarding CA we were discussing in Slack.

I think this:
```ts
export type UnsupportedFollowPolicy = {
  type: FollowPolicyType.UNKNOWN;
};

export type FollowPolicy = ChargeFollowPolicy | NoFeeFollowPolicy | UnsupportedFollowPolicy;
```

is better suited to live in `packages/api-bindings/src/graphql/FollowPolicy.ts`.

packages/api-bindings/src/graphql/utils/profile.ts
```diff
@@ -16,56 +17,71 @@ export type SupportedFollowModuleSettingsFragment =
...
 >;
 
 export function isProfileFieldsFragmentWithSupportedFollowModule(
   profile: ProfileFieldsFragment,
 ): profile is ProfileFieldsFragmentWithSupportedFollowModule {
-  const followPolicyType = getFollowPolicyTypeFromProfileFieldsFragment(profile);
-  return followPolicyType !== null;
+  const followPolicyType = getFollowPolicyTypeFromProfileFieldsFragment(profile.__followModule);
+  return followPolicyType !== FollowPolicyType.UNKNOWN;
 }
```
cesarenaldi, line 43: Is this helper used at all or just got migrated?

It seems if was used to guard UI from rendering in case of unknown follow module.

I think the new `followPolicy: FollowPolicy` edge is doing a better job. One just need to check `profile.followPolicy.type === FollowPolicyType.UNKNOWN`.

packages/api-bindings/src/graphql/utils/profile.ts
```diff
@@ -16,56 +17,71 @@ export type SupportedFollowModuleSettingsFragment =
...
 
 export function isProfileFieldsFragmentWithRevertFollowModule(
```
cesarenaldi, line 45: Same here, helper not used in this codebase nor in this PR. `profile.followPolicy.type === FollowPolicyType.NO_ONE` is pretty good for UI checks.

packages/react/src/transactions/useFollow.ts
```diff
@@ -32,36 +32,39 @@ type FollowProfilesFlowRequest = {
...
   const baseRequest: Pick<FollowProfilesFlowRequest, 'profileId' | 'followerAddress'> = {
     profileId: profile.id,
     followerAddress: activeWallet.address,
   };
 
+  const followPolicyType = profile.followPolicy.type;
+
   switch (followPolicyType) {
     case FollowPolicyType.CHARGE:
       invariant(
```
cesarenaldi, line 60: It should be possible to narrow down the `profile.followPolicy` discriminated union with just the `type`. Wondering if the assignment at line 56 is needed.

In other words it seems we can get away without this invariant check. Unless I am missing something.

packages/api-bindings/src/graphql/utils/profile.ts
```diff
@@ -16,56 +17,71 @@ export type SupportedFollowModuleSettingsFragment =
...
   profile: ProfileFieldsFragment,
 ): profile is ProfileFieldsFragmentWithRevertFollowModule {
-  return getFollowPolicyTypeFromProfileFieldsFragment(profile) === FollowPolicyType.NO_ONE;
+  return (
+    getFollowPolicyTypeFromProfileFieldsFragment(profile.__followModule) === FollowPolicyType.NO_ONE
+  );
 }
 
 export function isProfileFieldsFragmentWithFeeFollowModule(
```
cesarenaldi, line 53: Same rationale... the only call site for this helper seems it could get away without it. I would consider deleting it if not needed.

packages/api-bindings/src/graphql/utils/profile.ts
```diff
@@ -16,56 +17,71 @@ export type SupportedFollowModuleSettingsFragment =
...
+  return (
+    getFollowPolicyTypeFromProfileFieldsFragment(profile.__followModule) === FollowPolicyType.CHARGE
+  );
+}
+
+export function isFeeFollowModule(
+  followModule: ProfileFieldsFragment['__followModule'],
+): followModule is FeeFollowModuleSettingsFragment {
+  return getFollowPolicyTypeFromProfileFieldsFragment(followModule) === FollowPolicyType.CHARGE;
 }
```
cesarenaldi, line 65: See my other comment, this might not be necessary after all.

examples/web-wagmi/src/profiles/UseUpdateFollowPolicy.tsx
```diff
@@ -0,0 +1,280 @@
...
+function UpdateFollowPolicyForm({
+  profile,
+  currencies,
+  currentFollowPolicy,
+}: UpdateFollowPolicyFormProps) {
+  const { updateFollowPolicy, isPending, error } = useUpdateFollowPolicy();
+  const [selectedFollowPolicyType, setSelectedFollowPolicyType] = useState<FollowPolicyType | null>(
+    currentFollowPolicy?.type ?? null,
+  );
+  const [hasSuccessfulUpdatedFollowPolicy, setHasSuccessfulUpdatedFollowPolicy] =
```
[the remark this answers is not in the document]
reecejohnson: this is used to persist the form completed state to avoid the delayed state change appearing for the user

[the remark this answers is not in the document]
reecejohnson: as the `isPending` flag is reset as the promise resolves, `hasSuccessfulUpdatedFollowPolicy` retains it's value until the page is revisited / refreshed

packages/domain/src/use-cases/profile/UpdateFollowPolicy.ts
reecejohnson: Discussed on a call and changes implemented - much more streamlined approach 👍

cesarenaldi (approved):

mmyyrroonn:
@reecejohnson I assume this line should be wrong. Is that true? Since only the unknown type can be followed. https://github.com/lens-protocol/lens-sdk/blob/b5c0427393ef98bbf1c61e12bed3316719e3eea8/packages/react/src/transactions/useFollow.ts#L99

PR #88, opened 2023-01-16
fix: fixes flash of logged-out content from example app

cesarenaldi:
It also aligns `useActiveWallet` to what `useActiveProfile` waits for.


examples/web-wagmi/src/components/auth/LoginButton.tsx
```diff
@@ -32,17 +31,19 @@ export function LoginButton() {
 
   return (
     <>
-      {!isLoggedIn && (
+      <WhenLoggedInWithProfile>
```
desfero, line 34: 👍

desfero (approved):

PR #89, opened 2023-01-16
feat: adds useActiveProfileSwitcher hook

cesarenaldi:
Also removes unused method.


.changeset/afraid-insects-move.md
```diff
@@ -0,0 +1,7 @@
+---
+"@lens-protocol/api-bindings": minor
+"@lens-protocol/domain": minor
+"@lens-protocol/react": minor
```
desfero, line 4: nit: would be better to just mention the package that this change affects the most. Right now changeset will generate the changelog mentioning `Added useProfileOwnedBy hook` in `api-bindings, domain and react` but in reality `useProfileOwnedBy` is only a part of `react` package, and the changes in `api-bindings,domain` are internal one 
(same applies to the `happy-socks-design` changeset)

examples/web-wagmi/src/profiles/ProfilesPage.tsx
```diff
@@ -46,6 +46,16 @@ const profileHooks = [
...
     path: '/profiles/useFollow',
   },
+  {
+    label: 'useProfilesOwnedBy',
+    description: `Fetch a list of profiles owed by a given address.`,
+    path: '/profiles/useProfilesOwnedBy',
+  },
+  {
+    label: 'useActiveProfileSwitch',
+    description: `Switch active profile`,
```
desfero, line 56:
```suggestion
    description: `Switch active profile.`,
```

examples/web-wagmi/src/profiles/UseProfileOwnedBy.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { WhenLoggedInWithProfile, WhenLoggedOut } from '../components/auth/auth';
+import { Loading } from '../components/loading/Loading';
+import { useInfiniteScroll } from '../hooks/useInfiniteScroll';
+import { ProfileCard } from './components/ProfileCard';
+
+type UseProfilesOwnedByProps = {
+  address: EthereumAddress;
+};
+
+function UseProfilesOwnedByActiveWallet({ address }: UseProfilesOwnedByProps) {
```
desfero, line 13: nit: naming wise we don't know in this component if `address` is an active wallet

packages/domain/src/use-cases/profile/ActiveProfileLoader.ts
```diff
@@ -1,20 +1,11 @@
...
-  getProfileByHandle(handle: string): Promise<Profile | null>;
-}
-
-export interface IActiveProfileGateway {
-  setActiveProfile(profile: Profile): Promise<void>;
-  getActiveProfile(): Promise<Profile | null>;
-}
-
-export class ActiveProfile {
+export class ActiveProfileLoader {
```
desfero, line 8: good naming

packages/react/src/profile/adapters/ProfileGateway.ts
```diff
@@ -44,4 +44,21 @@ export class ProfileGateway implements IProfileGateway {
...
     });
   }
+
+  async getProfileById(profileId: string): Promise<Profile | null> {
+    const { data } = await this.apolloClient.query<GetProfileQuery, GetProfileQueryVariables>({
+      query: GetProfileDocument,
+      variables: { request: { profileId } },
+    });
+
+    invariant(data, `Could not query profiles by id: ${profileId}`);
```
desfero, line 54: nit: this is kinda tricky as we are shallowing the original error (the one in the `error` returned from the `apolloClient.query`). Can be confusing later to debug real production errors as only `Could not query profiles by id` message is there without a real reason.
Not saying we have to solve it now, just building awareness.

packages/react/src/publication/usePublications.ts
```diff
@@ -13,7 +13,7 @@ type UsePublicationArgs = PaginatedArgs<{
 export function usePublications({
   profileId,
   observerId,
-  limit,
+  limit = DEFAULT_PAGINATED_QUERY_LIMIT,
```
desfero, line 16: good call

desfero (changes requested):
LGTM, just small nitpicks

.changeset/afraid-insects-move.md
cesarenaldi: got it

examples/web-wagmi/src/profiles/UseProfileOwnedBy.tsx
cesarenaldi: I just assumed cause it's private to this module and only active wallet address is fed into it.

packages/react/src/profile/adapters/ProfileGateway.ts
cesarenaldi: I agree, and this is not the only place where this happens. I created a ticket to discuss this issue: T-4022

desfero (approved):

PR #92, opened 2023-01-17
Add `useWhoCollectedPublication` hook

paulxuca:


packages/react/src/publication/__tests__/useWhoCollectedPublication.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+import { renderHookWithMocks } from '../../__helpers__/testing-library';
+import { useWhoCollectedPublication } from '../useWhoCollectedPublication';
+
+describe('Given the useWhoCollectedPublication hook', () => {
+  const observer = mockProfileFieldsFragment();
+  const mockPublication: PostFragment = mockPostFragment();
+  const mockWallets: WalletFragment[] = [mockWalletFragment()];
+
+  describe('when the query returns data successfully', () => {
+    it.only('should return wallets who collected the publication', async () => {
```
desfero, line 20: nit: leftover `it.only`

packages/react/src/publication/useWhoCollectedPublication.ts
```diff
@@ -0,0 +1,27 @@
...
+export function useWhoCollectedPublication(
+  args: UseWhoCollectedPublicationArgs,
+): PaginatedReadResult<WalletFragment[]> {
+  const { apolloClient } = useSharedDependencies();
+
+  return usePaginatedReadResult(
+    useWhoCollectedPublicationQuery({
+      variables: {
+        publicationId: args.publicationId,
+        observerId: args?.observerId,
```
desfero, line 21: nit: elvis operator not require as `args` is always an `object` (same on the live below)

desfero (approved):

PR #76, opened 2023-01-12
Return `Amount<ERC20>` instead of `Erc20Amount` and `ModuleFeeAmount` in apollo cache

JoaquinBattilana:

PR #95, opened 2023-01-18
feat: added RevenueAggregate.totalAmount: Amount<Erc20>

cesarenaldi:
This PR extracts the excellent work done by @JoaquinBattilana in this PR: https://github.com/lens-protocol/lens-sdk/pull/76 so it's mergeable with most recent `main` changes.

It focused those changes to address just the `RevenueAggregate.totalAmount: Amount<Erc20>` local-only field.

It basically avoids changes to instances of `ModuleFeeAmount` simply because it make sense to be tackle as part of a new client scalar `FollowPolicy`  which will be taken care by @reecejohnson in https://github.com/lens-protocol/lens-sdk/pull/85.

### DIFF
--- a/packages/api-bindings/src/apollo/createRevenueAggregateTypePolicy.ts
+++ b/packages/api-bindings/src/apollo/createRevenueAggregateTypePolicy.ts
@@ -1,0 +1,15 @@
+import { never } from '@lens-protocol/shared-kernel';
+
+import { erc20Amount, RevenueAggregate } from '../graphql';
+import { TypePolicy } from './TypePolicy';
+
+export function createRevenueAggregateTypePolicy(): TypePolicy<RevenueAggregate> {
+  return {
+    fields: {
+      totalAmount(_, { readField }) {
+        const total = readField('total') ?? never('total is null');
+        return erc20Amount({ from: total });
+      },
+    },
+  };
+}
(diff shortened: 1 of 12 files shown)


packages/api-bindings/src/apollo/createRevenueAggregateTypePolicy.ts
```diff
@@ -0,0 +1,15 @@
+import { never } from '@lens-protocol/shared-kernel';
+
+import { erc20Amount, RevenueAggregate } from '../graphql';
+import { TypePolicy } from './TypePolicy';
+
+export function createRevenueAggregateTypePolicy(): TypePolicy<RevenueAggregate> {
+  return {
+    fields: {
+      totalAmount(_, { readField }) {
+        const total = readField('total') ?? never('total is null');
```
desfero, line 10: oh, that's interesting that we don't have to use aliased name `__total` and it can properly map `total` to `__total`

desfero (changes requested):

packages/api-bindings/src/apollo/createRevenueAggregateTypePolicy.ts
cesarenaldi: Yeah, this is not very clear in the Apollo docs... but my understanding is that FieldPolicy applies to field as in "edges of the GQL original nodes"... so if a field is aliased more than once in a given query the field policy will be called as any times as it's number of aliases. The 2nd argument of the field read method should have details of the field actual query including the alias name I think, but I didn't investigate further.

desfero (approved):

PR #96, opened 2023-01-18
feat: simplifies Create Post optimistic update

cesarenaldi:
This now uses an Apollo Reactive Variable rather than Apollo Cache. This will allow to decouple the optimistic update logic from the `useFeed` filters so that we don't have to worry about how to update the correct `feed` field entry in the cache when performing Create Post optimistic update.

Create Comment will inherit a very similar behaviour. It might even generalize some of the code in this PR,


cesarenaldi:
Note for myself: verify if the `feed` Field Policy can be simplified. Do we still need a `keyArgs` function or can it be just a `keyArgs` array?

examples/web-wagmi/src/publications/UseCreatePost.tsx
```diff
@@ -1,7 +1,38 @@
...
+import { useInfiniteScroll } from '../hooks/useInfiniteScroll';
 import { PostComposer } from './components/PostComposer';
+import { PublicationCard } from './components/PublicationCard';
+
+type TimelineProps = {
+  profileId: string;
+};
+
+export function Timeline({ profileId }: TimelineProps) {
+  const recentPosts = useRecentPosts();
```
desfero, line 15: do you think we could call the hook `useRecentPublications` to avoid having 2 hooks to maintain (or introducing breaking change in the next release by renaming this one)?

packages/api-bindings/src/apollo/createExploreProfileFieldPolicy.ts
```diff
@@ -1,5 +1,7 @@
 import { cursorBasedPagination } from './utils/cursorBasedPagination';
 
 export function createExploreProfilesFieldPolicy() {
-  return cursorBasedPagination([['request', ['observerId']]]);
+  return cursorBasedPagination([
+    ['request', ['sortCriteria', 'timestamp', 'publicationTypes', 'excludeProfileIds']],
```
desfero, line 5: so the changes to add these to the `useExploreProfiles` allowed args gonna be done later, right?

desfero (approved):

examples/web-wagmi/src/publications/UseCreatePost.tsx
cesarenaldi: I purposefully been very narrow to the task at hand cause I didn't want to worry about comments. We can introduce a breaking change in 0.x, no problems.

Maybe I would consider to have a `useRecentComment` hook even if under the hood it might/could leverage the same reactive var if need to.

The flows (create comment and create post) in our experience are very distinct and separate... I would carefully re-consider if worth to use the same reactive var or not.

packages/api-bindings/src/apollo/createExploreProfileFieldPolicy.ts
cesarenaldi: I think these are already there for `useExploreProfiles`, I checked the hook args to come up w/ this list.

Fixed it here cause I stumbled on this misconfiguration and it was a quick fix.

PR #80, opened 2023-01-13
feat: add `useReportPublication` hook (T-2513)

krzysu:
add `useReportPublication` hook

documentation https://docs.lens.xyz/docs/use-report-publication

PR #97, opened 2023-01-19
chore: add tests for ReportPublicationGateway

krzysu:
PR #80 was merged, but it was missing the last thing, tests for ReportPublicationGateway.
I added them in this PR, but one test is not working as expected and I can't figure out why.
See comments below.

### DIFF
--- a/packages/api-bindings/src/graphql/__helpers__/mutations.ts
+++ b/packages/api-bindings/src/graphql/__helpers__/mutations.ts
@@ -52,6 +52,9 @@ import {
   CreateSetProfileImageUriViaDispatcherMutationVariables,
   CreateSetProfileImageUriViaDispatcherMutation,
   CreateSetProfileImageUriViaDispatcherDocument,
+  ReportPublicationMutationVariables,
+  ReportPublicationMutation,
+  ReportPublicationDocument,
 } from '../generated';
 
 export function mockCreateProfileMutation(result: Required<RelayResult>): CreateProfileMutation {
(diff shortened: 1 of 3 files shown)


packages/react/src/publication/adapters/__tests__/ReportPublicationGateway.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+        publicationId,
+        reason: ReportReason.FAKE_ENGAGEMENT,
+        additionalComments,
+      });
+      const result = await gateway.report(request);
+
+      expect(result.isSuccess()).toBe(true);
+    });
+
+    it.skip(`should fail with AlreadyReportedError if publication was already reported`, async () => {
```
krzysu, line 49: notice that this test is skipped, as somehow the check for error instance is failing

packages/api-bindings/src/graphql/__helpers__/mutations.ts
```diff
@@ -418,3 +421,29 @@ export function createCreateSetProfileMetadataViaDispatcherMutationMockedRespons
...
+export function createReportPublicationMutationWithErrorMockedResponse(args: {
+  variables: ReportPublicationMutationVariables;
+}): MockedResponse<ReportPublicationMutation> {
+  return {
+    request: {
+      query: ReportPublicationDocument,
+      variables: args.variables,
+    },
+    result: {
+      errors: [new GraphQLError('Publication already reported')],
```
krzysu, line 446: I think the mocked error response is correct..

krzysu: If I log the output of `console.log(e instanceof ApolloError, e);` in the `catch` part of `ReportPublicationGateway.ts` I'm getting:

```
false ApolloError: Publication already reported
        at new ApolloError (...)
   (...)
   {
      graphQLErrors: [ GraphQLError [Object] { message: 'Publication already reported' } ],
      clientErrors: [],
      networkError: null,
      extraInfo: undefined
    }
```
so it looks fine but the `instanceof` check is returning `false`

I don't know why

desfero: @krzysu PR that's fixing the issue https://github.com/lens-protocol/lens-sdk/pull/103

krzysu: I checked, and indeed that solves the issue, thx!

desfero (approved):

PR #98, opened 2023-01-19
feat: Add dispatcher support to `useUpdateProfileImage` (T-3564)

desfero:
- Adds support for dispatcher when using `useUpdateProfileImage`.
- Updates docs https://docs.lens.xyz/docs/use-update-profile-image


desfero:
Link T-3564

desfero:
Link T-3564

packages/react/src/transactions/adapters/publication-call-gateways/__tests__/PostCallGateway.spec.ts
```diff
@@ -71,7 +71,7 @@ const mandatoryFallbackMetadata = (request: CreatePostRequest) => ({
   mainContentFocus: PublicationMainFocus[request.contentFocus],
 });
 
-describe(`Given an instance of ${PostCallGateway.name}`, () => {
+describe(`Given an instance of ${CreatePostCallGateway.name}`, () => {
```
cesarenaldi, line 74: Should we rename the test files for Post and Comment call gw as well?

examples/web-wagmi/src/profiles/UseUpdateProfileImage.tsx
```diff
@@ -63,7 +63,7 @@ function UploadAndUpdateNewImage({ profile }: { profile: ProfileFieldsFragment }
     error: updateError,
     isPending,
   } = useUpdateProfileImage({
-    profileId: profile.id,
+    profile,
   });
```
cesarenaldi, line 67: nit: do we need the wrapping object at all or can we pass `profile` directly?

reecejohnson (approved):
LGTM 👍

examples/web-wagmi/src/profiles/UseUpdateProfileImage.tsx
desfero: personally prefer the wrapping object for the API that's used externally as it's easier to add new fields later without introducing a breaking change. It appears it's also consistent with the hook that we already have: `useUpdateDispatcherConfig`, `useUnfollow`, and others

packages/react/src/transactions/adapters/publication-call-gateways/__tests__/PostCallGateway.spec.ts
desfero: oh, good catch. Fixed

examples/web-wagmi/src/profiles/UseUpdateProfileImage.tsx
cesarenaldi: I think some... the ones about publication do accept the object directly.

PR #62, opened 2023-01-04
feat: T-1508 `usePublicationRevenue` hook

JoaquinBattilana:
Implements `usePublicationRevenue` hook with queries, hook and example

PR #103, opened 2023-01-20
chore: Fix duplicated dependencies due to optional peer dependencies resolution logic

desfero:
This one fixes the duplicated `@apollo/client`. 
The most important part in this PR is the `pnpm-lock` change that's removing one of the `@apollo/client` duplicates (@apollo/client/3.7.1_graphql@15.5.1:)

Will raise the issue to https://github.com/ocavue/pnpm-deduplicate if we can add that as a part of duplicated dependencies checks there.

Also fixed the merge conflict that we have on main after merging #62 and #95.


packages/react/package.json
```diff
@@ -49,7 +49,7 @@
     "@lens-protocol/domain": "workspace:*",
     "@lens-protocol/shared-kernel": "workspace:*",
     "@lens-protocol/storage": "workspace:*",
-    "graphql": "15.5.1",
```
desfero, line 52: Does anyone remember why it was using strict versioning?

reecejohnson: No reason I can remember from my side

krzysu:
is this one ready or we drop it in favor of https://github.com/lens-protocol/lens-sdk/pull/105?

desfero:
@krzysu it's a separate one from #105, just decided to split them as they fix different issues.

krzysu (approved):

PR #104, opened 2023-01-20
feat: formalisation of publication filters

reecejohnson:
- [x] Updated Docs

### DIFF
--- a/packages/api-bindings/src/graphql/comments.graphql
+++ b/packages/api-bindings/src/graphql/comments.graphql
@@ -12,9 +12,16 @@ query Comments(
   $limit: LimitScalar!
   $cursor: Cursor
   $sources: [Sources!]
+  $metadata: PublicationMetadataFilters
 ) {
   result: publications(
-    request: { limit: $limit, cursor: $cursor, commentsOf: $commentsOf, sources: $sources }
+    request: {
+      limit: $limit
+      cursor: $cursor
+      commentsOf: $commentsOf
+      sources: $sources
+      metadata: $metadata
+    }
   ) {
     items {
       ... on Comment {
--- a/packages/react/src/publication/__tests__/filters.spec.ts
+++ b/packages/react/src/publication/__tests__/filters.spec.ts
@@ -1,0 +1,25 @@
+import { PublicationMetadataFilters } from '@lens-protocol/api-bindings';
+
+import { createPublicationMetadataFilters } from '../filters';
+
+describe(`Given ${createPublicationMetadataFilters.name}`, () => {
+  describe('when supplied with allowed filters', () => {
+    it('should return the correct filter for the api', () => {
+      const expectedFilters: PublicationMetadataFilters = {
+        tags: {
+          oneOf: ['sports'],
+        },
+      };
+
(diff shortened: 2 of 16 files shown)


packages/react/src/publication/filters.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export type PublicationFilters = {
+  restrictPublicationMainFocusTo?: PublicationMainFocus[];
+  restrictPublicationTagsTo?: {
+    all?: string[];
+    oneOf?: string[];
+  };
+  showPublicationsWithContentWarnings?: {
+    oneOf?: PublicationContentWarning[];
+  };
```
cesarenaldi, line 11: what is the default behaviour when this is not provided?

paulxuca: It should not be included in the API request.

reecejohnson:
we also need to add the publication metadata filters to `useComments` 🙏

packages/react/src/publication/__tests__/filters.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+        },
+      };
+
+      expect(
+        createPublicationMetadataFilters({
+          restrictPublicationTagsTo: {
+            oneOf: ['sports'],
+          },
+        }),
+      ).toEqual(expectedFilters);
```
reecejohnson, line 20: nice 🫡

packages/react/src/publication/filters.ts
```diff
@@ -0,0 +1,38 @@
+import {
+  PublicationContentWarning,
+  PublicationMainFocus,
+  PublicationMetadataFilters,
+} from '@lens-protocol/api-bindings';
+
+export type StandardPublicationMetadataFilter = {
```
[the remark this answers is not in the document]
paulxuca: Not sure what to name it without clashing with the existing `PublicationMetadataFilters` from the lens API; Will rename to match the existing type from the API.

reecejohnson:
Look good @paulxuca 👍 

I can't approve though as I opened the original PR

@cesarenaldi can you help here?

krzysu (approved):
is this documented?

packages/react/src/publication/filters.ts
cesarenaldi: nit: I believe as of now one can provide and empty object. Is this intentional?
```ts
showPublicationsWithContentWarnings: {}
```

cesarenaldi (approved):
Small comment.

packages/api-bindings/src/graphql/comments.graphql
```diff
@@ -12,9 +12,16 @@ query Comments(
   $limit: LimitScalar!
   $cursor: Cursor
   $sources: [Sources!]
+  $metadata: PublicationMetadataFilters
```
desfero, line 15: should we update our field policies to properly set key args? Right now it's will not keep different `metadata` queries as separate items in the apollo cache.

desfero (changes requested):

desfero (approved):

PR #106, opened 2023-01-23
Add `useProfilesWhoMirroredPublication` hook

paulxuca:


examples/web-wagmi/src/profiles/ProfilesPage.tsx
```diff
@@ -56,6 +56,11 @@ const profileHooks = [
     description: `Switch active profile.`,
     path: '/profiles/useActiveProfileSwitch',
   },
+  {
+    label: 'useProfilesWhoMirroredPublication',
+    description: 'Fetch profiles that have mirrored a specific publication',
+    path: '/discovery/useProfilesWhoMirroredPublication',
```
krzysu, line 62: wrong path
```suggestion
    path: '/profiles/useProfilesWhoMirroredPublication',
```

krzysu (changes requested):
in "publications" category, we have already a hook called `useWhoCollectedPublication` and I think your hook fits the same category, and I would also simplify the name to `useWhoMirroredPublication` to match the other hook.
wdyt?

in the results returned from the API, at least for the publicationId you used, there are many repeating profiles, what causes react `Warning: Encountered two children with the same key, 0x57` could you use solve that too?

examples/web-wagmi/src/profiles/ProfilesPage.tsx
krzysu: or in fact I would move this hook to `publications` category and rename it 
```suggestion
    path: '/publications/useWhoMirroredPublication',
```

packages/react/src/profile/index.ts
```diff
@@ -11,6 +11,7 @@ export * from './useCollectedPublications';
 export * from './useMutualFollowers';
 export * from './useUpdateProfileImage';
 export * from './useProfilesOwnedBy';
+export * from '../publication/useWhoMirroredPublication';
```
krzysu, line 14: is this necessary? I think you just export it from the publication/index.ts, not from profile folder

krzysu (approved):
I left one small comment regarding export, the rest LGTM 👍

PR #108, opened 2023-01-23
feat: add ability to collect a publication

reecejohnson:
- Adds `useCollect` hook
- Introduces client only field `collectState` to help UIs determine what actions to provide 
- Adds example in `web-wagmi`


packages/api-bindings/src/apollo/createPublicationTypePolicy.ts
```diff
@@ -43,6 +50,18 @@ export function createPublicationTypePolicy(): TypePolicy<R> {
       isOptimisticMirroredByMe(existing: boolean | undefined) {
         return existing ?? false;
       },
+
+      collectState(
+        existing: CollectState | undefined,
+        { readField }: { readField: (field: keyof R) => Readonly<unknown> | undefined },
```
reecejohnson, line 56: TypeScript is struggling the infer `readField` with the correct parameters here, I imagine because of the generic type used:
```typescript 
type R = Post | Comment | Mirror;
```
I've been able to define a type to capture what we know it could return but unfortunately having to use `as ...` on the lines below.

Any ideas?

packages/api-bindings/src/graphql/utils/publication.ts
```diff
@@ -1,41 +1,46 @@
...
   ReactionTypes,
+  TimedFeeCollectModuleSettingsFragment,
 } from '../generated';
+import { erc20Amount } from './amount';
 import { Typename, PickByTypename, JustTypename } from './types';
 
 type PublicationTypename = JustTypename<Mirror> | JustTypename<Comment> | JustTypename<Post>;
 
-export const isPostPublication = <T extends Typename<string>>(
+export function isPostPublication<T extends Typename<string>>(
```
reecejohnson, line 25: needed some consistency in this file.

they are functions 🤷

packages/api-bindings/src/apollo/createPublicationTypePolicy.ts
paulxuca:
```typescript
// Typescript not happy here
referencePolicy(
  existing: Readonly<R['referencePolicy']> | undefined,
  { readField }: { readField: ReadFieldFunction },
) {
  if (existing) return existing;

  const module = readField<ReferenceModule>('referenceModule');

  return resolveReferenceModule(module ?? null);
},
```

Saw this in my other PR as well.

.changeset/plenty-kings-sneeze.md
```diff
@@ -0,0 +1,7 @@
+---
+"@lens-protocol/api-bindings": minor
+"@lens-protocol/domain": minor
+"@lens-protocol/react": minor
```
cesarenaldi, line 4: I believe @desfero was suggesting to flag only the outer facing package.

packages/api-bindings/src/graphql/utils/publication.ts
```diff
@@ -93,8 +98,156 @@ export type PublicationOwnedByMeFragment = Overwrite<
...
+        fee: {
+          amount: erc20Amount({ from: publication.collectModule.amount }),
+          contractAddress: publication.collectModule.contractAddress,
+        },
+      };
+    case 'RevertCollectModuleSettings':
+    default:
+      never(
+        `Cannot collect publication with "${
+          publication.collectModule.__typename as string
```
cesarenaldi, line 139: is this because some times it's nullable? I think we can include it in the corresponding GQL fragment definition and it will not longer be `Maybe<string>`?

packages/api-bindings/src/apollo/createPublicationTypePolicy.ts
cesarenaldi: If I understood correctly, I think it's a mix of 2 things:
- `readField` is poorly typed to start with (in Apollo). This because it deals with normalized cache object. When using `readField` sometimes you get a so-called reference (a glorified identifier for the internal cache object) or the raw data (if it's a literal or it's an object and the cache is instructed to not normalize object of the given type via they corresponding `keyFields`). Now there is no easy way for the a generic TS type to figure out that a field pointing to an object (or array) is related to a type for which the cache normalization is turned off.
- the `TypePolicy` type we use is a patched version of the Apollo one (see `TypePolicy.ts`). It is something we partially inherited from an old project more than a year ago. I have reason to think that it does not account for the behaviour explained above. In other words it's an approximation that works in most of the cases.


Do you recognize in this anything familiar with the error you experienced?

packages/api-bindings/src/graphql/client.graphql
```diff
@@ -12,16 +13,19 @@ extend type Profile {
 extend type Post {
   hasOptimisticCollectedByMe: Boolean!
   isOptimisticMirroredByMe: Boolean!
+  collectState: CollectState!
```
cesarenaldi, line 16: Question: any reason to not have adopted the same strategy you used with `followPolicy: FollowPolicy`?

It did seems pretty analogous scenario given the number of details some collect policy requires compared to others?

packages/api-bindings/src/graphql/utils/publication.ts
```diff
@@ -93,8 +98,156 @@ export type PublicationOwnedByMeFragment = Overwrite<
...
+export function isPublicationOwnedByMe(
   publication: PublicationFragment,
-): publication is PublicationOwnedByMeFragment => {
+): publication is PublicationOwnedByMeFragment {
   return publication.profile.ownedByMe;
+}
+
+export function createCollectRequest(
+  publication: PublicationFragment,
+  activeProfile: ProfileFragment,
```
cesarenaldi, line 109: "Active profile" is a concept of the state management done by the `react` package. One can collect with any profile they own as far as they can sign the tx request (also not event signing required if they use a proxied collect action).

`collector` maybe?

packages/api-bindings/src/graphql/utils/publication.ts
reecejohnson: We always are returned the `collectModule`, this default case is saying if it isn't any of the already listed cases, throw an error when trying to create a collection request because we don't support the collect module.

I have removed the default case in favour of an exhaustive switch check here.

packages/api-bindings/src/graphql/utils/publication.ts
reecejohnson: nice idea 👍 added

packages/api-bindings/src/apollo/createPublicationTypePolicy.ts
cesarenaldi: Turns out that despite the `reafField` type although an approximation of the real thing... is enough for this use case:
image

Just by relying on the factory function return type it's enough to make it work.

I will be submitting a PR to fix this in Paul's Reference Policy PR.

I used `Post` instead of `R` (i.e. `Post | Comment | Mirror`) just to overcome another change I had to make (i.e. Mirror should not surface reference module settings).

cesarenaldi (approved):

examples/web-wagmi/src/publications/components/PublicationCard.tsx
```diff
@@ -51,13 +52,13 @@ export function CollectablePublicationCard({
         {publication.hidden ? 'This publication has been hidden' : publication.metadata.content}
       </p>
       {collectButton}
-      {'collectLimit' in publication.collectPolicy && (
+      {publication.collectPolicy.state === CollectState.COLLECT_LIMIT_REACHED && (
```
desfero, line 55: nice, read-only state machine. 
Quite helpful given we can build discriminate `switch` on the UI and cause some `tsc` errors when we add more `CollectStates`.

PR #109, opened 2023-01-23
feat: Cleanup `onError` to not handle local mutation errors (T-4290)

desfero:
- Removed local mutation errors from the global `onError` handler
- Refactored `useWalletLogin/useWalletLogout` to return an object to be in sync with other mutations 
- Updated docs at: https://docs.lens.xyz/docs/error-handling
- Updated docs at: https://docs.lens.xyz/docs/brave-trainers-repair-usewalletlogin
- Updated docs at: https://docs.lens.xyz/docs/f-usewalletlogout


examples/web-wagmi/src/profiles/UseFollowAndUnfollow.tsx
```diff
@@ -11,14 +11,12 @@ type FollowButtonProps = {
...
 };
 
-const toastNotification = (error: Error) => toast.error(error.message);
-
 function FollowButton({ profile }: FollowButtonProps) {
   const { follow, isPending: isFollowing } = useFollow({ profile });
   const { unfollow, isPending: isUnfollowing, error } = useUnfollow({ profile });
 
   useEffect(() => {
-    if (error && 'message' in error) toastNotification(error);
```
desfero, line 21: `'message' in error` caught my attention as it should not really be required when type definition is correct. 
Found the issue that `useUnfollow` was incorrectly referring to `IEquatableError` which is our internal error (strictly it's not an error as it's not extending `Error` due to type incompatibility).

cesarenaldi: > it's not extending Error due to type incompatibility

what you mean by that?

packages/domain/src/use-cases/wallets/WalletLogin.ts
```diff
@@ -50,7 +50,7 @@ export class WalletLogin {
     private readonly credentialsIssuer: ICredentialsIssuer,
     private readonly credentialsWriter: ICredentialsWriter,
     private readonly activeWalletPresenter: IActiveWalletPresenter,
-    private readonly connectionErrorPresenter: IConnectionErrorPresenter,
+    private readonly genericPresenter: IWalletLoginPresenter,
```
cesarenaldi, line 53: Naming wise, as far as this class know this is a `walletLoginPresenter`.
```suggestion
    private readonly walletLoginPresenter: IWalletLoginPresenter,
```

packages/domain/src/use-cases/wallets/ILogoutPresenter.ts
```diff
@@ -10,6 +11,6 @@ export type LogoutData = {
   logoutReason: LogoutReason;
 };
 
-export interface ILogoutPresenter {
+export interface ILogoutPresenter extends IGenericResultPresenter<void, never> {
```
cesarenaldi, line 14: Not sure about this choice. Leaving a comment about this line later the PR.

packages/react/src/wallet/adapters/LogoutPresenter.ts
```diff
@@ -1,11 +1,20 @@
...
+  extends PromiseResultPresenter<void, never>
+  implements ILogoutPresenter
+{
+  constructor(private readonly logoutHandler: LogoutHandler) {
+    super();
+  }
 
   presentLogout(data: LogoutData): void {
     this.logoutHandler(data);
+    this.present(success());
```
cesarenaldi, line 18: Seems convoluted to just reuse a `Deferred`.

The `present` method although defined as part of the public interface at the domain level is only used internally.

It seems that an implementation detail is affecting the domain definition.

I would instead define the `ILogoutPresenter` as:

```ts
type ILogoutPresenter = IGenericResultPresenter<LogoutData, never>
```

then the `LogoutPresenter` can still extend the `PromiseResultPresenter<void, never>`. It can override the `present` method and call the `logoutHandler` before invoking the `super.present(success())`.

desfero: oh wow, that's a very very good point

examples/web-wagmi/src/profiles/UseFollowAndUnfollow.tsx
desfero: the `IEquatableError` has a custom implementation of `name` that's not a broad `string`
```
export interface IEquatableError<T extends string = string, P = Narrow<T>> {
  name: P;
}
```
the `Error` interface from `lib.d.ts` marks the error as just a `string`
```
interface Error {
    name: string;
    message: string;
    stack?: string;
}
```
given that we can't just do `interface IEquatableError<...> extends Error` as that would fail on the tsc level
```
TS2430: Interface 'IEquatableError<T, P>' incorrectly extends interface 'Error'.   Types of property 'name' are incompatible.     Type 'P' is not assignable to type 'string'
```

but anyway i don't remember any place in the code where we expose the `IEquatableError` to the sdk consumers, we always wrap it with a custom one that's easy to understand.

cesarenaldi: Thank you for the explanation. Make sense.

packages/react/src/wallet/adapters/LogoutPresenter.ts
desfero: fixed

cesarenaldi (approved):

PR #110, opened 2023-01-23
feat: adds `isValidHandle` predicate

cesarenaldi:


examples/web-wagmi/src/profiles/UseCreateProfile.tsx
```diff
@@ -13,58 +13,53 @@ function ShowProfile({ handle }: { handle: string }) {
...
 export function UseCreateProfile() {
   const [handle, setHandle] = useState<string | null>(null);
-  const [submitted, setSubmitted] = useState<boolean>(false);
+  const [submitted, setSubmitted] = useState(false);
 
   const { create, error, isPending } = useCreateProfile();
 
+  const onSubmit = async (e: FormEvent) => {
+    e.preventDefault();
+    if (!handle) return;
```
desfero, line 22: nit: remember @reecejohnson added `invariant` into the `web-wagmi`

packages/api-bindings/src/graphql/utils/index.ts
```diff
@@ -1,6 +1,7 @@
 // generic utils
 export * from './amount';
 export * from './errors';
+export * from './isValidHandle';
```
desfero, line 4: Let's add `changeset`

packages/api-bindings/src/graphql/utils/isValidHandle.ts
```diff
@@ -0,0 +1,5 @@
+const validationRegex = /^[a-z](?:[a-z0-9_]{4,30})$/;
```
desfero, line 1: Previously we have been using this one
```
/(@[a-z\d-_.]{1,31})/g;
```
was there any change in the underlying protocol handle validation rules?

packages/react/src/profile/useCreateProfile.ts
```diff
@@ -29,6 +29,7 @@ export function useCreateProfile() {
 
         if (result.isFailure()) {
           setError(result.error);
+          return;
```
desfero, line 32: 😱

desfero (changes requested):

packages/api-bindings/src/graphql/utils/isValidHandle.ts
cesarenaldi: The `@` does not apply in this circumstance, also the `.` is part of the deployment suffix in Provide v1 (i.e. `.test` on test-net and `.lens` on main-net).

I am not sure if it can starts with number (maybe it can, I vaguely remember seeing few test handles) or `_`.

desfero: I remember you can have as many `.` as you can, even `pawel....pawel`

packages/api-bindings/src/graphql/utils/index.ts
desfero, line 4: on the other note which is not a part of this PR, wondering about the way to share the `isValidHandle` between `api-bindings` and this new package with just core api wrappers to avoid repeating the same regexes twice

packages/api-bindings/src/graphql/utils/isValidHandle.ts
cesarenaldi: I am receiving mixed messages about this. Will double check and submit a follow up PR to tweak the regexp.

packages/api-bindings/src/graphql/utils/index.ts
cesarenaldi: good point... I put it here cause it's a validation very related to the API. Not sure if the low-level client will need such helper.... fair to assume so though. Let's see when we get there.

desfero (approved):

PR #114, opened 2023-01-24
bug: followers following queries are missing field policy that allows infinite scroll

desfero:
- fixed `useProfileFollowers` to support infinite scroll
- fixed `useProfileFollowing` to support infinite scroll
- fixed `useProfilePublicationRevenue` to work and support infinite scroll
- Refactored example app profile selectors to have less code and be less error-prone (e.g. the `default` string returned from callback where id or handle was expected)

### DIFF
--- a/examples/web-wagmi/src/profiles/UseMutualFollowers.tsx
+++ b/examples/web-wagmi/src/profiles/UseMutualFollowers.tsx
@@ -49,8 +49,8 @@ export function UseMutualFollowers() {
 
       <article>
         <p>Select two profiles to see their mutual followers:</p>
-        <SelectProfile onProfileSelected={(profile) => setSelectedProfileOne(profile)} />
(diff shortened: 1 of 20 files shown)


examples/web-wagmi/src/profiles/UseMutualFollowers.tsx
```diff
@@ -49,14 +49,14 @@ export function UseMutualFollowers() {
 
       <article>
         <p>Select two profiles to see their mutual followers:</p>
-        <SelectProfileId onProfileSelected={(profile) => setSelectedProfileOne(profile)} />
-        <SelectProfileId onProfileSelected={(profile) => setSelectedProfileTwo(profile)} />
+        <SelectProfile onProfileSelected={(p) => setSelectedProfileOne(p)} />
```
cesarenaldi, line 52: ah, I pretty much thought about the same thing while I was refactoring these tests to add error handling. Thanks!

cesarenaldi (approved):

PR #119, opened 2023-01-25
feat: add additional utility methods to Amount

vadoff:
* add subtraction and division methods
* add tests

### DIFF
--- a/packages/shared-kernel/src/crypto/__tests__/Amount.spec.ts
+++ b/packages/shared-kernel/src/crypto/__tests__/Amount.spec.ts
@@ -33,6 +33,23 @@ describe('Given the Amount type', () => {
     });
   });
 
+  describe('when subtracting an Amount by another', () => {
+    it('should return a new Amount that is the difference of the 2 amounts', () => {
+      const amount1 = Amount.usd(2);
+      const amount2 = Amount.usd(3);
+
+      const expected = Amount.usd(-1);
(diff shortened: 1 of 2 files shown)


vadoff:
I was considering adding an additional invariant to the div method to check for division by 0, but decided against as the underlying method supports it.

packages/shared-kernel/src/crypto/__tests__/Amount.spec.ts
```diff
@@ -44,6 +61,23 @@ describe('Given the Amount type', () => {
     });
   });
 
+  describe('when dividing an Amount by another', () => {
+    it('should return the expected Amount', () => {
+      const amount1 = Amount.usd(42);
+      const amount2 = Amount.usd(2);
+
+      const expected = Amount.usd(21);
+      expect(amount1.div(amount2)).toEqual(expected);
```
cesarenaldi, line 70: Would division (like multiplication) involve a scalar value rather than another amount?

In other words:
```
amount1.div(2)
```

In that case the invariant check is not even needed.

The module defines an `AmountValue` that is the scalar type we could use (see `mul` method for example).

vadoff: I thought about it, and it made a lot of sense to support scalar division, so I changed the method to support it.

cesarenaldi (approved):

PR #120, opened 2023-01-26
`useApproveModule` Hook

JoaquinBattilana:
Adds `useApproveModule` hook to the `react` package


examples/web-wagmi/src/App.tsx
```diff
@@ -57,7 +58,10 @@ import { UseProfileFollowRevenue } from './revenue/UseProfileFollowRevenue';
 import { UseProfilePublicationRevenue } from './revenue/UseProfilePublicationRevenue';
 import { UsePublicationRevenue } from './revenue/UsePublicationRevenue';
 
-const { provider, webSocketProvider } = configureChains([polygon, optimism], [publicProvider()]);
```
JoaquinBattilana, line 60: was this a typo? I needed `mumbai` provider for the user to be able to sign in the example app, is there a reason why we need `optimism` ?

packages/blockchain-bindings/src/contracts.ts
```diff
@@ -3,8 +3,8 @@ import { EthereumAddress } from '@lens-protocol/shared-kernel';
 import { Contract } from 'ethers';
 
 import erc20Abi from './abi/erc-20.json';
-import { ERC20 } from './types/ERC20';
+import { IErc20Contract } from './types';
 
 export function erc20(address: EthereumAddress, provider: Provider) {
-  return new Contract(address, erc20Abi, provider) as unknown as ERC20;
+  return new Contract(address, erc20Abi, provider) as unknown as IErc20Contract & Contract;
```
JoaquinBattilana, line 9: I had to change the type from the generated to a custom one because there are two problems with the generated types:
- The `overrides` type of the write functions doesn't have all the options that we use, like `maxFeePerGas` or `maxPriorityFeePerGas`
- The type doesn't have `estimateGas` and `populateTransaction` methods that `ethers` exports in their contracts, and they don't export a generic or separated write/read methods types for me to be able to extend the generated type.

packages/react/src/transactions/useApproveModule.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  function isApproveTransaction(
+    transaction: TransactionState<SupportedTransactionRequest>,
+  ): transaction is TransactionState<TokenAllowanceRequest> {
+    return (
+      transaction.request.kind === TransactionKind.APPROVE_MODULE &&
+      transaction.request.spender === spender &&
+      transaction.request.amount.asset === amount.asset
+    );
+  }
```
JoaquinBattilana, line 47: I feel like we need to debate this predicate, if you test the example app you'll see what i mean. Let's say i have a UI that shows a 2 publications, both of them with a paid collectable module and the same asset. If the user approves the first one  with an amount that is not enough for the second one, the second one will always appear as "approved" because we're not checking the quantity here. Also, if we check the quantity is the same if the two publications use the same amount. Should we check by `id` ?

examples/web-wagmi/src/misc/UseApproveModule.tsx
```diff
@@ -0,0 +1,86 @@
...
+  const handleAmountChange = (e: ChangeEvent<HTMLInputElement>) => {
+    if (!e.target.value) setAmount(0);
+    else setAmount(e.target.valueAsNumber);
+  };
+
+  const handleCurrencyChange = (e: ChangeEvent<HTMLSelectElement>) =>
+    setSelectedCurrency(Number(e.target.value));
+
+  const handleModuleChange = (e: ChangeEvent<HTMLSelectElement>) =>
+    setSelectedModule(Number(e.target.value));
```
desfero, line 37: seems not working as you don't set `value` as number so it's using `moduleName` which is a string so it always set's the `Nan` as `selectedModule`

packages/blockchain-bindings/src/contracts.ts
desfero: I can see that the generator has something called `ContractContext` with `estimateGas` and `populateTransaction` (they are not fully typed as they are using indexed types).
But still, we lack the `maxFeePerGas` and `maxPriorityFeePerGas` from EIP1559.
@joshstevens19 is that something that we can easily add to your codegen to not added custom code into the sdk?

packages/blockchain-bindings/src/utils.ts
```diff
@@ -23,3 +24,7 @@ export function isTheSameAddress(address1: EthereumAddress, address2: EthereumAd
   // lowercase before comparing to avoid comparing checksum address with a normal one
   return address1.toLowerCase() === address2.toLowerCase();
 }
+
+export function bigNumber(from: CryptoAmount) {
```
desfero, line 28: can we add some tests for this please?

packages/react/src/transactions/adapters/useApproveModuleController.ts
```diff
@@ -0,0 +1,34 @@
...
+    const approveTransactionGateway = new ApproveTransactionGateway(providerFactory, gasEstimator);
+
+    const tokenAllowance = new TokenAllowance(
+      activeWallet,
+      approveTransactionGateway,
+      presenter,
+      transactionQueue,
+    );
+
+    void tokenAllowance.increaseAllowance(request);
```
desfero, line 30: can we `await` for consistency?

packages/react/src/transactions/useApproveModule.ts
desfero: We could for now assume that in normal circumstances the frontend should wait before trying to approve a new module (maybe we can even throw here an error if trying to approve more than one)
I would personally also remove `isApproved` flag as it may be confusing for the users. It doesn't say whether the given module is approved it just checks the transaction history to see if you made the transaction to approve the given module.

packages/shared-kernel/src/index.ts
```diff
@@ -5,6 +5,8 @@ export * from './crypto/Amount';
 export * from './crypto/Asset';
 export * from './crypto/ChainType';
 export * from './crypto/types';
+export * from './crypto/IGasEstimate';
+export * from './crypto/__helpers__/mocks';
```
desfero, line 9: mocks should be exposed from `@lens-protocol/shared-kernel/mocks`

desfero (changes requested):

package.json
```diff
@@ -36,8 +36,12 @@
         "rollup"
       ],
       "allowedVersions": {
-        "@solana/web3.js": "1.52.0"
+        "@solana/web3.js": "1.52.0",
```
desfero, line 39: any reason changing `allowedVersions` and `overrides`? Any bugs that we can track to revert that back at some point?

cesarenaldi: this is to silence the pnpm yellow warning you get when it has to make a decision on a version it's not fully happy with. Unfortunately it only works on top level `package.json` even though it's caused by one of the subpackage.

I think the `@solana/web3.js` was caused by a dependency in one of the example, not 100% sure might be `wagmi`.

PR #122, opened 2023-01-27
fix: logout error between wagmi and lens-sdk

reecejohnson:
There a three fixes here: 
1. There was an issue where a user could press login and their wallet would be connected with `wagmi` but not succeed at logging in with Lens. This caused a strange state where a user could then not attempt to log in again as the login function was wrapped in a check on whether there was no current account connected to `wagmi`. The fix was to check if there was a connected address and then await the async disconnect call from `wagmi` before proceeding with the rest of the log in process.
2. On logout, the Lens `logout` function is invoked and then a synchronous `wagmi` disconnect call was made. It makes sense to use the async disconnect method and ensure we await it, so to avoid any race conditions between logging out and logging back in.
3. Changing the way we retrieve the active profile. Beforehand, there was a "dangling subscription" watching changes to the apollo cache that was triggering an update to the older incarnation of the `useActiveProfile` reactive var, causing the intermittent persisted profile - even when logged out. Now we use the `useActiveProfile` to only store the profile identifiers and use these as parameters for a get profile by id apollo query which keeps us up to date with the cache by default. Less fancy but less code.


krzysu:
the same code (with the bug) is used in the example in docs, could you also update the docs? https://docs.lens.xyz/docs/use-wallet-login

desfero (approved):

krzysu (approved):

PR #123, opened 2023-01-27
feat: add `client` package (T-4339)

krzysu:
- new package setup
- graphql codegen setup
- LensClient
- simple authentication - follow-up in a separate ticket
- simple read operation example - profile.fetch(), profile.fetchAll()
- simple write operation example after authentication - reaction.add() 
- long TODO list as an overview of the power of this sdk


packages/client/tsconfig.base.json
```diff
@@ -0,0 +1,23 @@
...
+    "esModuleInterop": true,
+    "forceConsistentCasingInFileNames": true,
+    "importHelpers": true,
+    "lib": ["ESNext"],
+    "module": "ESNext",
+    "moduleResolution": "Node",
+    "noFallthroughCasesInSwitch": true,
+    "noImplicitReturns": true,
+    "noUnusedLocals": false,
+    "noUnusedParameters": false,
```
krzysu, line 15: I had to turn off `noUnusedLocals` and `noUnusedParameters` otherwise ts complains about the `fragments.generated` file, as it still has empty SDK generated. any idea how to disable ts check for a single generated file?

packages/client/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Examples
+
+For all exaples below we require the same setup, we need a LensClient instance and an ethers.js Wallet as a signer.
+
+```ts
+import LensClient, { mumbai } from '@lens-protocol/client';
+import { Wallet } from 'ethers';
+
+const lensClient = LensClient.init(mumbai);
```
krzysu, line 15: as there is only one way to init the client, this could also just be

```
const lensClient = new LensClient(mumbai);
```

packages/client/README.md
```diff
@@ -0,0 +1,69 @@
+# `@lens-protocol/client`
+
+This package contains low level client for GraphQL Lens API.
+
+**Its interface will change without notice, use it at your own risk.**
```
cesarenaldi, line 5: I think this paragraph comes from the README file of an internal package.

Given this is a publica facing package something akin to what we wrote in `packages/react/README.md` is more appropriate:

```suggestion
> **Warning**
>
> The Lens SDK is still in its initial development phase. Anything MAY change at any time.
> This is a Developer Preview aimed primarily at existing integrators so to gather [early feedback](https://github.com/lens-protocol/lens-sdk/discussions/48).
```

packages/client/TODO.md
```diff
@@ -0,0 +1,160 @@
...
+- [H] helper (no API interaction)
+
+## Environments
+
+- polygon, mumbai, mumbaiSandbox
+
+## Client and Authentication
+
+- [x] LensClient.init(environment)
+- [x] [Q] client.generateChallenge(address)
```
reecejohnson, line 18: what's the reasoning for not "namespacing" these methods similar to the others,? 

for example:

`client.auth.generateChallenge()`

packages/client/README.md
```diff
@@ -0,0 +1,69 @@
+# `@lens-protocol/client`
+
+This package contains low level client for GraphQL Lens API.
```
cesarenaldi, line 3:
```suggestion
The official framework-agnostic Lens API Client.

This package enables you to interact with the Lens API via a type safe interface that abstracts away some of the GraphQL intricacies.
```

packages/client/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+**Its interface will change without notice, use it at your own risk.**
+
+## Examples
```
cesarenaldi, line 7: Good to keep the conversation going... but the proper place to document this will be on each API page as alternative to the current GQL examples.

Maybe over time we can make this leaner and reference the official docs.

packages/client/TODO.md
```diff
@@ -0,0 +1,160 @@
...
+- authentication required error thrown on a group or on a method call?
+- allow for custom gql queries or compose query from subqueries (like profile base, profile interests, profile onChainIdentity etc.) (is that possible?)
+
+```ts
+client.profile.fetch(request, gql: {
+    ...baseProfileQuery,
+    ...profileInterestsQuery,
+    ...profileOnChainIdentityQuery
+})
+```
```
cesarenaldi, line 160: Wow, thank you for doing this analysis. It's a long list!

packages/client/codegen-api.yml
```diff
@@ -0,0 +1,58 @@
+overwrite: true
+hooks:
+  afterAllFileWrite:
+    - prettier --write
+watch: false
+config:
+  defaultScalarType: unknown
```
cesarenaldi, line 7: Maybe we can take the opportunity and use [`strictScalars: true`](https://the-guild.dev/graphql/codegen/plugins/typescript/typescript#strictscalars) instead so to detect scenarios where `unknown` surfaces without us knowing.

packages/client/src/LensClient.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+  return client;
+};
+
+describe(`Given a LensClient configured for testnet and a wallet`, () => {
+  const wallet = new Wallet(
+    'ca9a3a3d4026e6228713e683a9c45ef65a538b2f9336813bd597f5effa38668d', // found on reddit
+  );
+
+  describe(`when authenticating with wallet signature`, () => {
+    it(`authenticates with success`, async () => {
```
cesarenaldi, line 35: nit: can we keep test description consistent with the rest of the codebase by using: `it('should ...`

packages/client/src/LensClient.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+  const client = LensClient.init(mumbai);
+  const address = await wallet.getAddress();
+  const challenge = await client.generateChallenge(address);
+  const signature = await wallet.signMessage(challenge);
+
+  credentials = await client.authenticate(address, signature);
+  return credentials;
+};
+
+const getAuthenticatedClient = async (wallet: Wallet) => {
```
cesarenaldi, line 19: nit: I would suggest traditional function declaration for consistency with the rest of the codebase.

packages/client/src/LensClient.spec.ts
```diff
@@ -0,0 +1,99 @@
+import { Wallet } from 'ethers';
+
+import LensClient, { Credentials, mumbai, ReactionTypes } from '.';
+
+let credentials: Credentials;
+const getCredentials = async (wallet: Wallet) => {
+  if (credentials) {
+    return credentials;
+  }
+  const client = LensClient.init(mumbai);
```
cesarenaldi, line 10: Am I correct thinking this test suite hits the real API?

Part of me like the idea of testing against the real schema... but maybe we should consider to limit these to run on CI as part of a smoke tests suite? Thoughts?

packages/client/src/LensClient.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+
+      const { accessToken, refreshToken } = await client.refreshCredentials(storedRefreshToken);
+
+      expect(accessToken.length).toBeGreaterThan(1);
+      expect(refreshToken.length).toBeGreaterThan(1);
+    });
+  });
+
+  describe(`when checking if access token is valid`, () => {
+    it(`returns true`, async () => {
```
cesarenaldi, line 56: Even after adopting a gherkin-line syntax, tests written like this one fails to document the desired behaviour:
```
Given a LensClient configured for testnet and a wallet
when checking if access token is valid
[it] should return true
```

If we are to treat the test description as a specification, the specification above would lead to an odd result.

```
isAccessTokenValid(_: string) {
  return true;
}
```

Probably the intent of this specific test is not about the actual underlying `isAccessTokenValid` details, that seems more appropriate form `Authentication.spec.ts`, is the intent here to be an integration test of some sort? Maybe we can try communicate that.

packages/client/src/LensClient.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+      const address = await wallet.getAddress();
+      const myProfile = await client.profile.fetchDefault(address);
+
+      if (!myProfile) {
+        return;
+      }
+
+      expect(() => client.reaction).toThrow('Not authenticated');
+    });
+  });
```
cesarenaldi, line 98: How do you see this test suite growing as we add more "modules" to the `LensClient`?

packages/client/src/authentication/Authentication.ts
```diff
@@ -0,0 +1,43 @@
...
+import { GraphQLClient } from 'graphql-request';
+
+import { getSdk, Sdk } from './graphql/auth.generated';
+
+export type Credentials = { accessToken: string; refreshToken: string };
+
+export class Authentication {
+  private readonly sdk: Sdk;
+
+  constructor(backendUrl: string) {
```
cesarenaldi, line 11: nit: for consistency..

I think throughout the codebase we use `backendUrl` to identify the URL of the API. Then the API includes a `/graphql` endpoint and other endpoints that are not GQL.. we will have more for direct file uploads and so on.

What if we rename all these as:
```suggestion
  constructor(gqlEndpoint: string) {
```
and redefine `Environment` as:
```
type Environment = {
  url: string;
  
  get gqlEndpoint(): string
};
```

or simpler with a class as:
```ts
class Environment = {
  constructor(private url: string) {}
  
  get gqlEndpoint() {
    return `${this.url}/graphql`
  }
};
```

Thoughts?

packages/client/src/authentication/Authentication.ts
```diff
@@ -0,0 +1,43 @@
...
+  async verify(accessToken: string): Promise<boolean> {
+    const result = await this.sdk.AuthVerify({ accessToken });
+
+    return result.data.result;
+  }
+
+  async authenticate(address: string, signature: string): Promise<Credentials> {
+    const result = await this.sdk.AuthAuthenticate({ address, signature });
+
+    invariant(result.data, 'Not able to generate credentials. Credentials data not found.');
```
cesarenaldi, line 31: I recently addressed this in the `@lens-protocol/react` SDK. That implementation if very ApolloClient specific, how to errors propagates from the response to the call site via `GraphQLClient` and the generated code?

packages/client/src/profile/Profile.ts
```diff
@@ -0,0 +1,35 @@
...
+  ): Promise<ProfileFragment | null> {
+    const result = await this.sdk.Profile({
+      request,
+      observerId,
+    });
+
+    return result.data.result;
+  }
+
+  async fetchDefault(address: string, observerId?: string): Promise<ProfileFragment | null> {
```
cesarenaldi, line 27: The reason this edge exist is due to the fact Wallets-follow-Profiles in Lens Protocol v1.

It's very little publicized and underutilized.

In Lens Profile v2 Profiles will follow Profiles, I think we should omit it already from the SDK.

packages/client/src/consts/environments.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export const polygon: Environment = {
+  url: 'https://api.lens.dev',
+};
+
+export const mumbai: Environment = {
+  url: 'https://api-mumbai.lens.dev',
+};
+
+export const mumbaiSandbox: Environment = {
```
cesarenaldi, line 13: Although it exists is mostly used by us. I would omit it from the public interface. We might use it for integration tests instead (see my other comment), in that case we can define a dedicated `Environment` instance part of the tests support code.

cesarenaldi:
Amazing effort. I like a lot how you dropped the need for `fromCredentials` static factory function.

packages/client/TODO.md
krzysu: we decided to merge LensClient and Auth responsibilities as auth is an essential first step before using more advanced features of the API, and its usage/lifecycle is different than other groups/endpoints (ie. profile or publication).

packages/client/README.md
krzysu: I know, I just wanted to keep it somewhere for now, till we start writing proper docs, it also helps to get a feeling for the SDK usage. We will remove it when sdk gets the proper shape.

packages/client/src/LensClient.spec.ts
krzysu: I guess LensClient spec will be only checking if a module is available like

```
expect(client.profile).toBeInstanceOf(Profile);
```
but the details of the `Profile` class will be tested in `Profile.spec.ts`

packages/client/src/LensClient.spec.ts
krzysu: this is testing if `isAccessTokenValid` is correctly exposed directly from LensClient, I will improve test description

packages/client/src/profile/Profile.ts
krzysu: so as a dev, how do I get my own profile after connecting a wallet? which query to use?

packages/client/src/authentication/Authentication.ts
krzysu: is it fine if I just remove these `invariant` checks for now, and proper error handling we will add a bit later?

packages/client/codegen-api.yml
krzysu: I did it! but I assumed all of the unknown types are string, not sure how to check that

packages/client/tsconfig.base.json
krzysu: @cesarenaldi any ideas on how to solve the above?

packages/client/README.md
krzysu: @cesarenaldi @reecejohnson @desfero any opinions on the constructor?

packages/client/src/consts/environments.ts
krzysu: I changed tests to run on mumbaiSandbox and I'm not exporting this var from the package anymore

packages/client/src/authentication/Authentication.ts
krzysu: I did it as you suggested, thanks for the idea!

packages/client/tsconfig.base.json
desfero: We could try to model the `codegen` config in a way that it doesn't generate sdk for `fragment.graphql`

desfero (changes requested):
Looks really good, found some small nitpicks that we need to discuss

packages/client/README.md
desfero: I like that we have some initial setup in the README showcasing the API, we should do the same for `@lens-protocol/react`

packages/client/TODO.md
desfero: Enormous effort 🥇

packages/client/codegen-api.yml
cesarenaldi: Scalars in general are transmitted over the wire as strings... but this is about codegen config. Basically for any scalar that is not specified in the list below it will assume it's TS `unknown` type.

`strictScalars: true` makes the codegen step fails until all scalar types are specified.

packages/client/src/LensClient.spec.ts
cesarenaldi: Sounds good.

packages/client/src/authentication/Authentication.ts
cesarenaldi: > what about someone having some proxies that rewrite url to be just `foo.xyz` (or even `foo.xyz/lens-api)?

not sure I follow @desfero , can you please clarify the scenario?

packages/client/src/authentication/Authentication.ts
cesarenaldi: Sure, do you mind keep track of looking into cohesive approach to error handling with a dedicated ticket?

krzysu: ticket created https://avara.height.app/lens-sdk/T-5590

packages/client/src/profile/Profile.ts
krzysu: sounds good, I will do the same here and remove `defaultProfile` query from the scope of this sdk

packages/client/README.md
cesarenaldi: no strong opinion on constructor vs. static factory function now that we only have one. I agree with @desfero to make the argument and object could be beneficial if we are to add more non-env related config.

packages/client/README.md
```diff
@@ -0,0 +1,74 @@
...
+```ts
+const storedCredentials = {
+  accessToken: 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXV...',
+  refreshToken: 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXV...',
+};
+
+const isValid = await lensClient.isAccessTokenValid(storedCredentials.accessToken);
+
+if (!isValid) {
+  const { accessToken, refreshToken } = await lensClient.refreshCredentials(
```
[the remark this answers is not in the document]
cesarenaldi: I was initially advocating for this as well. One caveat to notice is that the consumer is responsible for storing these tokens, so somehow they are exposed about the details of`ICredentials` structure cause they either simply assert with it: `JSON.parse(serialized) as ICredentials` OR they use proper parsing (e.g. via zod).

I noticed a small mistake in the current "resume from stored tokens flow".

Let's say the consumer app resumes in a fresh thread from stored credentials (new tab for a SPA app, new process for a NodeJS app, hard refresh on a web app)... what happen if the `lensClient.isAccessTokenValid(accessToken)` returns true?

How does the `LensClient` can populate the `credentials` private member?

For context, the whole conversation about 2 static methods (now being just `init`) was part of a brainstorming session between me and @krzysu. The second method would have allowed to create a `LensClient` out of stored credentials.


A small snippet of code from that session to illustrate the point.
```ts

// this could be opaque to the end user
type Credentials = {
  accessToken: string
  refreshToken: string
  environment: Environment
}

// an example usage on a web app
function createLensClient(): LensClient {
  const serialized = localStorage.get('credentials') ?? null
  
  if (serialized) {
    const credentials = JSON.parse(serialized) // overly simplified to reduce noise
    return LensClient.fromCredentials(credentials)
  }

  return LensClient.init(environments.production)
}
```

The `Credentials` type was not 1:1 with the returned value from the API call, but somewhat an augmented object that was also encapsulating the environment details for which those credentials are valid. In the end they won't work in any other environment and that was a cheap solution to avoid possible mismatch between the specified environment details and stored credentials, but this is marginal for this conversation.

Now, I am not advocating to go back to this approach necessarily. I think we have good potential with current "single `init` factory method (or simply just `LensClient` constructor as it was suggested by @krzysu in a comment above).

An interesting alternative, which is reminiscent of the way we handle storing of auth tokens in the `@lens-protocol/react` package, is to just have the consumer deal with the Refresh Token.

An advantage of that is that not being a "complex structure" but just a string... it can be properly dealt with as opaque content. No need for TS assertion after recovering it from storage nor parsing, it's just an opaque string.


What does it mean for the initial authentication? Not much TBH
```ts
const token = await client.authenticate(address, signature)
```
Consumer responsibility is to just store it, they can ignore it's a JWT, they can ignore the fact it's the Refresh Token... it's all opaque to them. Easy.

On a start/restart from stored credentials (keep using the web app example above but can be adapted to other storage solutions):
```ts
const token = localStorage.get('token') ?? null

if (token) {
  const newToken = await client.refreshCredentials(token)
  
  localStorage.set(token, newToken)
}
// continue as usual
```

This approach promotes that each thread gets a fresh copy of Access Token... this will reduce the chances for the token to expire while in use and probably for a good class of applications interested in using the `LensClient` they might not even care much about it expiring as the expiry time could well be outside of their normal operating timeframe (think about the request lifecycle from a request handler on a server app).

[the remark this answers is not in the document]
cesarenaldi: As Access Token is not longer exposed... the `isAccessTokenValid` would need to change into just a check on an internal, that could be used to either inform the consumer it's time to try a `refreshCredentials` and if it also fails a brand new re-authentication.

```ts
const isAuthenticated = await client.isAuthenticated()
```

packages/client/src/LensClient.ts
```diff
@@ -0,0 +1,55 @@
...
+  async generateChallenge(address: string): Promise<string> {
+    return this.authentication.challenge(address);
+  }
+
+  async isAccessTokenValid(accessToken: string): Promise<boolean> {
+    return this.authentication.verify(accessToken);
+  }
+
+  async authenticate(address: string, signature: string): Promise<Credentials> {
+    this.credentials = await this.authentication.authenticate(address, signature);
```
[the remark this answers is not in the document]
cesarenaldi: good catch, see my comment here: https://github.com/lens-protocol/lens-sdk/pull/123#discussion_r1095501965

packages/client/src/LensClient.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  get profile(): Profile {
+    return new Profile(this.environment.url);
+  }
+
+  get reaction(): Reaction {
+    if (!this.credentials) {
+      throw new Error('Not authenticated');
+    }
+    return new Reaction(this.environment.url, this.credentials.accessToken);
```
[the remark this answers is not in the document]
cesarenaldi: I think there could be some options here:
- pass a reference to the `LensClient` which will need to implement a privileged interface to access internal credentials
- pass a callback (works, just feel old-school)
- make `credentials` mutable or observable

In all cases we need to account for our initial desire to let an advanced user being able to use just the submodules of the SDK they want:
```ts
import { Authentication, Reaction } from '@lens-protocol/client';
```

Is this still important?

packages/client/src/LensClient.ts
```diff
@@ -0,0 +1,55 @@
...
+    return this.authentication.verify(accessToken);
+  }
+
+  async authenticate(address: string, signature: string): Promise<Credentials> {
+    this.credentials = await this.authentication.authenticate(address, signature);
+    return this.credentials;
+  }
+
+  async refreshCredentials(refreshToken?: string): Promise<Credentials> {
+    const refreshTokenToUse = refreshToken || this.credentials?.refreshToken;
```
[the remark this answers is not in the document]
cesarenaldi: I agree to promote a single way to handle things until we have use cases to support.

packages/client/package.json
```diff
@@ -0,0 +1,65 @@
+{
+  "name": "@lens-protocol/client",
+  "version": "0.1.0",
+  "description": "Low level Lens API client",
+  "repository": {
+    "directory": "packages/client",
+    "type": "git",
+    "url": "git://github.com/lens-protocol/lens-sdk.git"
+  },
+  "type": "module",
```
cesarenaldi, line 10: should we flag it `private: true` for now?

packages/client/TODO.md
```diff
@@ -0,0 +1,182 @@
...
+
+## Feed
+
+- [PQ] client.feed.fetch({ profileId })
+
+## Other considerations
+
+- paginated response helper
+- more helpers like `isValidHandle`
+- how to split a group that has both authenticated and public requests?
```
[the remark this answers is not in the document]
cesarenaldi: I reckon authenticated vs. not-authenticated should be a reason to split into 2 separate sub-modules.

Also bear in mind some queries require authentication but conceptually they might not need to. I know about `enabledCurrencies` but there might be others. Maybe we can ask to remove such constraint at the API level.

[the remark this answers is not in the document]
krzysu: I'm thinking more about checking if the client is authenticated on a method call level, not a module level. That makes the interface nicer and more consistent.

look at Reactions (from TODO)

```
[A] authenticated
[P] paginated
[M] mutation
[Q] query
 
- [AM] client.reactions.add(request)
- [AM] client.reactions.remove(request)
- [PQ] client.reactions.toPublication({ publicationId })
```

it would be confusing to have
```
client.authReactions.add()
// and
client.publicReactions.toPublication()
```

nicer approach (if not auth)
```
client.reactions.toPublication() // all good
client.reactions.add(request) // throw NotAuthenticatedError()

// then
client.refreshToken()
client.reactions.add(request)  // all good now
```

packages/client/src/authentication/Authentication.ts
desfero: @cesarenaldi the interface to provide the `environment` is public. We can't stop anyone from having a proxy in between `client` and our `graphql api` (for .e.g someone wants to have custom metrics around perfomance, api usage by using SAAS similar to Stellate). The proxy may be configured in a way that doesn't expose graphql under `<url>/graphql` so we should allow passing the full graphql url without appending anything on our side.
Any thoughts? Does that make sense anyway?

packages/client/TODO.md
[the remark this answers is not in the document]
desfero: thinking out loud, we had a talk with @cesarenaldi to expose our `Result` "monad" to the external world to make it more typesafe, maybe we should follow the same convention for the sdk and return `Result` from methods? 
Thoughts?

packages/client/README.md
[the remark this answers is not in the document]
desfero: Pretty like the way to not expose token and use the same patter with `IStorageProvider` that we already make our sdk users familiar with

packages/client/tsconfig.base.json
krzysu: I tried that, but we have two plugins here that are not aware of each other:
- `typescript-graphql-request` - https://the-guild.dev/graphql/codegen/plugins/typescript/typescript-graphql-request
- `near-operation-file-preset` - https://the-guild.dev/graphql/codegen/plugins/presets/near-operation-file-preset

I didn't find anything in the docs to skip SDK generation for a selected file.
A solution could be to add some dummy query to `fragments.graphql` to generate SDK we don't use but that satisfies the `ts` checks.

krzysu: in fact dummy query worked fine

packages/client/src/LensClient.spec.ts
krzysu: there is a ticket to "Run only affected packages lint/tests on CI" https://avara.height.app/inbox/T-5453 and initially there are hardly any tests here so I'd worry about that later

packages/client/src/authentication/Authentication.ts
krzysu: I think it makes sense so I won't append anything to the API URL

packages/client/README.md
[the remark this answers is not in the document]
krzysu: after a long discussion, we settled on an approach to managing the token lifecycle and storage inside the client, ticket was created, it is now outside of the scope of this PR
https://avara.height.app/T-5607
I will leave these two methods in this PR

```
client.generateChallenge(address)
client.authenticate(address, signature) // returns nothing
```

packages/client/src/LensClient.ts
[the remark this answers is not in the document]
krzysu: https://github.com/lens-protocol/lens-sdk/pull/123#discussion_r1095813716

but I don't know how that will work with externally managed submodules, let's leave it for later

packages/client/src/LensClient.ts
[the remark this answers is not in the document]
krzysu: I removed `refreshCredentials` from the scope of this PR

packages/client/TODO.md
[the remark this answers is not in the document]
krzysu: I created a ticket to think about it later, for now, I moved throwing of NotAuthenticatedError to a method call level, see Reactions.ts
https://avara.height.app/lens-sdk/T-5676

krzysu:
I think I addressed all the feedback, and the PR is in a good condition to be merged, let me know if I missed anything. I also created a few tickets with follow-ups.

krzysu:
I lost this comment somewhere in resolved conversations

> @cesarenaldi Just observing that this not-exposing access token could have some implications with the idea that an advanced users could use the individual parts of the SDK

The nice thing is that this package is designed so that all submodules **must** work independently and the proof for that is submodule tests, see ie. `packages/client/src/reactions/Reactions.spec.ts`
the same way any advanced user can import a submodule class and init it as we do in tests.

cesarenaldi (approved):

desfero (approved):

PR #126, opened 2023-01-30
Small cleanup of the web-wagmi example app

desfero:


examples/web-wagmi/src/App.tsx
```diff
@@ -68,7 +68,9 @@ const client = createClient({
 const lensConfig: LensConfig = {
   bindings: wagmiBindings(),
   environment: staging,
-  sources: [sources.lenster, sources.orb, 'any-other-app-id'],
+  sources: [],
+  // or narrow to the one you are interested in
+  // sources: [sources.lenster, sources.orb, 'any-other-app-id'],
```
desfero, line 73: can be confusing, especially when you compare the result with `lenster`.
Or when you add a comment and then go to the comments section and publication is not there given we don't add any `appId` in the metadata.

examples/web-wagmi/src/App.tsx
```diff
@@ -57,7 +57,7 @@ import { UseProfileFollowRevenue } from './revenue/UseProfileFollowRevenue';
 import { UseProfilePublicationRevenue } from './revenue/UseProfilePublicationRevenue';
 import { UsePublicationRevenue } from './revenue/UsePublicationRevenue';
 
-const { provider, webSocketProvider } = configureChains([polygon, optimism], [publicProvider()]);
+const { provider, webSocketProvider } = configureChains([polygon], [publicProvider()]);
```
desfero, line 60: `optimism` is a leftover

krzysu (approved):

examples/web-wagmi/src/App.tsx
krzysu: shall we use `polygonMumbai` then?

Joaquin had similar finding in his PR https://github.com/lens-protocol/lens-sdk/pull/120/files

desfero: Good point, merging this one than and we will also keep the Joaquin changes in the #120

PR #130, opened 2023-01-31
useRecentTransactions

paulxuca:
@cesarenaldi I think I might have misunderstood the ticket and ended up trying to implement what was mentioned on the ticket -- https://avara.height.app/T-1416 please let me know if this is what the functionality of the hook is supposed to fulfil before I continue any further 🙏


cesarenaldi:
@paulxuca this ticket was more open-ended than others. At broad strokes what you did is sound... we might want to discuss together where to move this forward. I can give you some background around why this hook was used in the past and we can decide how to make it work in the SDK.

packages/react/src/transactions/useRecentTransactions.ts
```diff
@@ -0,0 +1,19 @@
...
+
+  const transactions = useRecentTransactionsVar();
+
+  return {
+    clearCompleted() {
+      transactionQueue.clearRecent();
+    },
+    transactions: uniqBy(transactions, (t) => JSON.stringify(t.request)).map((transaction) => ({
+      ...transaction,
+      completed: transaction.status === TxStatus.SETTLED,
```
paulxuca, line 16: Not sure if we want to expose the `TxStatus` enum since it seems a bit low level for someone consuming the react SDK.

packages/react/src/transactions/useRecentTransactions.ts
```diff
@@ -0,0 +1,19 @@
...
+export function useRecentTransactions() {
+  const { transactionQueue } = useSharedDependencies();
+
+  const transactions = useRecentTransactionsVar();
+
+  return {
+    clearCompleted() {
+      transactionQueue.clearRecent();
+    },
+    transactions: uniqBy(transactions, (t) => JSON.stringify(t.request)).map((transaction) => ({
```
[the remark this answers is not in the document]
cesarenaldi: mmmh, can you clarify this? In what circumstances? and was this the `TransactionQueue` state or the UI derived state managed by the `TransactionQueuePresenter`? You might might have found a bug.

packages/react/src/transactions/useRecentTransactions.ts
cesarenaldi: I see you point, one of the usages we were making of this data was to show the difference between: "pending", "failed" and "settled" (successfully).

`TxStatus` here is an enum designed for UI state reason, I personally think it's ok to surface at least `FAILED` and `SETTLED` options. Not 100% sure about `BROADCASTING` and `MINING` myself.

Soon we will add a feature (which you might have heard under the name DA - Data Availability) that might stir these concepts a little. Let's review this idea after that?

cesarenaldi (approved):

PR #131, opened 2023-01-31
Restrict FeedItemEventType

paulxuca:
https://avara.height.app/T-5148

### DIFF
--- a/packages/react/src/feed/FeedEventItemType.ts
+++ b/packages/react/src/feed/FeedEventItemType.ts
@@ -1,0 +1,9 @@
+import { FeedEventItemType as LensFeedEventItemType } from '@lens-protocol/api-bindings';
+
+export enum FeedEventItemType {
+  Comment = LensFeedEventItemType.Comment,
+  Post = LensFeedEventItemType.Post,
+  Mirror = LensFeedEventItemType.Mirror,
+  CollectComment = LensFeedEventItemType.CollectComment,
+  CollectPost = LensFeedEventItemType.CollectPost,
+}
--- a/packages/react/src/feed/useFeed.ts
+++ b/packages/react/src/feed/useFeed.ts
@@ -3,12 +3,26 @@ import {
   FeedItemFragment,
   useFeedQuery,
 } from '@lens-protocol/api-bindings';
-import { never } from '@lens-protocol/shared-kernel';
+import { nonNullable } from '@lens-protocol/shared-kernel';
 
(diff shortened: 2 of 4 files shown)


packages/react/src/feed/useFeed.ts
```diff
@@ -1,15 +1,37 @@
...
+import { FeedEventItemType } from './FeedEventItemType';
 
 export type UseFeedArgs = PaginatedArgs<{
   observerId?: string;
   profileId: string;
   restrictEventTypesTo?: FeedEventItemType[];
 }>;
 
-export { FeedEventItemType };
+const mapRestrictEventTypesToLensTypes = (restrictEventTypesTo?: FeedEventItemType[]) => {
```
desfero, line 18: nit: wondering if we could get rid of `ifs` at all
```
const SupportedFeedEvenTypesMap: Record<FeedEventItemType, LensFeedEventItemType> = {
  [FeedEventItemType.Comment]: LensFeedEventItemType.Comment,
  [FeedEventItemType.Post]: LensFeedEventItemType.Post,
  [FeedEventItemType.Mirror]: LensFeedEventItemType.Mirror,
  [FeedEventItemType.CollectComment]: LensFeedEventItemType.CollectComment,
  [FeedEventItemType.CollectPost]: LensFeedEventItemType.CollectPost,
};

const mapRestrictEventTypesToLensTypes = (restrictEventTypesTo?: FeedEventItemType[]) => {
  return restrictEventTypesTo?.map((type) =>
    nonNullable(SupportedFeedEvenTypesMap[type], "Can't map the feed even type"),
  );
};
```

packages/react/src/feed/FeedEventItemType.ts
```diff
@@ -0,0 +1,9 @@
+import { FeedEventItemType as LensFeedEventItemType } from '@lens-protocol/api-bindings';
+
+export enum FeedEventItemType {
```
desfero, line 3: nit: just thinking, maybe it's not the responsibility of `@lens-protocol/react` to protect which `FeedEvenItemType` is supported but rather `@lens-protocol/react` should overwrite the `useFeedHook` to guide what can be used?

desfero (changes requested):
Looks good, just small nitpicks to discuss

packages/react/src/feed/FeedEventItemType.ts
desfero: after thinking a little bit more it seems not worth the effort to add more complexity to the `api-bindings`

krzysu (approved):

desfero (approved):

PR #117, opened 2023-01-25
Add referencePolicy

paulxuca:

PR #132, opened 2023-01-31
feat: no comments on mirrors

cesarenaldi:
Reduces opportunities to comment on mirrors
by making changes that goes into Lens Protocol v2 paradigm
where users cannot comment on mirrors (nor setup reference
module settings on it).

It also fixes TS issue on publication type policy found in https://github.com/lens-protocol/lens-sdk/pull/117 and https://github.com/lens-protocol/lens-sdk/pull/108


packages/api-bindings/src/graphql/ReferencePolicy.ts
```diff
@@ -0,0 +1,26 @@
...
+type UnknownPolicy = {
+  type: ReferencePolicyType.UNKNOWN;
+  contractAddress: string;
+  data: string;
+};
+
+type DegreesOfSeparationPolicy = {
+  type: ReferencePolicyType.DEGREES_OF_SEPARATION;
+  degreesOfSeparation: number;
+};
```
cesarenaldi, line 16: The reason I am omitting:
```ts
    commentsRestricted: boolean;
    mirrorsRestricted: boolean;
```

from the "read path" is because it introduces another way of checking for capability of mirroring and/or commenting.

The canonical way of doing that is to use the `canMirror` and `canComment` edges on `Post` and `Comment` nodes:
image

Also compared to initial approach where the `ReferencePolicy` (used by the `CreatePostRequest` and `CreateCommentRequest` request models, now called `ReferencePolicyConfig` to make a distinction) I am also flattening out the data (no `params` wrapper). The `params` wrapper is reminiscent of reference module implementation details, instead the `ReferencePolicy` is an abstraction that protect the consumer from such details. Take for example `AnyonePolicy` which is used in case of no-reference module being set up at all for a given publication.

packages/api-bindings/src/apollo/TypePolicy.ts
```diff
@@ -15,16 +15,23 @@ type TypedFieldFunctionOptions<TAll> = Overwrite<
...
+  options: FieldFunctionOptions<TAll>,
 ) => TExisting | undefined;
 
+export type FieldPolicy<TExisting, TAll> = Overwrite<
+  UnpatchedFieldPolicy<TExisting>,
+  { read?: FieldReadFunction<TExisting, TAll> }
+>;
+
+type FieldPolicies<T> = Partial<{
+  [key in keyof T]: FieldPolicy<T[key], T> | FieldReadFunction<T[key], T>;
```
cesarenaldi, line 29: Previously if the field policy was implemented as the object:

```ts
{
  read(...) {...},
  merge(...) {...}
}
```
The read function was poorly typed by the original Apollo `FieldReadFunction<any>`. While I was fixing the types here I aligned this other syntax for field policies to take benefit of our patched `FieldReadFunction` implementation.

packages/api-bindings/src/graphql/ReferencePolicy.ts
reecejohnson: Makes sense, good to maintain a single source of truth as to where to check if a publication can be mirrored / commented on 👍

reecejohnson (approved):

PR #133, opened 2023-01-31
Add ability to add downvotes (Missing reaction type)

paulxuca:
https://avara.height.app/T-4999

### DIFF
--- a/packages/react/src/publication/adapters/ReactionPresenter.ts
+++ b/packages/react/src/publication/adapters/ReactionPresenter.ts
@@ -1,53 +1,55 @@
-import { ReactionTypes } from '@lens-protocol/api-bindings';
+import { getApiReactionType, getDomainReactionType } from '@lens-protocol/api-bindings';
 import { ReactionType } from '@lens-protocol/domain/entities';
-import { IReactionPresenter, ReactionRequest } from '@lens-protocol/domain/use-cases/publications';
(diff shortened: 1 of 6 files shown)


packages/react/src/publication/adapters/ReactionPresenter.ts
```diff
@@ -39,46 +50,37 @@ export class ReactionPresenter
...
-        }
-        break;
-
-      case ReactionType.DOWNVOTE:
-        throw new Error('Downvotes support not implemented');
-      default:
-        throw new Error('Unknown reaction type');
-    }
+    this.publicationCacheManager.update(request.publicationId, (current) => {
+      const removedStatKey = current.reaction
```
krzysu, line 54: what is `removedStatKey` for?

krzysu:
could you also add tests for downvote case and for both upvote and downvote, I remember when I was testing them both manually with API the results were confusing, sth like downvote was removing upvote, but I don't remember the details now.

packages/react/src/publication/adapters/ReactionPresenter.ts
paulxuca: If we add a downvote to a publication that already has an upvote, we have to remove 1 from the stat value that the active profile previously had -- `removedStatKey` is that key in `PublicationStats` (`totalUpvotes` | `totalDownvotes`)

krzysu (approved):
👍

PR #137, opened 2023-02-02
Add a way to log into a specific profile by id

paulxuca:


packages/domain/package.json
```diff
@@ -46,6 +46,7 @@
     "prettier:fix": "prettier --write .",
     "prettier": "prettier --check .",
     "test": "jest",
+    "test:watch": "jest --watch",
```
krzysu, line 49: if we agree to add that, then I'd add it to all packages so that we have the same scripts available everywhere

packages/domain/src/use-cases/profile/ActiveProfileLoader.ts
```diff
@@ -12,14 +12,26 @@ export class ActiveProfileLoader {
...
-  async loadActiveProfileByOwnerAddress(walletAddress: string) {
+  async loadActiveProfileByOwnerAddress(walletAddress: string, profileId?: string) {
     const activeProfile = await this.activeProfileGateway.getActiveProfile();
 
     if (activeProfile) {
       await this.activeProfilePresenter.presentActiveProfile(activeProfile);
       return;
     }
-    const [firstProfile] = await this.profileGateway.getAllProfilesByOwnerAddress(walletAddress);
+    const profiles = await this.profileGateway.getAllProfilesByOwnerAddress(walletAddress);
```
desfero, line 22: wondering if we can use `handle` over `profileId` which seems to be closer to how protocol works.

We already have a method `loadActiveProfileByHandle` that we can call in that case and keep this one to load first active profile by the wallet address. 

nit: maybe it's also worth checking in `loadActiveProfileByHandle` if the `handle` belongs to the active wallet and throw some invariant error.

desfero: We could go also one step further and remove `loadActiveProfileByOwnerAddress` and lift the logic to select the first profile handle to the caller side and call `loadActiveProfileByHandle` directly.

packages/react/src/wallet/useWalletLogin.ts
```diff
@@ -16,7 +16,7 @@ export function useWalletLogin() {
   const loginWallet = useWalletLoginController();
 
   return {
-    login: async (signer: Signer) => {
+    login: async (signer: Signer, profileId?: string) => {
```
desfero, line 19: Missing the example in the `web-wagmi`. 
It seems there is also an important bug as after refreshing you still log in as a first profile and it should be the same profile you choose in the login flow (see `startWithCredentials` from `Bootstrap`)

desfero (changes requested):

packages/react/src/wallet/useWalletLogin.ts
cesarenaldi, line 19: Wondering... in general knowing your own profile Id is very unlikely. Should we use the profile handle instead?

paulxuca: Perhaps but the usecase I was thinking for this functionality would be querying the api and displaying a list of profiles to log into. Should we support handle in a separate task?

PR #139, opened 2023-02-02
feat: auto bundle package in dev and auto generate needed export config

reecejohnson:
I have setup the `preconstruct dev` command to be ran on `postinstall`, this will link the packages together for development so that there is no need to continuously rebuild when working between packages.

Another cool thing here is that all of the manual work around creating the multiple entry points is now automated by Preconstruct. This is powered by defining the `entrypoints` in the specific package's `package.json` and then running `pnpm preconstruct fix`.

Try this out:
- go to the `domain` package
- run `rm -rf entities mocks use-cases` to delete the three directories containing the extra package.json required for the multiple exports 
- then run `pnpm preconstruct fix` and the directories and files are generated, linking to the correct files in `dist`


packages/api-bindings/package.json
```diff
@@ -69,11 +69,18 @@
...
     "rimraf": "^3.0.2",
     "ts-jest": "^29.0.3",
-    "typescript": "^4.9.3",
-    "tsup": "^6.5.0"
+    "typescript": "^4.9.3"
   },
   "peerDependencies": {
-    "react": "^18.2.0"
+    "react": "^18.2.0",
+    "@faker-js/faker": "^7.6.0"
```
reecejohnson, line 76: Not 100% sure if we need this here, potentially a workaround but `preconstruct` was upset with us using `mocks` as an entry point and exporting things that may use `@faker-js/faker`

packages/blockchain-bindings/package.json
```diff
@@ -8,21 +8,21 @@
...
   "exports": {
     ".": {
-      "types": "./dist/index.d.ts",
-      "default": "./dist/index.js"
+      "module": "./dist/lens-protocol-blockchain-bindings.esm.js",
+      "default": "./dist/lens-protocol-blockchain-bindings.cjs.js"
     },
     "./package.json": "./package.json"
   },
-  "types": "./dist/index.d.ts",
```
reecejohnson, line 18: no need for `types` property at all

packages/api-bindings/mocks/package.json
```diff
@@ -1,4 +1,5 @@
 {
```
reecejohnson, line 1: these extra  dirs and `package.json` are autogenerated, we could consider adding to .gitignore?

packages/api-bindings/package.json
```diff
@@ -10,24 +10,24 @@
   "type": "module",
   "exports": {
     ".": {
-      "types": "./dist/index.d.ts",
```
cesarenaldi, line 13: how types are exported now?

packages/api-bindings/package.json
cesarenaldi: Can you please clarify "upset"? In what sense and why adding the fakerjs as peer dep solves it?

packages/domain/entities/package.json
```diff
@@ -1,4 +1,4 @@
 {
-  "type": "module",
-  "main": "../dist/entities/index.js"
```
cesarenaldi, line 3: Maybe I misunderstood... but does `preconstruct` creates these fallback directories?

packages/react/src/LensProvider.tsx
```diff
@@ -1,5 +1,5 @@
 import { invariant } from '@lens-protocol/shared-kernel';
-import { ReactNode, useEffect, useRef, useState } from 'react';
+import React, { useEffect, useRef, useState } from 'react';
```
cesarenaldi, line 2: What complained about this?

packages/react/tsconfig.json
```diff
@@ -3,5 +3,6 @@
   "compilerOptions": {
     "module": "ESNext",
     "outDir": "dist"
-  }
+  },
+  "include": ["./src"]
 }
```
cesarenaldi, line 8: Do we still need the ditinction between `tsconfig.json` and `tsconfig.base.json`?

packages/domain/entities/package.json
reecejohnson: It does, yes

packages/api-bindings/package.json
reecejohnson: inside each `dist` dir, a `declarations` dir is created with `.d.ts` files for all of the types

there is no need to reference this from the `package.json`

packages/api-bindings/package.json
reecejohnson: Preconstruct requires that all packages that are imported by a package be specified in dependencies or peerDependencies. This is to avoid the mistake of publishing a package that imports a package that isn't specified in its dependencies causing a consumer to be unable to use the package.

packages/react/src/LensProvider.tsx
reecejohnson: The nextjs app build... was having a few issues with it. This is a quick workaround.

In Next.js `React` is a global and is not needed to be imported, I believe that would be the cause of this issue.

This PR is a working POC, if we decide to go ahead with this approach, I will come back and iron out some of these quirks.

packages/react/tsconfig.json
reecejohnson: Another strange quirk I deferred to after we agree on a path for this - seems the base TS config `includes` is being picked up by preconstruct

cesarenaldi:
@reecejohnson when we moved to `tsup` we lost ability to have source maps for all our packages. As result debugging from the browser is annoying. Is this solution including support for that or can it be added?

packages/domain/entities/package.json
cesarenaldi: so we can delete them? Sorry I am confused... I would assume this file would be delete then or am I missing something?

packages/api-bindings/package.json
cesarenaldi: how do packages that depends on these find them? Is `declarations` dir a standard de-factor in this space?

packages/api-bindings/mocks/package.json
desfero: nope, can't ignore them as they should be pushed to the `npm` registry so bundlers can find properly imports
Like technically you can ignore and than make sure you still push them to the registry but it's probably not worth to add one layer of possible mistakes

packages/api-bindings/package.json
desfero: nope, works a little bit differently.

Typescript when looking for types is checking the `types` in `package.json` if that's not avaiable he tries to find the types by looking into where `main` points (in this case to `dist/lens-protocol-api-bindings.cjs.js`). He then goes to that folder and tries to see if there is a file called `dist/lens-protocol-api-bindings.cjs.d.ts`. 

Preconstruct just assumes the second behavior and relies on typescript to find the type definition by following where `main` points.

The generated `dist/lens-protocol-api-bindings.cjs.d.ts` points under the hood to `delcarations` folder 
```
exort * from "./declarations/src/index";

```
That's it, hope it clarifies your question @cesarenaldi.

packages/api-bindings/package.json
desfero: that is really no go IMO as would throw the 
```
`@faker-js/faker` is missing but required by `@lens-protocol/api-bindings` 
```
on the consumer side when installing `@lens-protocol/react`.

Will have a look on monday at how we can solve that in a way that would not affect consumers.

desfero: Same affects all packages that have modified `peerDependencies`

desfero (changes requested):
looks nice and clean really, DX is fixed

packages/api-bindings/package.json
reecejohnson: Agreed, that would be less than ideal.

Will explore our options around this

packages/domain/entities/package.json
reecejohnson: we can add them to gitignore but every time we run a build they are generated

examples/web-wagmi/src/components/header/Header.tsx
```diff
@@ -41,7 +41,7 @@ export function Header() {
           <NavLink
             key={path}
             to={path}
-            className={({ isActive }) => (isActive ? 'current' : undefined)}
+            className={({ isActive }: { isActive: boolean }) => (isActive ? 'current' : undefined)}
```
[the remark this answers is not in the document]
reecejohnson: no need for this, removed

package.json
```diff
@@ -43,5 +48,20 @@
...
+    "packages": [
+      "packages/domain",
+      "packages/react",
+      "packages/shared-kernel",
+      "packages/api-bindings",
+      "packages/wagmi",
+      "packages/blockchain-bindings",
+      "packages/storage"
+    ],
+    "exports": true,
```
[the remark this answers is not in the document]
reecejohnson: ah nice spot 👍

packages/api-bindings/package.json
krzysu: I guess moving `faker` to dependencies (not devDeps) would also solve it, or we don't export mocks at all.
`preconstruct ` is correct that if we want to export `mocks` and they use `faker`, then `faker` has to be part of the bundle or a peer dep, no?

krzysu:
I noticed that while running web-wagmi example app and doing changes in packages, the app is exploding with `Maximum call stack size exceeded` or breaks with `Error: InvariantError: LensProvider: config cannot be changed`
So I have to reload (or sometimes restart) the example app after every change.

reecejohnson:
> @reecejohnson when we moved to `tsup` we lost ability to have source maps for all our packages. As result debugging from the browser is annoying. Is this solution including support for that or can it be added?

Yes this works pretty well @cesarenaldi 

https://user-images.githubusercontent.com/34844044/217234841-9fa3d2b0-9f2a-49bc-a4cf-358bc51c0819.mov

reecejohnson:
> I noticed that while running web-wagmi example app and doing changes in packages, the app is exploding with `Maximum call stack size exceeded` or breaks with `Error: InvariantError: LensProvider: config cannot be changed` So I have to reload (or sometimes restart) the example app after every change.

I do see the config cannot be changed error, that's interesting.

But I'm not experiencing the call stack exceeded. What package did you update?

babel.config.cjs
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+  presets: ["@babel/preset-typescript", "@babel/preset-react"],
```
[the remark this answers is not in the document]
reecejohnson: yes 😅

krzysu:
> > I noticed that while running web-wagmi example app and doing changes in packages, the app is exploding with `Maximum call stack size exceeded` or breaks with `Error: InvariantError: LensProvider: config cannot be changed` So I have to reload (or sometimes restart) the example app after every change.
> 
> I do see the config cannot be changed error, that's interesting.
> 
> But I'm not experiencing the call stack exceeded. What package did you update? @krzysu

I think it was api-bindings or domain, sth down there in the stack.

packages/domain/package.json
```diff
@@ -68,8 +45,31 @@
     "rimraf": "^3.0.2",
     "ts-jest": "^29.0.3",
     "typescript": "^4.9.3",
-    "tsup": "^6.5.0",
     "wait-for-expect": "^3.0.2"
   },
-  "prettier": "@lens-protocol/prettier-config"
+  "peerDependencies": {
+    "@faker-js/faker": "^7.6.0",
```
desfero, line 51: should all of these be marked as `optional` dependency? 
Otherwise we would get an error on the consumer side when installing the `@lens-protocol/react` that some peers for `domain` package are missing.

Still i think we need to find a better way for mocks that would not force us having `optional` test dependencies in peers as that's confusing from the consumer point of view.

desfero: Could the middle ground approach for now would be to remove `mocks` from `preconstruct` `entrypoints` and point to the unprocessed file?

mocks/package.json
```
{
  "main": "../src/mocks.ts"
}
```

It should still work properly if we transpile the `@lens-protocol` packages (and we do if you have a look at `transformIgnorePatterns`).

In that case we never publish mocks to the registry and we don't need to have any `peerDependency` to our internal `devDependencies`?
What do you think?

turbo.json
```diff
@@ -23,6 +28,7 @@
       "inputs": ["src/**/*.tsx", "src/**/*.ts"]
     },
     "dev": {
+      "dependsOn": ["//#dev:link"],
```
desfero, line 31: what does that mean? Can you please point me to some documentation to understand that part a little bit more?

desfero (changes requested):

turbo.json
reecejohnson: this means that the `dev` task, which runs the example apps in development mode, is dependant upon the `dev:link` task being run.

docs are here: https://turbo.build/repo/docs/reference/configuration#dependson

package.json
```diff
@@ -6,21 +6,30 @@
     "packages/*",
     "examples/*"
   ],
+  "type": "module",
   "scripts": {
-    "build": "turbo run build",
+    "build": "pnpm clean && turbo build",
+    "postinstall": "preconstruct dev",
+    "clean": "rimraf packages/*/*/dist packages/*/dist",
```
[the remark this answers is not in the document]
reecejohnson: can confirm it is needed, without this we see some strange behaviour of files rewritten due to the symlink

package.json
```diff
@@ -6,21 +6,30 @@
...
+    "dev": "turbo dev",
+    "dev:packages": "pnpm preconstruct dev",
     "lint": "turbo run lint --parallel",
     "lint:fix": "turbo run lint:fix --parallel",
     "lint:examples": "sh ./scripts/lint-examples.sh",
     "deduplicate:check": "pnpm-deduplicate --list",
     "deduplicate:fix": "pnpm-deduplicate"
   },
   "devDependencies": {
+    "@babel/preset-env": "^7.20.2",
```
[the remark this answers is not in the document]
reecejohnson: Agreed, sounds good - will tackle outside of this PR as it's been hanging around for a while now

packages/api-bindings/jest.config.js
```diff
@@ -1,4 +1,4 @@
-export default {
+module.exports = {
```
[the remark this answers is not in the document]
reecejohnson: migrated all files to `.ts` now with `export default`

desfero (approved):
LGTM, let's push this forward and make our lives happier 🥇

PR #140, opened 2023-02-02
Migrate lens protocol sdk gated to monorepo (T-3748)

desfero:
This pr moves the `@lens-protocol/sdk-gated` to the monorepo.
Code was moved as it is with some fixes to the unit tests to make sure the all pass during our CI pipeline.
Refer to the commits to see all changes done to the migrated files.


.github/workflows/ci.yaml
```diff
@@ -12,16 +12,28 @@ jobs:
...
+
       - run: pnpm build
+
       - run: pnpm lint
-      - run: pnpm test
+
+      - name: "Run tests"
+        run: pnpm test
+        env:
+          GATED_SDK_TEST_PK:  ${{ secrets.GATED_SDK_TEST_PK }}
```
desfero, line 35: maybe anyone know some better way to provide env to just `sdk-gated` as this one is polluting our global github actions config

krzysu: you would need to run a separate task only to run tests of sdk-gated and then define the env var there, but I think there is nothing wrong with the way it is now. we just need some env vars naming convention. shall they be prefixed with subpackage name?

packages/sdk-gated/package.json
```diff
@@ -0,0 +1,85 @@
...
+  "browser": "./dist/browser.js",
+  "types": "./dist/index.d.ts",
+  "license": "MIT",
+  "files": [
+    "dist",
+    "package.json",
+    "CHANGELOG.md"
+  ],
+  "scripts": {
+    "build": "pnpm clean && pnpm rollup:prod",
```
krzysu, line 18: `rollup:prod` script is running `clean` already so here it is run twice

packages/sdk-gated/package.json
```diff
@@ -0,0 +1,85 @@
+{
+  "name": "@lens-protocol/sdk-gated",
+  "version": "1.2.0",
```
krzysu, line 3: what is the strategy for future releases? will sdk-gated be integrated into react-sdk and client-sdk and not available as a separate package anymore?

krzysu (approved):

packages/sdk-gated/package.json
desfero: Good catch, removed

packages/sdk-gated/package.json
desfero: that is still to be explored. 

Right now there is still a plan to somehow provide low-level gated SDK in case someone talks with the API directly (and at the moment 100% folks use API directly).

We need to see how the `@lens-protocol/client` gonna be adopted when released and make the proper decision. 

But no matter how awesome would be the new client sdk it would take some time to have good adoption across ecosystem.

TL;DR; we still want to release gated sdk as a separate beast.

.github/workflows/ci.yaml
desfero: good point, fixed.

PR #145, opened 2023-02-06
Add a way to log into a specific profile by handle

paulxuca:
@desfero Saw your comments on the other PR and wanted to get some feedback before fixing tests, adding examples and cleaning up. Let me know what you think!

### DIFF
--- a/examples/web-wagmi/src/misc/LoginSpecificProfile.tsx
+++ b/examples/web-wagmi/src/misc/LoginSpecificProfile.tsx
@@ -1,0 +1,116 @@
+import {
+  ProfileOwnedByMeFragment,
+  useProfilesOwnedBy,
+  useWalletLogin,
+  useWalletLogout,
+  WalletData,
+} from '@lens-protocol/react';
+import { useState } from 'react';
+import { useAccount, useConnect, useDisconnect } from 'wagmi';
+import { InjectedConnector } from 'wagmi/connectors/injected';
+
+import { LoginButton } from '../components/auth/LoginButton';
+import { WhenLoggedInWithProfile, WhenLoggedOut } from '../components/auth/auth';
+
+function UnauthenticatedContent() {
+  const [handle, setHandle] = useState('');
+
+  return (
+    <>
+      <h2>Login to a Specific Profile</h2>
+
+      <p>Handle</p>
+
+      <input
+        placeholder="@me"
+        onChange={(e) => {
+          setHandle(e.target.value);
+        }}
+        value={handle}
+        style={{ display: 'block' }}
+      />
+
+      <LoginButton handle={handle} />
+    </>
+  );
+}
+
+function AuthenticatedContent({
+  wallet,
+  profile,
+}: {
+  wallet: WalletData;
+  profile: ProfileOwnedByMeFragment;
+}) {
+  const { data: profiles } = useProfilesOwnedBy({ address: wallet.address });
+  const [handle, setHandle] = useState<string>();
+
+  const { execute: login } = useWalletLogin();
+  const { execute: logout } = useWalletLogout();
+
+  const { isConnected } = useAccount();
+  const { disconnectAsync } = useDisconnect();
+
+  const { connectAsync } = useConnect({
+    connector: new InjectedConnector(),
+  });
+
+  const onLoginClick = async () => {
+    await logout();
+
+    if (isConnected) {
+      await disconnectAsync();
+    }
+
+    const { connector } = await connectAsync();
+
+    if (connector instanceof InjectedConnector) {
+      const signer = await connector.getSigner();
+      await login(signer, handle);
+    }
+  };
+
+  return (
+    <>
+      <h2>Login to a Specific Profile</h2>
+
+      <p>
+        You are already logged into a profile; Pick a different one below to log into a different
+        one after logging out.
+      </p>
+
+      <select
+        onChange={(e) => {
+          setHandle(e.target.value);
+        }}
+        value={handle || profile.handle}
+      >
+        {profiles?.map((p) => {
+          return (
+            <option key={p.handle} value={p.handle}>
+              {p.handle}
+            </option>
+          );
+        })}
+      </select>
+
+      <button onClick={onLoginClick}>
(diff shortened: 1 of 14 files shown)


packages/domain/src/use-cases/lifecycle/Bootstrap.ts
```diff
@@ -70,7 +71,9 @@ export class Bootstrap<T extends TransactionRequestModel> {
   private async startWithCredentials(wallet: Wallet) {
     this.activeWalletPresenter.presentActiveWallet(wallet);
 
-    await this.activeProfileLoader.loadActiveProfileByOwnerAddress(wallet.address);
+    const profiles = await this.profileGateway.getAllProfilesByOwnerAddress(wallet.address);
+
+    await this.activeProfileLoader.loadActiveProfileByHandle(profiles[0].handle);
```
cesarenaldi, line 76: `ActiveProfileLoader` is a convenience object aimed at encapsulating some application specific domain logic around active profile management. This also helped isolate dependencies that are needed for this purpose (gateway and presenter) from the interactors that uses this helper object.

In this new configuration the `Bootstrap` interactor has a new reference to the `IProfileGateway`, retains a reference to the `ActiveProfileLoader` and uses the first to instruct the second on what profile to make as active. The second then retrieve the profile again from the `IProfileGateway`.

I would consider to put this responsibility back into the `ActiveProfileLoader` so that the data flow inefficiency disappaer.

But before doing that I would also observe that using the `profiles[0]` at bootstrap time is the cause of the bug @desfero mentioned here: https://github.com/lens-protocol/lens-sdk/pull/137#discussion_r1097123961

Let's fix that and optimise this once we have the full picture of the final data flow.

cesarenaldi (changes requested):
I think we need to address this: https://github.com/lens-protocol/lens-sdk/pull/145/files#r1102735185

packages/domain/src/use-cases/lifecycle/Bootstrap.ts
paulxuca: Got it, makes sense and will fix!

examples/web-wagmi/src/misc/LoginSpecificProfile.tsx
```diff
@@ -0,0 +1,31 @@
...
+  const [handle, setHandle] = useState('');
+  const activeProfile = useActiveProfile();
+
+  return (
+    <>
+      {activeProfile.data && <Navigate to="/" />}
+
+      <h2>Login to a Specific Profile</h2>
+
+      <p>Handle</p>
```
krzysu, line 17: I guess better to use `useProfilesOwnedBy` hook here and display a list of owned profiles instead of just an input.

examples/web-wagmi/src/misc/LoginSpecificProfile.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+import { LoginButton } from '../components/auth/LoginButton';
+
+export function LoginSpecificProfile() {
+  const [handle, setHandle] = useState('');
+  const activeProfile = useActiveProfile();
+
+  return (
+    <>
+      {activeProfile.data && <Navigate to="/" />}
```
krzysu, line 13: instead of redirecting away from this page, better to display an infobox with sth like "you are already logged in as @handle, if you want to change your active profile select from the list below"

examples/web-wagmi/src/misc/MiscPage.tsx
```diff
@@ -1,39 +1,53 @@
...
+      description: `View all the modules supported by the Lens Protocol`,
+      path: '/misc/useEnabledModules',
+    },
+    {
+      label: 'useRecentTransactions',
+      description: `Example of listing + clearing recent completed transactions.`,
+      path: '/misc/useRecentTransactions',
+    },
+
+    ...(activeWallet.data === null
```
krzysu, line 40: if you allow visiting this page as a logged-in user then you don't need this condition here. it is the example app so hiding features from users don't make much sense. they should be able to explore as much as possible.

examples/web-wagmi/src/misc/LoginSpecificProfile.tsx
paulxuca: Is there a a way to get the wallet address without triggering the sign-in flow? (Not sure, generally asking for a way to do so)

packages/domain/src/use-cases/profile/__tests__/ActiveProfileLoader.spec.ts
```diff
@@ -57,7 +57,7 @@ describe(`Given the ${ActiveProfileLoader.name} interactor`, () => {
       expect(activeProfilePresenter.presentActiveProfile).toHaveBeenCalledWith(profile);
     });
 
-    it('should bail out if not active profile is present and the wallet does not own any profile', async () => {
+    it('should throw if not active profile is present and the wallet does not own any profile', async () => {
```
cesarenaldi, line 60: I think we need to account for login with wallet that do not have an profile (cause they cannot claim one at first or because they have to go through onboarding once open profile minting will be enable).

So throwing is probably not necessary.

packages/domain/src/use-cases/profile/ActiveProfileLoader.ts
```diff
@@ -12,18 +12,19 @@ export class ActiveProfileLoader {
...
+  async loadActiveProfileByOwnerAddress(walletAddress: string, handle?: string) {
     const activeProfile = await this.activeProfileGateway.getActiveProfile();
 
     if (activeProfile) {
       await this.activeProfilePresenter.presentActiveProfile(activeProfile);
       return;
     }
-    const [firstProfile] = await this.profileGateway.getAllProfilesByOwnerAddress(walletAddress);
+    const profiles = await this.profileGateway.getAllProfilesByOwnerAddress(walletAddress);
+    const profile = handle ? profiles.find((p) => p.handle === handle) : profiles[0];
```
cesarenaldi, line 23: LGTM, I think it addresses initial @desfero comment.

cesarenaldi (approved):

PR #146, opened 2023-02-07
Allow any signer (Remove assertRequiredSigner)

paulxuca:


packages/react/src/wallet/infrastructure/__tests__/SignerFactory.spec.ts
```diff
@@ -126,5 +126,23 @@ describe(`Given an instance of the ${SignerFactory.name}`, () => {
...
     });
+
+    describe('with a wallet', () => {
+      it('should retrieve the signer', async () => {
+        // This is the private key of the `@jsisthebest.test` profile
+        // It's a public private key so anyone can modify the profile
+        // For your own convenience change to the private key of a new wallet
+        const testWalletPrivateKey =
+          '6c434da5e5c0e3a8e0db5cf835d23e04c7592037854f0700c26836be7581c68c';
+
```
krzysu, line 137: we decided to not have any private keys openly in the codebase, for tests in other packages we use .env vars and secrets for github actions, see https://github.com/lens-protocol/lens-sdk/settings/secrets/actions and CLIENT_TEST_WALLET_PRIVATE_KEY or SDK_GATED_TEST_PK in the main branch

so I suggest you add a new env var to react package, call it REACT_TEST_WALLET_PRIVATE_KEY and use it in the test instead

packages/react/src/wallet/infrastructure/SignerFactory.ts
```diff
@@ -81,7 +82,9 @@ export class SignerFactory implements ISignerFactory {
 
   private async addChain(signer: RequiredSigner, chainConfig: AddEthereumChainParameter) {
     try {
-      await signer.provider.send('wallet_addEthereumChain', [chainConfig]);
+      if (signer.provider && signer.provider instanceof JsonRpcProvider) {
+        await signer.provider?.send('wallet_addEthereumChain', [chainConfig]);
```
desfero, line 86: seems we don't need an elvis operator over here.

packages/react/src/wallet/infrastructure/__tests__/SignerFactory.spec.ts
desfero: I think it's okay to have the private key here until it's a unit test that doesn't reach `api/smart-contracts`.

If it becomes more like the integration tests that talk with something external we need to hide the private key to avoid someone external changing the state of a given address (for .e.g by burning profile) and those affecting our CI pipeline.

packages/react/src/wallet/infrastructure/SignerFactory.ts
```diff
@@ -92,7 +95,11 @@ export class SignerFactory implements ISignerFactory {
     chainConfig: AddEthereumChainParameter,
   ): PromiseResult<void, WalletConnectionError> {
     try {
-      await signer.provider.send('wallet_switchEthereumChain', [{ chainId: chainConfig.chainId }]);
+      if (signer.provider && signer.provider instanceof JsonRpcProvider) {
```
desfero, line 98: this now modified the logic that relies on the `INCORRECT_CHAIN` error to be thrown.

Previously when switching chain failed we returned the `WalletConnectError` as a failure.
Now when there is no way to switch the network we just move forward making the logic incorrect as we can't move forward connected to the wrong network cause it will cause broadcasting to fail later down the line (chain id is a part of signed message).

So in case we can't try to `switchChain` we should throw the error right away

desfero (changes requested):
Also would be good if we can remove `assertRequiredSigner` entirely (remember to add changeset so folks know it was removed)

packages/react/src/wallet/infrastructure/SignerFactory.ts
```diff
@@ -49,6 +49,12 @@ export class SignerFactory implements ISignerFactory {
     }
 
     if (chainType) {
+      try {
+        signer._checkProvider('switchChain');
```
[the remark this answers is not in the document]
cesarenaldi: AFAIK the most effective way is to call a method verify the error.code among `ethers.js` error codes: https://github.com/ethers-io/ethers.js/blob/main/src.ts/utils/errors.ts#L121

We do something similar here (for a different error): https://github.com/lens-protocol/lens-sdk/blob/main/packages/react/src/wallet/adapters/ConcreteWallet.ts#L145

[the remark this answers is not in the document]
cesarenaldi: I don't know about using `_checkProvider`, but this is a proper internal as far as we know. Compared to `Signer. _signTypedData` which is fully documented and part of public interface from v6.

packages/react/src/utils.ts
```diff
@@ -1,16 +1,16 @@
...
+import { Signer } from 'ethers';
 
 import { RequiredSigner } from './wallet/adapters/ConcreteWallet';
 
 export function isRequiredSigner(signer: Signer): signer is RequiredSigner {
-  return '_signTypedData' in signer && signer.provider instanceof providers.JsonRpcProvider;
+  return '_signTypedData' in signer;
 }
 
 export function assertRequiredSigner(signer: Signer): asserts signer is RequiredSigner {
```
desfero, line 10: Seems the `assertRequiredSigner` can be removed from the codebase (and from relevant docs)?

packages/react/src/wallet/infrastructure/SignerFactory.ts
```diff
@@ -37,6 +39,16 @@ export class SignerFactory implements ISignerFactory {
     }
 
     if (chainType) {
+      try {
+        await signer.getChainId();
```
desfero, line 43: when it's possible for that part of code to fail? 
Wondering if we can simplify the code to not call the `signer.getChainId` twice and to not return `WalletConnectionError(WalletConnectionErrorReason.INCORRECT_CHAIN)` twice (here and inside `switchChain`).

desfero (changes requested):
Some comments that we need to clarify before moving forward

packages/react/src/wallet/infrastructure/SignerFactory.ts
paulxuca: ah yep you're right that would be simpler

desfero (changes requested):
Could you also push a changeset?
E.g. "Removed the need for `assertRequiredSigner` when getting signer"
as the code changes here are touching consumer-facing API

desfero (approved):

PR #147, opened 2023-02-07
`useProfilesOwnedBy` should be updated as new profile is created

paulxuca:
Docs: https://dash.readme.com/project/lensprotocol/v1.1/docs/use-profiles-owned-by-me

### DIFF
--- a/examples/web-wagmi/src/profiles/UseCreateProfile.tsx
+++ b/examples/web-wagmi/src/profiles/UseCreateProfile.tsx
@@ -4,11 +4,9 @@ import { never } from '../utils';
 import { ProfileCard } from './components/ProfileCard';
 
 function OwnedProfiles({ activeWalletAddress }: { activeWalletAddress: string }) {
-  const { data, loading } = useProfilesOwnedBy({
+  const { data } = useProfilesOwnedBy({
     address: activeWalletAddress,
   });
-
-  console.log(loading);
 
   return (
     <div>
--- a/examples/web-wagmi/src/profiles/UseCreateProfile.tsx
+++ b/examples/web-wagmi/src/profiles/UseCreateProfile.tsx
@@ -1,39 +1,34 @@
-import { useCreateProfile, useProfile } from '@lens-protocol/react';
-import { useState } from 'react';
+import { useActiveWallet, useCreateProfile, useProfilesOwnedBy } from '@lens-protocol/react';
 
-import { ErrorMessage } from '../components/error/ErrorMessage';
 import { never } from '../utils';
 import { ProfileCard } from './components/ProfileCard';
 
-function ShowProfile({ handle }: { handle: string }) {
-  const { data: profile, error, loading } = useProfile({ handle: `${handle}.test` });
+function OwnedProfiles({ activeWalletAddress }: { activeWalletAddress: string }) {
+  const { data } = useProfilesOwnedBy({
+    address: activeWalletAddress,
+  });
 
-  if (loading) return null;
-
-  if (error) return <ErrorMessage error={error} />;
-
-  return <ProfileCard profile={profile} />;
+  return (
+    <div>
+      {data?.map((profile) => (
+        <ProfileCard key={profile.id} profile={profile} />
+      ))}
+    </div>
+  );
 }
 
 export function UseCreateProfile() {
-  const [newProfileHandle, setNewProfileHandle] = useState<string | null>(null);
-
   const { execute: create, error, isPending } = useCreateProfile();
 
   const onSubmit = async (event: React.FormEvent<HTMLFormElement>) => {
-    setNewProfileHandle(null);
-
     event.preventDefault();
 
     const formData = new FormData(event.currentTarget);
     const handle = (formData.get('handle') as string) ?? never();
-
-    const result = await create(handle);
-
-    if (result.isSuccess()) {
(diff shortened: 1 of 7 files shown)


packages/react/src/transactions/adapters/responders/CreateProfileResponder.ts
```diff
@@ -0,0 +1,16 @@
...
+  GetAllProfilesByOwnerAddressDocument,
+  LensApolloClient,
+} from '@lens-protocol/api-bindings';
+import { CreateProfileRequest } from '@lens-protocol/domain/dist/use-cases/profile';
+import { ITransactionResponder } from '@lens-protocol/domain/use-cases/transactions';
+
+export class CreateProfileResponder implements ITransactionResponder<CreateProfileRequest> {
+  constructor(private readonly client: LensApolloClient) {}
+
+  async commit() {
```
paulxuca, line 11: I couldn't figure out how to check if an existing `GetAllProfilesByOwnerAddress` query was pending with the a partial match on the variables matching `address: activeWallet.address` so for now just invalidating all of the above queries.

cesarenaldi: Ah, this is an interesting approach. Does it work fine?

examples/web-wagmi/src/profiles/UseCreateProfile.tsx
```diff
@@ -15,7 +22,24 @@ function ShowProfile({ handle }: { handle: string }) {
   return <ProfileCard profile={profile} />;
 }
 
+function OwnedProfiles({ activeProfileAddress }: { activeProfileAddress: string }) {
```
desfero, line 25: it's more like an `activeWalletAddress`

examples/web-wagmi/src/profiles/UseCreateProfile.tsx
```diff
@@ -63,6 +89,10 @@ export function UseCreateProfile() {
         {error && <p>{error.message}</p>}
       </form>
 
+      {showExistingProfiles && activeProfile.data && (
```
desfero, line 92: i think there is no harm in showing that always on the UI, can be good to showcase to our users that we properly update `useProfilesOwnedBy` and it can be relied on when building onboarding UI's

Just let's add some header: `All profiles owned by the wallet` so it's clear what the list is about.

What do you think?

packages/react/src/transactions/adapters/responders/CreateProfileResponder.ts
cesarenaldi: If it were to be a more generic query (e.g. `GetProfile`) I would have some reservation as it might trigger way more queries that it needs to. But here we have a very specific query, so there is a good change that there is at most on active query watching it.

I think we need to be careful of not spreading indiscriminately this approach in other areas as it might cause unnecessary coupling and unwanted network requests and re-rendering, but if it works here seems an elegant solution.

I would test only one thing. Given the `useProfilesOwnedBy` is paginated (infinite scroll pagination implies that the cache contains all pages content up to current). What happen if let's say we are at page 2 when we refresh the query?

You can reduce the page size with `limit` and create as many profile as you want via `useCreateProfile` (see example app UI).

desfero: interesting, can you double-check if the `refetchQueries` doesn't trigger the `loading` flag to switch back to `true`?

So the code like below doesn't jump back to loading when you trigger profile creation.
```
const { loading, items } = useProfilesOwnedBy();
const { create } = useCreateProfile();

if (loading) {
  return "loading"
}

if (items.length === 0) {
  return // create profile UI
}

const profile = items[0];

return ... // profiel ui

```

paulxuca: @cesarenaldi It looks it does not properly include the newly created profile in the paginated result if the query is currently on the 2nd page for example. I think I'll need to rethink this approach

@desfero I checked and it does not trigger the `loading` flag to be switched back to true when it gets refetched

packages/react/src/transactions/adapters/responders/CreateProfileResponder.ts
```diff
@@ -0,0 +1,36 @@
...
+  ITransactionResponder,
+} from '@lens-protocol/domain/use-cases/transactions';
+
+export const createdProfilesVar = makeVar<ProfileFragment[]>([]);
+
+export class CreateProfileResponder implements ITransactionResponder<CreateProfileRequest> {
+  constructor(private readonly client: LensApolloClient) {}
+
+  async commit({ request }: BroadcastedTransactionData<CreateProfileRequest>) {
+    const res = await this.client.query<SearchProfilesQuery, SearchProfilesQueryVariables>({
```
desfero, line 21: interesting, was there any downside of an old method just out of curiosity?

paulxuca: Are you referring to old implementation of `commit`? the issue there was that if the user was reading page 2 of `useProfilesOwnedBy` hook then it would not include the newly created profile in that results list.

desfero: got it, thank you

desfero (approved):

PR #148, opened 2023-02-07
feat: client sdk authentication token storage and management

krzysu:
- Authentication class is managing Credentials and CredentialsStorage, it also returns an authentication header for the API request


packages/client/src/authentication/adapters/Credentials.ts
```diff
@@ -0,0 +1,71 @@
...
+      return true;
+    }
+
+    const now = Date.now();
+    const tokenExpDate = this.getTokenExpDate(accessToken);
+
+    return now >= tokenExpDate - TOKEN_EXP_THRESHOLD;
+  }
+
+  getAccessTokenExpDate(): number | null {
```
cesarenaldi, line 54: It says date but seems a timestamp.

krzysu: that method was removed but I renamed the others to `getTokenExpTimestamp`

packages/client/src/authentication/Authentication.ts
```diff
@@ -1,39 +1,61 @@
...
-
-    return result.data.result;
+  async isAuthenticated(): Promise<boolean> {
+    const credentials = await this.storage.get();
+    return credentials ? !credentials.isExpired() : false;
   }
 
-  async refresh(refreshToken: string): Promise<Credentials> {
-    const result = await this.sdk.AuthRefresh({ refreshToken });
+  async getRequestHeader(): PromiseResult<
```
krzysu, line 32: this is the core of this PR, a single method, called before each request that requires auth, see Reactions.ts how it is used. It is a lazy way of checking expiration or refreshing token. And will call auth API only if required.

packages/client/src/reactions/Reactions.ts
```diff
@@ -1,37 +1,47 @@
...
 import { NotAuthenticatedError } from '../consts/errors';
 import { ReactionRequest } from '../graphql/types.generated';
 import { getSdk, Sdk } from './graphql/reactions.generated';
 
 export class Reactions {
+  private readonly authentication: Authentication | undefined;
   private readonly sdk: Sdk;
 
-  constructor(config: LensConfig, private readonly credentials?: Credentials) {
+  constructor(config: LensConfig, authentication?: Authentication) {
```
krzysu, line 13: `Authentication` is optional, we can have modules that contain both auth-required and public methods and inform user (currently by throwing an error) if misused.

packages/client/TODO.md
```diff
@@ -16,7 +16,7 @@
 
 Client manages the token lifecycle and storage
 
-- LensClient.init({ environment, storage })
+- new LensClient({ environment, storage? })
```
krzysu, line 19: we decided that there is no need for a static constructor and `new` is a more obvious way

packages/client/src/LensClient.spec.ts
```diff
@@ -6,34 +6,30 @@ const testConfig = {
...
+    });
 
-  //     await expect(client.authenticate(address, signature)).resolves.not.toThrow();
-  //   });
-  // });
+    it(`should offer ${LensClient.prototype.authenticate.name} function`, async () => {
+      expect(typeof client.authenticate).toEqual('function');
+    });
+
+    it(`should offer ${LensClient.prototype.isAuthenticated.name} function`, async () => {
```
krzysu, line 20: LensClient.spec is only testing the interface exposed to end-user

packages/client/src/authentication/__helpers__/setupAuthentication.ts
```diff
@@ -0,0 +1,28 @@
...
+
+let auth: Authentication;
+
+export async function setupAuthentication(): Promise<Authentication> {
+  if (auth) {
+    return auth;
+  }
+
+  // const wallet = Wallet.createRandom();
+  const wallet = setupTestWallet();
```
krzysu, line 19: there is a new ticket to rethink how we do tests in this package https://avara.height.app/lens-sdk/T-6018

packages/client/src/authentication/adapters/Credentials.ts
```diff
@@ -0,0 +1,38 @@
+import { DateUtils, invariant } from '@lens-protocol/shared-kernel';
+import jwtDecode, { JwtPayload } from 'jwt-decode';
+
+// Threshold in seconds that will mark token as expired even it's still valid
+// Adds some time for all communications that's required to refresh tokens
+const TOKEN_EXP_THRESHOLD = DateUtils.secondsToMs(3);
```
cesarenaldi, line 6: Should we make this 30 seconds?... seems relatively narrow margin.

packages/client/src/authentication/adapters/Credentials.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export class Credentials {
+  constructor(readonly accessToken: string | undefined, readonly refreshToken: string) {}
+
+  canRefresh(): boolean {
+    const now = Date.now();
```
cesarenaldi, line 12: Thinking more about this, how does this work if client clock is skewed?

packages/client/src/reactions/Reactions.ts
```diff
@@ -1,37 +1,47 @@
...
-    if (!this.credentials) {
+    if (!this.authentication) {
       throw new NotAuthenticatedError();
     }
 
-    return this.sdk.AddReaction(request);
+    const headerResult = await this.authentication.getRequestHeader();
+
+    if (headerResult.isFailure()) {
+      throw headerResult.unwrap();
```
cesarenaldi, line 28: In case the `Result<T, E>` is ` Failure<T, E>`, `unwrap already throws the inner error, so technically you are just doing all of this with line 31.

I think we should have a cohesive narrative around known failure modes throughout the SDK. As we are exposing `Result<T, E>` in the "write hooks", I would strongly entertain the idea of doing the same here. It will allow for a more explicit error handling which can help with SDK upgrades (imagine a new error being returned or removed). Consumer's code if written in TS could be automatically informed about the change by simple TS and I would say very self-descriptive errors.

packages/client/src/authentication/Authentication.ts
cesarenaldi: At the moment `Authentication` is an internal of the client SDK. If the plane of exposing individual submodules to very advanced users is still in place, this method should be segregated behind a privileged interface that only submodules can read.

krzysu: yes, I want to keep the option to use submodules independently open. I'm using that in submodule tests.
how would the privileged interface look like (is it just ts definition?) and would it be possible to access the `getRequestHeader` method in tests?

packages/client/src/authentication/adapters/Credentials.ts
krzysu: it won't work but I don't know how we can go around it

cesarenaldi: **TL;DR;**
I would say let's compare clock time with `iat` timestamp when we get the credentials the first time (and at every renew time) and if the time discrepancy is too large we can throw a dedicated `ClockSkewedError` so the integrator can do something about it.

There are few strategies that can work but relies on assumptions on the number of processes/threads and the storage.

More or less all based on storing the client time at the same time of the credentials and use relative difference and the server knowledge of current time at the time the credentials has been issued (i.e. the `iat` claim in the Access Token JTW is a good one).

But it falls short if you consider the low-level client is designed to work on many scenarios/enviornments. From browser with local storage as credentials storage to a server side with multiple instances of the client process with a shared storage (e.g. think about a dockerized process with a shared redis as storage). In the latter all client process must have clock in sync, which probably is fine in most docker/containerized environments, but is an implicit assumption nonetheless.

packages/client/src/reactions/Reactions.ts
krzysu: I changed it to return `PromiseResult` but it's quite a lot of code, especially to change one definition of `PromiseResult` into another (with catching errors while unwrapping to wrap them again). Is there any better way to do it?

```
async add(
  request: ReactionRequest,
): PromiseResult<void, CredentialsExpiredError | NotAuthenticatedError | Error> {
  if (!this.authentication) {
    return failure(new NotAuthenticatedError());
  }

  const headerResult = await this.authentication.getRequestHeader();

  if (headerResult.isFailure()) {
    try {
      headerResult.unwrap();
    } catch (error) {
      assertError(error);
      return failure(error);
    }
  }

  try {
    await this.sdk.AddReaction(request, headerResult.unwrap());
  } catch (error) {
    assertError(error);
    return failure(error);
  }

  return success();
}
```

packages/client/src/authentication/adapters/Credentials.ts
krzysu: I added check if clock is not skewed, based on system time and token issue time (https://github.com/lens-protocol/lens-sdk/pull/148/commits/ec446db89d23f33bc88f00120a8ab19f7c1e031f)
but it makes tests harder, I skipped two tests for now, I guess I have to introduce mocked auth server response to have control over the credentials returned.

packages/client/src/authentication/adapters/AuthenticationApi.ts
```diff
@@ -0,0 +1,46 @@
...
+
+    return credentials;
+  }
+
+  async refresh(refreshToken: string): Promise<Credentials> {
+    const result = await this.sdk.AuthRefresh({ refreshToken });
+    const { accessToken: newAccessToken, refreshToken: newRefreshToken } = result.data.result;
+
+    const credentials = new Credentials(newAccessToken, newRefreshToken);
+    credentials.checkClock();
```
krzysu, line 42: I'm checking local system clock vs. api clock on auth and refresh

packages/client/src/authentication/adapters/Credentials.ts
krzysu: I added `checkClock` method on Credentials class, it makes testing with useFakeTimers harder :)

packages/client/src/reactions/Reactions.ts
cesarenaldi: I see, yeah you don't need to unwrap it at all if you are to use the result.

```ts
async add(
  request: ReactionRequest,
): PromiseResult<void, CredentialsExpiredError | NotAuthenticatedError | Error> {
  if (!this.authentication) {
    return failure(new NotAuthenticatedError());
  }

  const headerResult = await this.authentication.getRequestHeader();

  if (headerResult.isFailure()) {
    return failure(error);
  }

  await this.sdk.AddReaction(request, headerResult.value); // better not unwrap so if the `if` above gets mistakenly removed you get a TS error cause the result was not narrowed down to a Success<T, E> and `.value` becomes illegal.
  return success();
}
```

cesarenaldi: I suspect after some iterations a code re-use pattern will emerge.

krzysu: check inline comments
```ts
// in your example above
if (headerResult.isFailure()) {
  return failure(error); // where error is coming from? 
}

```

```ts
try {
  // API call can also throw (ie. on connect issue), then the whole ResultPromise pattern doesn't make sense if it can still throw
  await this.sdk.AddReaction(request, headerResult.unwrap());
} catch (error) {
  assertError(error);
  return failure(error);
}
```

```

File: packages/client/src/authentication/Authentication.spec.ts [hunk 11]
```diff
@@ -1,19 +1,89 @@
+import { DateUtils } from '@lens-protocol/shared-kernel';
+
 import { mumbaiSandbox } from '../consts/environments';
+import { CredentialsExpiredError, NotAuthenticatedError } from '../consts/errors';
 import { Authentication } from './Authentication';
+import { setupRandomAuthentication } from './__helpers__/setupAuthentication';
```
krzysu, line 6: `setupRandomAuthentication` means using `ethers.Wallet.random()` and not the one from env var. I'm trying to promote the usage of random wallets in tests and deprecate the need to share private key.

packages/client/src/authentication/Authentication.spec.ts
```diff
@@ -1,19 +1,89 @@
...
+        const result = await auth.getRequestHeader();
+
+        expect(result.isSuccess()).toBeTruthy();
+        expect(result.unwrap()).toMatchObject({
+          authorization: expect.any(String),
+        });
+      });
+    });
+
+    describe.skip(`and credentials are expired but can be refreshed`, () => {
```
krzysu, line 78: I had to skip this test after introducing ClockSkew check, not sure how to solve it

packages/client/src/reactions/Reactions.ts
```diff
@@ -1,37 +1,76 @@
...
     this.sdk = getSdk(client);
+    this.authentication = authentication;
   }
 
-  async add(request: ReactionRequest) {
-    if (!this.credentials) {
-      throw new NotAuthenticatedError();
+  async add(
+    request: ReactionRequest,
+  ): PromiseResult<void, CredentialsExpiredError | NotAuthenticatedError | Error> {
```
krzysu, line 23: the amount of code to return PromiseResult from the action method doesn't look good to me.

packages/client/src/LensClient.spec.ts
cesarenaldi: I don't know how much value this type of tests bring... but in general this class of tests could also be done with TS typedef tests. It would test not just the presence of functions but also their arguments and return type.

I am not suggesting to action this in this PR.

packages/client/src/authentication/Authentication.spec.ts
cesarenaldi: I see, I think the name is somewhat confusing, hence my question. Maybe we can think of something more clear as we iterate over other features.

packages/client/src/authentication/Authentication.spec.ts
cesarenaldi: Yeah, probably requires to mock the first auth response.

packages/client/src/authentication/Authentication.ts
cesarenaldi: Let's cross that bridge once we are in a position to export `Authentication`. Can you please keep track of this aspect so we can have that conversation later?

packages/client/src/authentication/adapters/Credentials.ts
```diff
@@ -0,0 +1,52 @@
+import { DateUtils, invariant } from '@lens-protocol/shared-kernel';
+import jwtDecode, { JwtPayload } from 'jwt-decode';
+
+import { ClockSkewedError } from '../../consts/errors';
```
cesarenaldi, line 4: nit: I don't think grouping things by their types scales much. Better to colocate things that change at the same time and for the same reason. The error should live in this module IMO.

cesarenaldi (approved):

packages/client/src/reactions/Reactions.ts
cesarenaldi: Sorry the typo, the error is from the result:
```ts
if (headerResult.isFailure()) {
  return failure(headerResult.error);
}
```
**Side note:** as we do this relatively often an TS requires a new `failure` to reconcile the return type in case of `T` (in `Result<T, E>` mismatch).. we can consider a `forward` method in `Result`. It's something I saw in one `Result` implementation (cannot recall which one) but it was very neat.

The return line would change into:
```ts
return headingResult.forward()
```

Food for thoughts.

packages/client/src/authentication/Authentication.ts
krzysu: task created for the `IAuthentication` public interface https://avara.height.app/lens-sdk/T-6208

packages/client/src/LensClient.spec.ts
krzysu: they will be removed in https://avara.height.app/lens-sdk/T-6208

packages/client/src/reactions/Reactions.ts
krzysu: done as discussed, thx!

packages/client/src/authentication/Authentication.spec.ts
krzysu: we have a ticket to rethink testing strategy, let's do it there https://avara.height.app/lens-sdk/T-6018

PR #149, opened 2023-02-07
feat: consolidates write hooks signatures

cesarenaldi:


packages/react/src/helpers.ts
```diff
@@ -110,3 +112,43 @@ export function usePaginatedReadResult<
...
+
+export function useOperation<TResult, TError extends IEquatableError, TArgs extends unknown[]>(
+  handler: OperationHandler<TResult, TError, TArgs>,
+): Operation<TResult, TError, TArgs> {
+  const [isPending, setIsPending] = useState(false);
+  const [error, setError] = useState<TError | undefined>(undefined);
+
+  return {
+    error,
+    execute: async (...args: TArgs) => {
```
reecejohnson, line 136: So all operations will be invoked using `execute`?

I did like the more descriptive name in same cases, it helped with readability at the call site.

This is similar to where we label everything `result` for the reads, so happy to go with this.

reecejohnson (approved):
Looks awesome! Left one comment 👍

packages/react/src/helpers.ts
krzysu: in the docs we can use this pattern to make it more descriptive

```
const {execute: follow, error, isPending } = useFollow()
```

krzysu (approved):

PR #150, opened 2023-02-08
feat: client SDK - add search module

paulxuca:
Let me know If I'm missing anything super obvious here!

### DIFF
--- a/packages/client/src/search/Search.ts
+++ b/packages/client/src/search/Search.ts
@@ -1,0 +1,27 @@
+import { GraphQLClient } from 'graphql-request';
+
+import { LensConfig } from '../consts/config';
+import {
+  getSdk,
+  Sdk,
+  SearchProfilesQueryVariables,
+  SearchPublicationsQueryVariables,
+} from './graphql/search.generated';
+
+export class Search {
+  private readonly sdk: Sdk;
+
+  constructor(config: LensConfig) {
+    const client = new GraphQLClient(config.environment.gqlEndpoint);
+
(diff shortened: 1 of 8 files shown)


packages/client/src/search/Search.ts
```diff
@@ -0,0 +1,27 @@
...
+    const client = new GraphQLClient(config.environment.gqlEndpoint);
+
+    this.sdk = getSdk(client);
+  }
+
+  async searchProfiles(request: SearchProfilesQueryVariables) {
+    return this.sdk.SearchProfiles(request);
+  }
+
+  async remove(request: SearchPublicationsQueryVariables) {
```
krzysu, line 24: the function name is wrong :wink:

krzysu:
nice! 👏 
exactly, adding submodules for "read" operations is super obvious. And I hope it will be the same with "writes" soon.
Regarding this PR, I think we should add some tests. On `LensClient`: just a simple check that the Search class is exposed on `client.search` and for the `Search` module as a simple integration test using real API. We were talking about running tests against mumbaiSandbox api before with @cesarenaldi and @desfero.
I'd always cover all user-facing (exported) methods a least with a test that the method is available to never break the contract with the user.

krzysu (approved):
I'd add one more test in LensClient.spec.ts

```
describe(`when accessing the Search module`, () => {
    it(`should return a new instance of Search`, async () => {
      expect(client.search).toBeInstanceOf(Search);
    });
  });
```

the rest looks great!

PR #151, opened 2023-02-08
feat: client SDK - add explore module 

paulxuca:


packages/client/src/explore/Explore.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+
+      await expect(
+        explore.explorePublications({
+          sortCriteria: PublicationSortCriteria.TopCommented,
+        }),
+      ).resolves.not.toThrow();
+    });
+  });
+
+  describe(`when the instance is authenticated and method ${Explore.prototype.exploreProfiles.name} is called`, () => {
```
krzysu, line 22: there is sth wrong with the descriptions here and above 😄

packages/client/src/explore/graphql/explore.graphql
```diff
@@ -0,0 +1,33 @@
...
+        ...Comment
+      }
+    }
+    pageInfo {
+      ...CommonPaginatedResultInfo
+    }
+  }
+}
+
+query ExploreProfiles($observerId: ProfileId, $limit: LimitScalar!, $cursor: Cursor) {
```
krzysu, line 22: I think in this low level SDK we should just expose all what API accepts, so `ExploreProfilesRequest` in this case

packages/client/src/explore/Explore.ts
```diff
@@ -0,0 +1,27 @@
...
+    const client = new GraphQLClient(config.environment.gqlEndpoint);
+
+    this.sdk = getSdk(client);
+  }
+
+  async explorePublications(request: ExplorePublicationRequest) {
+    return this.sdk.ExplorePublications({ request });
+  }
+
+  async exploreProfiles(request: ExploreProfilesRequest) {
```
krzysu, line 21: think about pagination helper like we have `PaginatedReadResult` in react package

paulxuca: Adding it as a separate PR to explore first 👍

krzysu (approved):
LGTM!

PR #152, opened 2023-02-08
feat client sdk - Paginated helper

paulxuca:


packages/client/src/helpers.ts
```diff
@@ -0,0 +1,55 @@
...
+  prev(): Promise<PaginatedResult<T> | null>;
+};
+
+export async function buildPaginatedQueryResult<
+  V,
+  T extends PaginatedQueryData<K>,
+  K = InferPaginatedItemsType<T>,
+>(queryFn: (variables: V) => Promise<K>, variables: V): Promise<PaginatedResult<K>> {
+  const result = await queryFn(variables);
+  const paginatedResult = result as PaginatedQueryData;
```
paulxuca, line 29: @krzysu I couldn't figure out why `result` wasn't being typed as being extending `PaginatedQueryData` to use in the `next` and `prev` functions. Let me know if you have any ideas!

packages/client/src/helpers.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+          },
+        },
+      },
+    };
+
+    const result = await buildPaginatedQueryResult(async () => response, {});
+
+    expect(result.data).toMatchObject(response.data);
+
+    expect(true).toBe(true);
```
paulxuca, line 43: woops

packages/client/src/helpers.ts
desfero: Would refactor to something like below so types are nicely maintained
```

type PaginatedQueryData<Item> = {
  pageInfo: CommonPaginatedResultInfoFragment;
  items: Item[];
};

export type PaginatedResult<T> = PaginatedQueryData<T> & {
  next(): Promise<PaginatedResult<T> | null>;
  prev(): Promise<PaginatedResult<T> | null>;
};

export async function buildPaginatedQueryResult<V, R>(
  queryFn: (variables: V) => Promise<PaginatedQueryData<R>>,
  variables: V,
): Promise<PaginatedResult<R>> {
  const result = await queryFn(variables);

  return {
    ...result,
    async next() {
      if (result.pageInfo.next) {
        return buildPaginatedQueryResult(queryFn, {
          ...variables,
          cursor: result.pageInfo.next,
        });
      }

      return null;
    },

    async prev() {
      if (result.pageInfo.prev) {
        return buildPaginatedQueryResult(queryFn, {
          ...variables,
          cursor: result.pageInfo.prev,
        });
      }

      return null;
    },
  };
}
```

packages/client/src/search/Search.spec.ts
```diff
@@ -17,6 +17,22 @@ describe(`Given the ${Search.name} configured to work with sandbox`, () => {
...
+      const search = new Search(testConfig);
+
+      const result = await search.searchProfiles({
+        query: 'test',
+        limit: 10,
+      });
+
+      expect(result.items.length).not.toBe(0);
+
+      const nextResult = await result.next();
```
desfero, line 31: nit: an open question given we are modeling here the public api:

Can we assume that no one would try to save a reference to the first `result` and try to always call `next` from there? 
Below is a potential pseudo code:

```
const list = await search.searchProfiles({
        query: 'test',
        limit: 10,
      }); 

const secondPage = await list.next();
const thirdPage =  await list.next(); // noop, still second page really
```

desfero (changes requested):

packages/client/src/helpers.spec.ts
```diff
@@ -0,0 +1,38 @@
+import { CommonPaginatedResultInfoFragment } from './graphql/fragments.generated';
+import { buildPaginatedQueryResult } from './helpers';
+
+describe('Given a query function and request for the paginated helper', () => {
+  it('should return the next page of results if there are more', async () => {
+    const response = {
```
krzysu, line 6: are these tests actually testing anything? 😄

packages/client/src/helpers.ts
```diff
@@ -0,0 +1,58 @@
...
+    },
+
+    async next() {
+      if (result.pageInfo.next) {
+        const nextResult = await buildPaginatedQueryResult(queryFn, {
+          ...variables,
+          cursor: result.pageInfo.next,
+        });
+
+        result = nextResult;
```
desfero, line 35: i like it that you can now call it both ways, can we add unit tests to cover that behavior?

krzysu (approved):
now we need a follow-up task to wrap all paginated queries with this helper. 😄 

could you also add an example usage to the README file? we will move them later to the proper documentation page but it will be easy if we have some use cases written down as we build them. The use-case could be sth like "Search for profiles and browse through the results" that will showcase the usage of `.next()` method.

desfero (approved):
LGTM 🚢

PR #154, opened 2023-02-09
feat: enhance typing for dealing with files

reecejohnson:
- Consolidates file type typings into a single file
- Exports types in `@lens-protocol/react`
- Replaces duplicated `ImageType` in example app
- Removes unused file type helpers


krzysu (approved):
LGTM

PR #155, opened 2023-02-09
feat: client sdk - add revenue module

paulxuca:


packages/client/src/LensClient.spec.ts
```diff
@@ -1,5 +1,6 @@
 import LensClient, { Profile, Reactions, Search, Explore } from '.';
 import { mumbaiSandbox } from './consts/environments';
+import { Revenue } from './revenue';
```
krzysu, line 3: this should be exported from '.' just to make sure it is also exported from the package

packages/client/src/revenue/Revenue.ts
```diff
@@ -0,0 +1,37 @@
...
+    return result.data.result;
+  }
+
+  async profileFollowRevenue(request: ProfileFollowRevenueQueryRequest) {
+    const result = await this.sdk.ProfileFollowRevenue({ request });
+
+    return result.data.result;
+  }
+
+  async publicationRevenue(request: PublicationRevenueQueryRequest) {
```
krzysu, line 32: shall we add return types on the methods here to make it obvious what is returned?

packages/client/src/graphql/fragments.graphql
```diff
@@ -420,3 +420,36 @@ fragment CommonPaginatedResultInfo on PaginatedResultInfo {
...
+
+    ... on Comment {
+      ...Comment
+    }
+  }
+
+  ...Revenue
+}
+
+fragment Revenue on PublicationRevenue {
```
krzysu, line 450: are these fragments used anywhere outside of the `revenue` module? if not, then let's keep them together with queries in the modules .graphql file

krzysu (approved):
nice!

PR #156, opened 2023-02-09
feat: client sdk - add remaining read profile methods 

paulxuca:


packages/client/src/profile/Profile.ts
```diff
@@ -24,4 +28,30 @@ export class Profile {
...
+
+  async recommendedProfiles(observerId?: string) {
+    const result = await this.sdk.RecommendedProfiles({
+      observerId,
+    });
+
+    return result.data.result;
+  }
+
+  async mutualFollowersProfiles(request: MutualFollowersProfilesQueryRequest, observerId?: string) {
```
krzysu, line 49: the method names in TODO are shorter to avoid repeating "profile"
```
- [Q] client.profile.fetch(request) (by id or handle, forSources) or client.profiles.fetchOne()
- [PQ] client.profile.fetchAll(request) (by ids, ownedBy, handles, whoMirroredPublicationId)
- [Q] client.profile.fetchRecommended()
- [PQ] client.profile.mutualFollowers({ viewingProfileId, yourProfileId })

```

packages/client/codegen-api.yml
```diff
@@ -72,6 +72,8 @@ generates:
       extension: .generated.ts
       baseTypesPath: src/graphql/types.generated.ts
     plugins:
+      - add:
+          content: '// @ts-nocheck'
```
krzysu, line 76: I think there is one too much white space here as for a yml file (I know it works but it's not formated correctly, is it?)

krzysu:
inside profile module you could also add

```
- [Q] client.profile.doesFollow(request)
- [PQ] client.profile(s).(all)following({ address })
- [PQ] client.profile.fetchFollowers({ profileId })
- [Q] client.profile.followerNftOwnedTokenIds({ address, profileId })
```

we need to think about consistency in naming

or we go with (singular) `profile` and then `profile.fetch` and `profile.fetchAll`
or `profiles` and then `profiles.fetchOne` and profiles.fetch` 

so 
`client.profile.allFollowing()` // returns collection of profiles
`client.profiles.following()` // also return collection
the name should tell you if you get single or multiple (possibly paginated) results

packages/client/codegen-api.yml
paulxuca: do you mean between the `//` and the `@ts-nocheck`?

krzysu: oh, my bad, all good, sorry
I thought it should be 

```
- add:
  content: '// @ts-nocheck'
```
but I'm wrong

maybe wrapping add in single quotes would help with consistency but whatever :) 

`- 'add':`

packages/client/src/profile/Profile.ts
```diff
@@ -24,4 +32,64 @@ export class Profile {
...
+  async fetchAll(request: ProfileQueryRequest, observerId?: string) {
+    const result = await this.sdk.Profiles({
+      request,
+      observerId,
+    });
+
+    return result.data.result;
+  }
+
+  async allRecommended(observerId?: string) {
```
krzysu, line 45: please, let's be explicit with return types for all sdk methods

krzysu (changes requested):
could you also update `client/TODO.md` with new method names and mark them as done to keep better overview of work left to be done? (mark as `[x]`)

packages/client/TODO.md
```diff
@@ -53,10 +53,10 @@ Client manages the token lifecycle and storage
 
 ### Query profile
 
-- [Q] client.profile.fetch(request) (by id or handle, forSources) or client.profiles.fetchOne()
-- [PQ] client.profile.fetchAll(request) (by ids, ownedBy, handles, whoMirroredPublicationId)
-- [Q] client.profile.fetchRecommended()
-- [PQ] client.profile.mutualFollowers({ viewingProfileId, yourProfileId })
+- [x] client.profile.fetch(request) (by id or handle, forSources) or client.profiles.fetchOne()
+- [x] client.profile.fetchAll(request) (by ids, ownedBy, handles, whoMirroredPublicationId)
+- [x] client.profile.fetchRecommended()
```
krzysu, line 58: the method you added is called `allRecommended` (not `fetchRecommended`) and there are more changes like that, which is ok, but could you also update to the new naming in the TODO file?

packages/client/TODO.md
```diff
@@ -84,10 +84,10 @@ Client manages the token lifecycle and storage
 - [AM] client.profile.createSetFollowNFTUriTypedData(request)
 - [APQ] client.profile.pendingApprovalFollows(request)
 
-- [Q] client.profile.doesFollow(request)
-- [PQ] client.profile(s).following({ address })
-- [PQ] client.profile.fetchFollowers({ profileId })
-- [Q] client.profile.followerNftOwnedTokenIds({ address, profileId })
+- [x] client.profile.doesFollow(request)
+- [x] client.profile(s).allFollowing({ address })
```
krzysu, line 88: for the future leave the annotations in [] like [PQ], which will tell us which methods require auth or pagination etc.

PR #158, opened 2023-02-10
draft: amends useFollow to leverage useOperation

cesarenaldi:


examples/web-wagmi/src/profiles/UseFollowAndUnfollow.tsx
```diff
@@ -9,46 +13,65 @@ import { Loading } from '../components/loading/Loading';
...
-
-function ProfileFollow({ profile }: ProfileFollowProps) {
-  return (
-    <article>
-      <ProfileCard profile={profile} />
-      <FollowButton profile={profile} />
-    </article>
+    <>
+      <button
+        onClick={follow}
```
krzysu, line 58: I remember a discussion on discord to allow to pass profile to follow to the "execute" method, not to define it all as a hook args. is that not relevant anymore?

krzysu:
I see this in the console when running the example app and following a profile, and I remember fixing it once in bean webapp.

Screenshot 2023-02-14 at 09 25 41

krzysu (approved):
the UX feels less optimistic than before 😉 but it's more accurate to what is possible in a certain moment.

examples/web-wagmi/src/profiles/UseFollowAndUnfollow.tsx
```diff
@@ -9,46 +13,65 @@ import { Loading } from '../components/loading/Loading';
 import { ProfileCard } from './components/ProfileCard';
 
 type FollowButtonProps = {
-  profile: ProfileFragment;
+  follower: ProfileOwnedByMeFragment;
+  followee: ProfileFragment;
```
reecejohnson, line 17: nice terminology here, `followee` works well

packages/api-bindings/src/graphql/utils/profile.ts
```diff
@@ -0,0 +1,9 @@
+import { Overwrite } from '@lens-protocol/shared-kernel';
+
+import { ProfileFragment } from '../generated';
+
+export type ProfileOwnedByMeFragment = Overwrite<ProfileFragment, { ownedByMe: true }>;
+
+export function isProfileOwnedByMe(profile: ProfileFragment): profile is ProfileOwnedByMeFragment {
```
reecejohnson, line 7: nice 👌

packages/react/src/transactions/useFollow.ts
```diff
@@ -1,126 +1,109 @@
...
+  const hasPendingUnfollowTx = useHasPendingTransaction(
+    isUnfollowTransactionFor({ profileId: followee.id }),
+  );
 
-        if (result.isFailure()) {
-          setError(result.error);
-        }
-      } finally {
-        setIsPending(false);
+  return useOperation(
```
reecejohnson, line 87: `useOperation` looking clean in the wild

examples/web-wagmi/src/profiles/UseFollowAndUnfollow.tsx
cesarenaldi: It depends... in general the item that is the subject of the operation has to be passed as hook arg cause the `isPending` will be relevant for the item. This is the case of `followee` in `useFollow`, `publication` in `useCollect` and so on.
Passing the subject as callback arg will need to figure out another way to communicate "pending status for the given item", and why need to do things in 2 ways?!


Some hooks requires a profile as the "actor" performing the operation which is so far always one of the user's profiles, typed as `ProfileOwnedByMeFragment` (I really hate the `Fragment` suffix in those types, but that's for another day).
This is the case of `follower` in `useFollow`, `collect` in `useCollect` (see following PR), etc.

The reason the `profile` is passed as hook arg is to guide the user to make sure they have in scope a profile they own and not just a random profile id. From a pure TS perspective, if you look at the hooks that gives you instances `ProfileOwnedByMeFragment` you are automatically drawn to  use `useActiveProfile` (which is the right thing to do).

The integrator, if they want, can still figure out if they own a random profile from a list via the `isProfileOwnedByMe`, which BTW I just checked and realized I am not re-exporting from `@lens-protocol/react` so going to do it in a follow PR.

packages/react/src/transactions/useFollow.ts
```diff
@@ -1,126 +1,109 @@
...
+    async (): PromiseResult<
+      void,
+      | InsufficientAllowanceError
```
reecejohnson, line 90: nit: could be worth conflating the errors here and on line 70 into a single type to avoid duplication

e.g
```typescript
type FollowErrors =
  | InsufficientAllowanceError
  | InsufficientFundsError
  | PendingSigningRequestError
  | PrematureFollowError
  | UserRejectedError
  | WalletConnectionError;
  ```

**Comment by cesarenaldi** [comment]
> I see this in the console when running the example app and following a profile, and I remember fixing it once in bean webapp.
> 
> Screenshot 2023-02-14 at 09 25 41

I saw this as well, it's a pre-existing issue. The GQL node involved are not affected by it as they are ephemeral objects. Still would be nice to get rid of the warning.

File: packages/react/src/transactions/useUnfollow.ts [hunk 6]
```diff
@@ -1,61 +1,70 @@
...
 
-  return {
-    unfollow: async () => {
-      if (!profile.isFollowedByMe && profile.isOptimisticFollowedByMe) {
-        setError(
-          new PrematureUnfollowError(`Your follow request for ${profile.handle} is still pending.`),
+  return useOperation(
+    async (): PromiseResult<
+      void,
+      | PendingSigningRequestError
```
reecejohnson, line 45: nit: same thought here for an reusable union of potential errors as in `useFollow`

reecejohnson (approved):
Looking awesome, `useOperation` is a lovely little abstraction.

Approved and left some small nit comments 👍

cesarenaldi:
> the UX feels less optimistic than before 😉 but it's more accurate to what is possible in a certain moment.

Technically is the same.... maybe disabling the button feels odd I know. Unfortunately there is no way to make a follow go through while an unfollow is in progress and the other way around. API validation will scream at us. Maybe from an example perspective we could rework it so that the buttons are not disabled but a warning is shown if pressed too early.

packages/react/src/transactions/useFollow.ts
cesarenaldi: I gave it a try but seem it hides the union in IDE suggestions at the call site:
image

which IMO is counterproductive. We are exposing failure modes so to increase awareness of them.

I understand the repetition but don't have a better suggestion than keep the union.

PR #159, opened 2023-02-10
release #4

reecejohnson:
Release number four.

### DIFF
--- a/packages/client/CHANGELOG.md
+++ b/packages/client/CHANGELOG.md
@@ -1,8 +1,0 @@
-# @lens-protocol/client
-
-## 0.1.0
-
-### Patch Changes
-
-- @lens-protocol/shared-kernel@0.4.0
-- @lens-protocol/storage@0.4.0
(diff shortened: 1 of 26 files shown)


packages/client/CHANGELOG.md
```diff
@@ -0,0 +1,8 @@
+# @lens-protocol/client
+
+## 0.1.0
+
+### Patch Changes
+
+- @lens-protocol/shared-kernel@0.4.0
+- @lens-protocol/storage@0.4.0
```
cesarenaldi, line 8: I don't think we should check in this (yet).

packages/sdk-gated/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # 1.2.0 - 20th January 2023
 
+## 1.2.0
+
 - Remove `cross-blob` and `cross-fetch` dependencies
 - Package now only supports Node 18+.
```
cesarenaldi, line 7: Same here, this is still private (not published from this repo yet).

cesarenaldi:
Beyond those points LGTM. Approving ahead of time so to move this fwd.

cesarenaldi (approved):

PR #160, opened 2023-02-13
feat: client sdk - unify auth interface, add execute helper, search and explore modules improvements

krzysu:
- unify auth interface (`IAuthentication`), 
- add `execute` helper for API calls that require authentication 
- search and explore modules improvements - correct method names and return values

### DIFF
--- a/packages/client/src/LensClient.spec.ts
+++ b/packages/client/src/LensClient.spec.ts
@@ -8,20 +8,6 @@ const testConfig = {
 describe(`Given the LensClient configured for sandbox`, () => {
   const client = new LensClient(testConfig);
 
-  describe(`when checking authentication features exposed by the client`, () => {
-    it(`should offer ${LensClient.prototype.generateChallenge.name} function`, async () => {
-      expect(typeof client.generateChallenge).toEqual('function');
-    });
-
-    it(`should offer ${LensClient.prototype.authenticate.name} function`, async () => {
-      expect(typeof client.authenticate).toEqual('function');
-    });
-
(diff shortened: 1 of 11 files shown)


packages/client/src/LensClient.spec.ts
```diff
@@ -8,20 +8,6 @@ const testConfig = {
...
-  describe(`when checking authentication features exposed by the client`, () => {
-    it(`should offer ${LensClient.prototype.generateChallenge.name} function`, async () => {
-      expect(typeof client.generateChallenge).toEqual('function');
-    });
-
-    it(`should offer ${LensClient.prototype.authenticate.name} function`, async () => {
-      expect(typeof client.authenticate).toEqual('function');
-    });
-
-    it(`should offer ${LensClient.prototype.isAuthenticated.name} function`, async () => {
```
krzysu, line 20: with `IAuthentication` interface we don't need these tests anymore

packages/client/src/reactions/Reactions.ts
```diff
@@ -21,36 +22,18 @@ export class Reactions {
...
-    if (headerResult.isFailure()) {
-      return failure(headerResult.error);
-    }
-
-    await this.sdk.AddReaction(request, headerResult.value);
-
-    return success();
+    return execute(this.authentication, async (headers) => {
+      await this.sdk.AddReaction(request, headers);
+      return;
```
paulxuca, line 27: Do we need the explicit return here?

paulxuca (approved):

packages/client/src/reactions/Reactions.ts
krzysu: removed

PR #162, opened 2023-02-13
test pagination helper

krzysu:
here are some ideas regarding pagination helper, the type is not working properly I think, otherwise it would detect the issue that I tried to solve in this PR.
also as you can see, the "real" test for pagination found an issue

this PR is just to show that sth is wrong with the original PR, you can merge if you want or find a better solution


paulxuca:
ah I changed the return types in the other PR which broke it but the types were not working prior to that change as well will look into it more

paulxuca:
Copied to the other open PR, thanks @krzysu 👍

PR #163, opened 2023-02-13
Spread `useOperation` to more write operation hooks

cesarenaldi:
Amends the following hooks to leverage `useOperation` signature
- `useActiveProfileSwitcher`
- `useUpdateProfileImage`
- `useUpdateProfileImage`
- `useApproveModule`
- `useCollect`
- `useCreateMirror` (also leverages `ProfileOwnedByMeFragment` as published of `useCreatePost` and `useCreateComment`)
- `useUpdateDispatcherConfig`
- `useReportPublication`
- `useHidePublication`


packages/react/src/transactions/useCollect.ts
```diff
@@ -1,47 +1,28 @@
...
-          setError(result.error);
-        }
-      } finally {
-        setIsPending(false);
-      }
-    },
-    error,
-    isPending,
-  };
+export function useCollect({ collector, publication }: UseCollectArgs): CollectOperation {
```
krzysu, line 25: I think `collector` should be provided as hook arg but publication as `execute` arg, so the same collector can collect different publications, as in `useCreatePost`

krzysu: as a general rule, we can think about it as an `actor` executing an `action` where actor props we define on a hook but action props we define when calling `execute`.

so collector(profile) is collecting(publication)

krzysu (approved):
in general looks good to me, I see most tests are not affected by these changes so we are not breaking anything, would the docs require an update?

packages/react/src/transactions/useCollect.ts
cesarenaldi: > I think `collector` should be provided as hook arg but publication as `execute` arg, so the same collector can collect different publications, as in `useCreatePost`

The challenge with passing `publication` as callback arg is that the hook has a single `isPending` flag. Without further changes `isPending` will become true on collecting of the first publication and probably disabling the collect button for the entire list.

Sure we can engineer a different strategy to report progress on a case by case (e.g. an "is pending" callback), but why introduce 2 patterns to achieve the same thing if all is need is just nesting `useCollect` closer to the individual list item?

cesarenaldi:
> in general looks good to me, I see most tests are not affected by these changes so we are not breaking anything, would the docs require an update?

Yes, documentation will need some love.

examples/web-wagmi/src/App.tsx
```diff
@@ -72,7 +72,6 @@ const client = createClient({
 const lensConfig: LensConfig = {
   bindings: wagmiBindings(),
   environment: staging,
-  sources: [sources.lenster, sources.orb, 'any-other-app-id'],
```
cesarenaldi, line 75: Removed it for now that [T-4762](https://avara.height.app/T-4762) is not implemented yet as it won't show publications created from within the example app.

packages/react/src/profile/useActiveProfileSwitch.ts
```diff
@@ -1,23 +1,16 @@
...
-    }
-  }
+export function useActiveProfileSwitch(): ActiveProfileSwitchOperation {
+  const switchActiveProfile = useSwitchActiveProfileController();
 
-  return {
-    isPending,
-    switchProfile,
-  };
+  return useOperation(async (profileId: ProfileId) => {
```
reecejohnson, line 12: i really like how this looks at the call site, simplifies the hook a lot 👍

reecejohnson (approved):
nice one 👍

PR #165, opened 2023-02-14
feat: client sdk - transaction and proxy action modules

krzysu:
- add Transaction and ProxyAction modules
- add a few mutations to Profile module
- try different approaches to test authenticated modules - I decided to skip them from this PR and I will add more tests in a follow-up PR


packages/client/README.md
```diff
@@ -59,3 +59,41 @@ const reactionRequest = {
...
+
+### Set a dispatcher with gasless, signed data type
+
+```ts
+const typedDataResult = await lensClient.profile.createSetDispatcherTypedData({
+  profileId: '0x0185', // your profileId
+});
+
+// typedDataResult is a PromiseResult object
+const setDispatcherData = typedDataResult.unwrap().result;
```
krzysu, line 71: all returned from `execute` helper is a `PromiseResult` so requires checks for isSuccess, isFailure, unwrap or just `.value` (if success)

packages/client/src/authentication/__helpers__/setupAuthentication.ts
```diff
@@ -8,38 +8,34 @@ const testConfig = {
...
+    const address = await wallet.getAddress();
+    const challenge = await auth.generateChallenge(address);
+    const signature = await wallet.signMessage(challenge);
 
-  const wallet = setupTestWallet();
-  auth = new Authentication(testConfig);
-  const address = await wallet.getAddress();
-  const challenge = await auth.generateChallenge(address);
-  const signature = await wallet.signMessage(challenge);
+    await auth.authenticate(address, signature);
```
krzysu, line 21: this is executed separately per test file causing race conditions, conflicts with credentials, and API returns "not authenticated" for the tests that were too slow (or too fast). I skipped many tests in this PR and want to fix that in a separate ticket https://avara.height.app/T-6018

packages/client/src/profile/Profile.ts
```diff
@@ -16,16 +19,29 @@ import {
...
+  private readonly authentication: Authentication | undefined;
   private readonly sdk: Sdk;
+  private readonly mutationSdk: MutationSdk;
 
-  constructor(config: LensConfig) {
+  constructor(config: LensConfig, authentication?: Authentication) {
     const client = new GraphQLClient(config.environment.gqlEndpoint);
+
     this.sdk = getSdk(client);
+    this.mutationSdk = getMutationSdk(client);
```
krzysu, line 43: I don't know if it makes sense to separate queries from mutations, wdyt?

packages/client/src/proxy-action/ProxyAction.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+      it(`should return a failure`, async () => {
+        const result = await proxyAction.checkStatus('7624f076-446d-4f38-8277-326b269fe8d8');
+
+        expect(result.isFailure()).toBeTruthy();
+        expect(() => result.unwrap()).toThrow(NotAuthenticatedError);
+      });
+    });
+  });
+
+  describe.skip(`and the instance is authenticated`, () => {
```
krzysu, line 42: notice all authenticated tests are skipped because of the issue mentioned in my other comment

packages/client/src/profile/Profile.ts
paulxuca: What are the benefits of separating it? Adds one more step to rename the same-named imports (And also messes up auto import 😂 )

krzysu: ok, I will keep one sdk then :)

packages/client/README.md
krzysu: yeah, I know, I thought the same in the beginning, then we decided with @cesarenaldi to first separate user-flow-errors (the ones we know that can happen -> result.isFailure()) from unexpected errors (they will still throw) and introduce our users to the concept of `Result` type which is common in other programming languages.

I'm working on the next PR when I'm using this Result type a lot (from an end-user perspective) so I will share my experience later.

packages/client/src/__tests__/AuthenticatedClient.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+  setupTestWallet,
+  testWalletProfileId,
+} from '../authentication/__helpers__/setupTestWallet';
+import { ProxyAction } from '../proxy-action';
+import { Reactions, ReactionTypes } from '../reactions';
+import { Transaction } from '../transaction';
+
+describe(`Given the authenticated LensClient configured to work with sandbox`, () => {
+  const wallet = setupTestWallet();
+  const getClient = setupAuthenticatedClient(wallet);
```
krzysu, line 12: all tests that require authentication in a single file, this way there is one user session for all tests, it's not ideal but it works. any better ideas?

krzysu:
I can also remove all authenticated tests from this PR and move them to the follow-up one with the e2e test setup

packages/client/TODO.md
```diff
@@ -93,11 +93,11 @@ Client manages the token lifecycle and storage
 
 ### Query publications
 
-- [x] [Q] client.publication.fetch(request) or client.publications.fetchOne()
+- [x] [Q] client.publication.fetch(request)
 - [x] [PQ] client.publication.fetchAll(request)
 - [x] [Q] client.publication.validateMetadata(request)
-- [x] [PQ] client.publication.allWhoCollected({ publicationId })
-- [x] [PQ] client.publication(s).allForSale({ profileId })
+- [x] [PQ] client.publication.allWalletsWhoCollected({ publicationId })
```
krzysu, line 99: the returned items are an array of Wallets, not Publications like in other methods in this module

packages/client/src/authentication/Authentication.spec.ts
```diff
@@ -14,6 +14,8 @@ describe(`Given the ${Authentication.name} configured to work with sandbox`, ()
     jest.useRealTimers();
   });
 
+  const getAuthentication = setupRandomAuthentication();
```
krzysu, line 17: random means random wallet address, it will be improved in a follow up PR

paulxuca (approved):

PR #167, opened 2023-02-15
chore: update non breaking dependencies

krzysu:
there are more dependencies to update/upgrade but I will do it in a separate PR
I had to keep previous versions of `@apollo/client` and `@types/jest` as they introduced breaking changes - ts checks were failing


examples/web-wagmi/src/profiles/UseCreateProfile.tsx
```diff
@@ -1,9 +1,9 @@
 import { useCreateProfile, useProfile } from '@lens-protocol/react';
 import { useState } from 'react';
 
+import { ProfileCard } from './components/ProfileCard';
```
reecejohnson, line 4: as discussed, personally, I prefer the order of most far away -> closer

```typescript
import external from 'external'
import { Something } from '../something'
import { SomethingElse } from './something'
```

reecejohnson (approved):

package.json
```diff
@@ -16,7 +16,7 @@
     "dev": "turbo dev",
     "dev:packages": "pnpm preconstruct dev",
     "lint": "turbo run lint --parallel",
-    "lint:fix": "turbo run lint:fix --parallel",
+    "lint:fix": "turbo run lint:fix --parallel --no-cache",
```
cesarenaldi, line 19: 👍

packages/sdk-gated/package.json
```diff
@@ -39,47 +39,48 @@
...
+    "@rollup/plugin-commonjs": "^24.0.1",
     "@rollup/plugin-inject": "^5.0.3",
     "@rollup/plugin-json": "^6.0.0",
     "@rollup/plugin-node-resolve": "^15.0.1",
     "@rollup/plugin-replace": "^5.0.2",
     "@rollup/plugin-terser": "^0.2.1",
     "@rollup/plugin-typescript": "^10.0.1",
-    "@types/jest": "^29.2.3",
-    "@types/node": "^16.11.66",
+    "@types/jest": "29.2.3",
```
cesarenaldi, line 66: is the dropping of `^` intentional?

packages/shared-kernel/package.json
```diff
@@ -40,23 +40,23 @@
...
+    "tslib": "^2.5.0",
     "uuid": "^9.0.0"
   },
   "devDependencies": {
     "@babel/preset-env": "^7.20.2",
     "@babel/preset-typescript": "^7.18.6",
     "@lens-protocol/eslint-config": "workspace:*",
     "@lens-protocol/prettier-config": "workspace:*",
-    "@types/jest": "^29.2.3",
+    "@types/jest": "29.2.3",
```
cesarenaldi, line 51: Same here, is the `^` missing intentional?

cesarenaldi (approved):

packages/shared-kernel/package.json
krzysu: yes, the latest version of types/jest is missing the correct type resolution for `expect.any()` 
I don't know why

packages/eslint-config/package.json
```diff
@@ -4,17 +4,17 @@
...
   "dependencies": {
-    "@typescript-eslint/eslint-plugin": "^5.45.0",
-    "@typescript-eslint/parser": "^5.45.0",
-    "eslint-config-prettier": "^8.3.0",
+    "@typescript-eslint/eslint-plugin": "^5.52.0",
+    "@typescript-eslint/parser": "^5.52.0",
+    "eslint-config-prettier": "^8.6.0",
     "eslint-config-turbo": "0.0.7",
-    "eslint-plugin-import": "^2.26.0"
+    "eslint-plugin-import": "2.26.0"
```
krzysu, line 11: this plugin was changing the order of imports in multiple files, I reverted it

examples/web-wagmi/src/profiles/UseCreateProfile.tsx
krzysu: I reverted update of `eslint-plugin-import` so the order is not changed

PR #168, opened 2023-02-15
feat: use preconstruct to build `client` package and example usage in node

reecejohnson:


packages/client/package.json
```diff
@@ -7,24 +7,15 @@
...
+  "main": "dist/lens-protocol-client.cjs.js",
+  "module": "dist/lens-protocol-client.esm.js",
   "private": true,
   "sideEffects": false,
   "files": [
     "dist/**/*"
   ],
   "scripts": {
-    "build": "pnpm clean && tsup",
-    "clean": "rimraf ./dist",
```
krzysu, line 27: if you removed the `clean` script, do we still need `rimraf` as a dep here?

reecejohnson: no we don't, cheers

krzysu (approved):
thanks!

krzysu:
shall we add @babel/core as a dev dependency everywhere to resolve these? 

Screenshot 2023-02-16 at 10 37 55

PR #169, opened 2023-02-16
featu: adds GQL fragments to fetch gated content details

cesarenaldi:
The idea is that `publication.isGated` and `publication.canObserverDecrypt` can and should be used by the consumer to determine if the observer profile can decrypt the content and if not why so.

The `publication.metadata.__encryptionParams` is not fully type safe as it is. In the following PR I will provide a new field (a local-inly field akin to `FollowPolicy` for `__followModule` or `CollectPolicy` for `__collectModule`) that is more straightforward to use an more importantly type safe. The design of this object is effectively a DSL describing the access criteria. This will inform how the `CreatePostRequest` and `CreateCommetnRequest` request models will change to support creation of gated content.

### DIFF
--- a/packages/api-bindings/codegen-api.yml
+++ b/packages/api-bindings/codegen-api.yml
@@ -15,6 +15,7 @@ config:
     ContractAddress: string
     Cursor: string
     DateTime: string
+    EncryptedValueScalar: string
     EthereumAddress: string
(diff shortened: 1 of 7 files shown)


reecejohnson (approved):
looks good, naming is clear 👍

PR #171, opened 2023-02-16
chore: update more dependencies

krzysu:
let's see if everything still works


examples/web-wagmi/package.json
```diff
@@ -26,19 +26,20 @@
     "react-hot-toast": "^2.4.0",
     "react-router-dom": "^6.8.1",
     "readable-web-to-node-stream": "^3.0.2",
-    "wagmi": "0.9.2"
+    "wagmi": "^0.11.6"
```
cesarenaldi, line 29: I think we need to be careful with wagmi. It's in initial development phase (`0.x`) and they definitely make breaking changes, we experiences one ourself between `0.8.x` and `0.9.x`. Given there is no code changes required on our side between `0.9.2` and `0.11.6` I would consider an range: `">=0.9.2 <=0.11.6"`.

@desfero what do you think?

desfero: agree, we need to narrow the `wagmi` to a specific range that we know works properly inside `@lens-sdk/wagmi` `peerDependencies` configuration.
We probably needs some changeset that's mentioning which important deps were updated

packages/wagmi/package.json
```diff
@@ -44,12 +44,12 @@
...
     "react-dom": "^18.2.0",
     "typescript": "^4.9.5",
-    "wagmi": "^0.11.6"
+    "wagmi": ">=0.9.2 <=0.11.6"
   },
   "peerDependencies": {
     "@lens-protocol/react": "workspace:*",
     "ethers": "^5.7.2",
-    "wagmi": "^0.11.5"
+    "wagmi": ">=0.9.2 <=0.11.6"
```
desfero, line 52: We should have the range only here inside `peerDependencies`, all the other places are considered internal and can use specific versions (with or without `^` prefix).

krzysu: ok 👌

cesarenaldi (approved):
Thank you for looking into this and follow up on review comments.

PR #173, opened 2023-02-16
feat: client sdk - e2e test setup with random wallet

krzysu:


packages/client/src/authentication/__helpers__/signAndBroadcast.ts
```diff
@@ -0,0 +1,40 @@
...
+import { getTxIdFromRelayerResult, Transaction } from '../../transaction';
+
+type TypedData = {
+  domain: TypedDataDomain;
+  types: Record<string, Array<TypedDataField>>;
+  value: {
+    nonce: number;
+    [k: string]: unknown;
+  };
+};
```
cesarenaldi, line 13: nit: this could well go into `@lens-protocol/blockchain-bindings` and be used by both `@lens-protocol/client` and `@lens-protocol/react` packages that way.

packages/client/src/helpers/poll.spec.ts
```diff
@@ -0,0 +1,45 @@
+import { poll } from './poll';
+
+describe('Given a polling function', () => {
+  it('should return when condition is fullfiled', async () => {
```
cesarenaldi, line 4:
```
Given a polling function
when condition is fullfiled
[it] should return ?
```

packages/client/src/helpers/poll.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+      interval: 1,
+      maxAttempts: 3,
+    });
+
+    expect(result).toEqual({
+      indexed: true,
+    });
+  });
+
+  it('should throw if condition is not fullfiled after maxAttempts', async () => {
```
cesarenaldi, line 25:
```
Given the "${poll.name}" helper
when the condition is not fullfilled after the specified maxAttempts
[it] should throw a <?>
````

File: packages/client/src/authentication/__helpers__/setupAuthentication.ts [hunk 4]
```diff
@@ -1,41 +1,203 @@
...
+};
+
+const defaultOptions: SetupOptions = {
+  withNewProfile: false,
+  withDispatcher: false,
+};
+
+const DOES_BURN_PROFILE_WORK = false;
+
+export function setupAuthentication(options?: SetupOptions): () => TestSetup {
```
cesarenaldi, line 33: As it just does not return the authentication module but all relevant data to run authenticated scenarios I would rename this into:

```suggestion
export function setupAuthenticatedScenario(options?: SetupOptions): () => TestSetup {
```

packages/client/src/proxy-action/ProxyAction.ts
```diff
@@ -83,4 +85,21 @@ export class ProxyAction {
...
+    | Result<
+        ProxyActionStatusResultFragment | ProxyActionErrorFragment | ProxyActionQueuedFragment,
+        CredentialsExpiredError | NotAuthenticatedError
+      >
+    | undefined
+  > {
+    return poll({
+      fn: () => this.checkStatus(proxyActionId),
+      validate: (result: Awaited<ReturnType<typeof this.checkStatus>>) => {
+        return getIsStatusCompleteFromProxyActionStatus(result.unwrap());
```
cesarenaldi, line 101: `unwrap` would cause to throw the underling error, is `poll` then catching it and piping it back in the returned result?

packages/client/src/transaction/helpers/index.ts
```diff
@@ -0,0 +1,47 @@
...
+export function getTxIdFromRelayerResult(
+  result: RelayerResultFragment | RelayErrorFragment,
+): string {
+  if (isRelayerResult(result)) {
+    return result.txId;
+  }
+  throw new Error(result.reason);
+}
+
+export function getIsIndexedFromTransactionResult(
```
cesarenaldi, line 40: We have the tendency of naming helpers that take a lump of data and either extract part of it or figure out an a aspect of it (e.g. resolves a discriminated union) with the prefix "get".

These are not technically getters, I think we should be more precise in general: `resolve`, `extract` are better suited in most of the cases in this PR.

packages/client/src/transaction/helpers/index.ts
```diff
@@ -0,0 +1,47 @@
...
+  return 'reason' in result;
+}
+
+export function getTxIdFromRelayerResult(
+  result: RelayerResultFragment | RelayErrorFragment,
+): string {
+  if (isRelayerResult(result)) {
+    return result.txId;
+  }
+  throw new Error(result.reason);
```
cesarenaldi, line 37: Not sure if this error throw by these helpers is intelligible. What is the intent? The error is also not custom typed, what will happen to it? Will it be wrapped or something?

packages/client/src/authentication/__helpers__/setupAuthentication.ts
```diff
@@ -1,41 +1,203 @@
...
+  const wallet = Wallet.createRandom();
+
+  const authentication = new Authentication(testConfig);
+  const profile = new Profile(testConfig, authentication);
+  const transaction = new Transaction(testConfig, authentication);
+
+  let _walletAddress: string;
+  let _testProfileId: string;
 
   beforeAll(async () => {
```
cesarenaldi, line 46: I found this approach used in some other internal repo testing (e2e specifically). I am not sure one expects `setupAuthentication` to have side effects at the Jest runtime level, but I get what is happening here and why.

Maybe together with this suggestion https://github.com/lens-protocol/lens-sdk/pull/173/files#r1129279335 we can think of something like this?

```ts
describe('Given XYZ', () => {


  describeAuthenticatedScenario(({ }: TestSetup) => {}
  
  })
})
```

where `describeAuthenticatedScenario` also includes a `describe('and the client is authenticated', () => {...})` wrapper so that all relevant test specs includes:

```
Given XYZ
and the client is authenticated
...
```
The `describeAuthenticatedScenario` named after Jest `describe` has the intent of reminding about Jest runtime so maybe the side effect does not appear as a surprise. Hopefully.

cesarenaldi: (`and the client is authenticated` suggested description could well be improvided)

packages/client/src/transaction/helpers/index.ts
krzysu: exactly, I don't know, what should happen if the result is an instance of RelayErrorFragment? 

I have a need for a helper (ie. `extractTxIdFromRelayerResult`) to extract the values from return types. The full usage is

```ts
const profileCreateResult = await client.profile.create({ handle });

if (profileCreateResult.isSuccess()) {
  const value = profileCreateResult.value; // RelayerResultFragment | RelayErrorFragment
  
  if('txId' in value) {
    // tx to create a profile submitted with txId
  } else {
    // handle RelayErrorFragment
  }
} else {
  // profileCreateResult is a failure, means CredentialsExpiredError | NotAuthenticatedError
}
```

using helper
```ts
// profileCreateResult is a Result type, profileCreateResult.value is RelayerResultFragment | RelayErrorFragment
const txId = extractTxIdFromRelayerResult(profileCreateResult.unwrap()); // can we assume it is a success as client is authenticated?
  
if (!txId) {
  throw Error(`Profile creation failed`);
}
```

how would you model it?

packages/client/src/authentication/__helpers__/setupAuthentication.ts
```diff
@@ -1,41 +1,203 @@
...
+    }
   });
 
-  return () => auth;
+  return () => ({
+    authentication,
+    profileId: _testProfileId,
+    wallet,
+    walletAddress: _walletAddress,
+  });
```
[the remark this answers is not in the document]
krzysu: it won't work like that as `setupAuthentication` is an async function (it calls multiple async func. like authenticate, fetch profiled etc.) and you can't have jest's `describe()` to run async functions. that's why beforeAll step is needed. If you return right away, then the `authentication` is still not authenticated, profiled undefined etc.

packages/client/src/proxy-action/ProxyAction.ts
krzysu: I improved that and a few other issues around polling in this PR https://github.com/lens-protocol/lens-sdk/pull/214

packages/client/src/authentication/__helpers__/setupAuthentication.ts
krzysu: that's a very good idea, I will try it!

packages/client/src/authentication/__helpers__/signAndBroadcast.ts
krzysu: We have the same type defined both in `client` and in `react`, but the `client` package doesn't depend on `blockchain-bindings` nor on `api-bindings`. So, I guess, it's better to duplicate one type def than to couple a whole package.

packages/client/src/transaction/helpers/index.ts
krzysu: I just removed that helper, not needed anymore

packages/client/src/authentication/__helpers__/setupAuthentication.ts
krzysu: this is the last thing left from the code review, but I'd like to focus on other tasks for SDK v1 release for now.
from what I found out that requires extending jest functions like described here https://spin.atomicobject.com/2020/01/30/jest-add-custom-functions/

krzysu: ok, I added `describeAuthenticatedScenario`, it works like

```ts
describeAuthenticatedScenario({ withNewProfile: true })((getTestSetup) => {
  describe(`when ${Reactions.prototype.add.name} method is called`, () => {
    it(`should execute with success`, async () => {
      const { authentication, profileId } = getTestSetup();
      
      // run test
    });
  });
});
```

packages/client/.env.example
```diff
@@ -1 +1,5 @@
-CLIENT_TEST_WALLET_PRIVATE_KEY=
\ No newline at end of file
+TESTING_ENV_URL=
+TESTING_HEADER_KEY=
+TESTING_HEADER_VALUE=
+
+
```
bigint, line 5:
```suggestion
TESTING_ENV_URL=
TESTING_HEADER_KEY=
TESTING_HEADER_VALUE=
```

Removed two extra lines at last

bigint (changes requested):

packages/client/.env.example
krzysu: done!

packages/client/src/__helpers__/describeAuthenticatedScenario.ts
```diff
@@ -0,0 +1,26 @@
...
+  never('withDispatcher cannot be used without withNewProfile');
+}
+
+type GetTestSetupFn = () => TestSetup;
+
+export const describeAuthenticatedScenario =
+  (options?: SetupOptions) => (callback: (f: GetTestSetupFn) => void) => {
+    const getTestSetup = setupAuthenticatedScenario(options);
+
+    describe(buildDescribeName(options), () => callback(getTestSetup));
```
cesarenaldi, line 25: Do we need to pass a function at all? Can we give the object with all the testing mocks and the consumer can just access what they need?

packages/client/src/consts/environments.ts
```diff
@@ -1,17 +1,14 @@
 /**
  * @internal
- * @privateRemarks DO NOT export constructor, just the type and variables defined with it
  */
-class Environment {
```
krzysu, line 5: the constructor is not exported from the package, but it is used internally in `buildTestEnvironment`

packages/client/src/__helpers__/describeAuthenticatedScenario.ts
krzysu: the thing is that `describe` function of jest is sync, but `beforeAll` contains async calls (to create profile etc), so we only know that `beforeAll` is finished inside the `it` function.
I tried to call getTestSetup right inside `describeAuthenticatedScenario` and that's what it returns. Notice `accessToken`, `profileId`, even `walletAddress` are undefined - as they all are the result of async code.

```
getTestSetup {
      authentication: Authentication {
        api: AuthenticationApi { sdk: [Object] },
        storage: CredentialsStorage {
          refreshTokenStorage: [Storage],
          accessToken: undefined
        }
      },
      profileId: undefined,
      wallet: Wallet {
        _isSigner: true,
        _signingKey: [Function (anonymous)],
        address: '0xe449202461fe4E32d6fEd8D99D240eD1671eA574',
        _mnemonic: [Function (anonymous)],
        provider: null
      },
      walletAddress: undefined
    }
```

krzysu: I pushed the code that you requested here, but it is not working as expected https://github.com/lens-protocol/lens-sdk/commit/e2bf7cc8c6d8ddea72970cdeafb11a7681043715

cesarenaldi (approved):

PR #174, opened 2023-02-17
feat: simplifies MirrorFragment by removing redundant fields

cesarenaldi:
All fields mentioned in the ticket but `hidden` for which I will create a separate ticket. Might worth considering evicting items from cache on `HidePublication` rather than flagging them (to be discussed).


packages/api-bindings/src/graphql/utils/publication.ts
```diff
@@ -65,24 +65,37 @@ export function getApiReactionType(reaction: ReactionType): ReactionTypes {
   }
 }
 
-export type PublicationFragment = PostFragment | CommentFragment | MirrorFragment;
+export type AnyPublicationFragment = CommentFragment | MirrorFragment | PostFragment;
+
+export type ContentPublicationFragment = CommentFragment | PostFragment;
```
cesarenaldi, line 70: Once we will get to remove `Fragment` these types will look pretty self-explanatory in hooks responses/inputs.

packages/api-bindings/src/graphql/common.graphql
```diff
@@ -172,26 +172,11 @@ fragment PublicationStats on PublicationStats {
...
-    ...CollectModule
-  }
-  createdAt
-  hidden
-  isGated
-  reaction(request: { profileId: $observerId })
-  hasCollectedByMe(isFinalisedOnChain: true)
-
-  hasOptimisticCollectedByMe @client
-  collectPolicy @client
```
krzysu, line 194: for consistency, could you also apply the same changes on `MirrorBase` fragment in `packages/client/src/graphql/fragments.graphql`?

examples/web-wagmi/src/publications/UseReaction.tsx
```diff
@@ -69,6 +71,8 @@ function ReactionInner({ profile }: ReactionInnerProps) {
 
   if (error) return <ErrorMessage error={error} />;
 
+  invariant(!isMirrorPublication(publication), 'Publication is not a post');
```
desfero, line 74: can be also a comment, right?

packages/api-bindings/src/apollo/publicationTypePolicy.ts
```diff
@@ -66,38 +68,26 @@ const collectPolicy = <T extends Post | Mirror | Comment>(
 ) => {
   if (existing) return existing;
 
-  const profile = readField('profile') as PublicationFragment['profile'];
-  const collectModule = readField('collectModule') as PublicationFragment['__collectModule'];
-  const publicationStats = readField('stats') as PublicationFragment['stats'];
+  const profile = readField('profile') as ProfileFragment;
```
desfero, line 71: nit: we could at some point fix the typings to be fully typed (use `FieldFunctionOptions` under the hood so the `readField` returns correct types)

packages/api-bindings/src/graphql/__helpers__/fragments.ts
```diff
@@ -299,18 +299,10 @@ export function mockMirrorFragment(
 
   return {
     id: faker.datatype.uuid(),
-    stats: mockPublicationStatsFragment(),
```
desfero, line 302: looks a lot cleaner now

packages/api-bindings/src/graphql/common.graphql
```diff
@@ -267,6 +252,7 @@ fragment Comment on Comment {
       ...Post
     }
 
+    # we do not want to promote comments on mirrors, should we remove this?
```
desfero, line 255: if we remove the mirror edge on `commentOn` do we need to filter out manually all empty edges when receiving a response in case someone directly called the protocol and puts the comment on the mirror? 

```
__typename: "Comment",
commentOn: {},
...
```

packages/api-bindings/src/graphql/utils/publication.ts
desfero: yes, agree

packages/react/src/transactions/adapters/responders/__tests__/CollectPublicationResponder.spec.ts
```diff
@@ -62,125 +54,155 @@ function setupTestScenario({
...
-          hasCollectedByMe: false,
-          stats: mockPublicationStatsFragment({ totalAmountOfCollects: 1 }),
+            },
+          },
         });
+      });
+    });
+
+    describe(`when "${CollectPublicationResponder.prototype.commit.name}" method is invoked`, () => {
+      it(`should confirm the confirm the changes done during the "${CollectPublicationResponder.prototype.prepare.name}" method call`, async () => {
```
desfero, line 123:
```suggestion
      it(`should confirm the changes done during the "${CollectPublicationResponder.prototype.prepare.name}" method call`, async () => {
```

packages/react/src/transactions/adapters/responders/__tests__/CollectPublicationResponder.spec.ts
```diff
@@ -62,125 +54,155 @@ function setupTestScenario({
...
+        expect(scenario.updatedPublicationFragment).toMatchObject({
+          hasCollectedByMe: true,
+          hasOptimisticCollectedByMe: false,
+          stats: {
+            totalAmountOfCollects: 2,
+          },
         });
+      });
+
+      it('should confirm the update the publication also if collected via a mirror', async () => {
```
desfero, line 142:
```suggestion
      it('should confirm the changes also if collected via a mirror', async () => {
```

packages/react/src/transactions/adapters/responders/__tests__/CollectPublicationResponder.spec.ts
```diff
@@ -62,125 +54,155 @@ function setupTestScenario({
...
+        await scenario.responder.rollback(transactionData);
+
+        expect(scenario.updatedPublicationFragment).toMatchObject({
+          hasCollectedByMe: false,
+          hasOptimisticCollectedByMe: false,
+          stats: knownInitialPublicationStats,
         });
+      });
+
+      it('should confirm the update the publication also if collected via a mirror', async () => {
```
desfero, line 185:
```suggestion
      it('should revert the changes also if collected via a mirror', async () => {
```

desfero (approved):
small nitpicks, approving to push it forward

packages/api-bindings/src/apollo/publicationTypePolicy.ts
cesarenaldi: I gave it a couple of tries in the last weeks. It's kinda tricky, simply because `TypePolicy` and `FieldPolicy` looks at the original GQL node and edges definitions. While the data retrieved from the cache are fragments, which in the best case scenarios are a subset of the nodes, but they could also be a superset of the node definitions if they include aliased fields.

examples/web-wagmi/src/publications/UseReaction.tsx
cesarenaldi: I blame co-pilot.

packages/api-bindings/src/graphql/common.graphql
cesarenaldi: I thought about that... I am just not sure when and where to do it reliably. It could be done as `merge` function of all fields that could return publications, but seems daunting and error prone (if we don't add such filtering to a new publications query for example).

packages/api-bindings/src/graphql/common.graphql
cesarenaldi: For posterity: agreed to look at this holistically as part of the low-level client effort.

PR #177, opened 2023-02-20
chore: enable dependabot T-1334

krzysu:
this is a simple dependabot config, if it won't work for monorepo, then we have to configure each directory separately, as in the example

but first we have to merge https://github.com/lens-protocol/lens-sdk/pull/171


cesarenaldi:
cool, looking forward to get this updating deps for us.

desfero (approved):
cool, let's see if it won't annoy us in the following weeks :D

krzysu:
that's why I set `open-pull-requests-limit: 2` :)

PR #181, opened 2023-02-21
feat: client sdk - remaining profile queries and mutations

krzysu:


packages/client/src/profile/Profile.ts
```diff
@@ -163,6 +190,10 @@ export class Profile {
     });
   }
 
+  isValidHandle(handle: string): boolean {
+    return isValidHandle(handle);
+  }
```
krzysu, line 195: I'm not sure yet how we should provide helpers from the client SDK, as part of a module or as stand-alone functions or maybe as `client.utils.*` module as ethers@5

cesarenaldi: If they don't need any state from the client instance if would suggest to export at the module level. Thoughts?

krzysu: there is no state kept in the client instance outside of the credentials, helpers are pure functions

so it would look like
```
import { isValidHandle } from "@lens-protocol/client/profile"
```
or 
```
import { isValidProfileHandle } from "@lens-protocol/client"
```

desfero: what if we keep them here for consistency? 
Later when we add let's say `isHandleUsed` which would be asynchronous it would be harder to understand which one i use directly as a function and which one I access from the `Profile` class

desfero (approved):
LGTM 🥇

packages/client/src/profile/Profile.ts
cesarenaldi: I don't know, I think if we go down the habit of using namespace objects we are kinda working against the tools provided by the language (in this case the module system).

The whole client lib is a massive namespace object but we intend to provide a way to cherry pick only the parts one wants/need so that more expert integrators can tweak it to their app needs.

krzysu: I have another branch with more helpers, I created a ticket to figure out the consistent way of exposing them, I will handle them all in a follow up PR, for now I will merge as it is.

https://avara.height.app/T-7156

PR #185, opened 2023-02-22
chore: configure dependabot for monorepo

krzysu:
I assume the current setup for dependabot is only on root package.json and all packages are up-to-date there.
so let's add all our packages to the config too.
I excluded `gated-sdk` and all `examples` for now.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,7 +10,62 @@ updates:
     open-pull-requests-limit: 2
 
   # Configuration for monorepo packages, one per directory
-  # - package-ecosystem: "npm"
-  #   directory: "/packages/client"
-  #   schedule:
-  #     interval: "weekly"
+  - package-ecosystem: "npm"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -1,16 +1,44 @@
 # Please see the documentation for all configuration options:
 # https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
 
+__: &npm_config
+  package-ecosystem: "npm"
+  schedule:
+    interval: "weekly"
+  open-pull-requests-limit: 2
```
krzysu, line 8: I wonder if github will complain about this config, otherwise I can't reuse the block

krzysu: and I have to repeat myself as in https://github.com/lens-protocol/lens-sdk/pull/185/commits/fc01a6d239b37d11000eac978c897e8e760f1aeb

PR #187, opened 2023-02-23
feat: makes space for `EncryptedPublicationMetadataUploader`

cesarenaldi:


packages/api-bindings/src/metadata.ts
```diff
@@ -0,0 +1,23 @@
...
+  traitType?: string;
+  value: string;
+};
+
+export type ProfileMetadata = {
+  version: '1.0.0';
+  metadata_id: string;
+  attributes: ProfileMetadataAttribute[];
+  bio: string | null;
+  cover_picture: string | null;
```
desfero, line 21: nit: unusual to see pascal case here over the camelcase.

packages/react/src/transactions/infrastructure/PublicationMetadataUploader.ts
```diff
@@ -0,0 +1,26 @@
...
+import { createPublicationMetadata } from './createPublicationMetadata';
+
+export class PublicationMetadataUploader<T extends CreatePostRequest | CreateCommentRequest>
+  implements IMetadataUploader<T>
+{
+  constructor(private readonly handler: MetadataUploadHandler) {}
+
+  async upload(request: T): Promise<Url> {
+    try {
+      const metadata = createPublicationMetadata(request);
```
desfero, line 18: nit: just API disproportion which can be good,  `PublicationMetadataUploader` takes the `request` whereas the `ProfileMetadataUploader` takes the metadata directly.

packages/react/src/transactions/useCreatePost.ts
```diff
@@ -33,7 +36,8 @@ export type CreatePostOperation = Operation<
 >;
 
 export function useCreatePost({ publisher, upload }: UseCreatePostArgs): CreatePostOperation {
-  const createPost = useCreatePostController({ upload });
+  const uploader = useMemo(() => new PublicationMetadataUploader(upload), [upload]);
```
desfero, line 39: cool, no changes to the consumer-facing API

desfero (approved):

packages/api-bindings/src/metadata.ts
cesarenaldi: This is the metadata JSON file format, we have to use what the metadata specification tells u.

packages/react/src/transactions/infrastructure/PublicationMetadataUploader.ts
cesarenaldi: I know, that's why `T` in `IMetadataUploader<T>` is without restrictions.

PR #191, opened 2023-02-27
chore: handle search result empty types after graphql-codegen upgrade

krzysu:
context discussion https://github.com/dotansimha/graphql-code-generator/discussions/5567

### DIFF
--- a/packages/client/src/helpers/buildPaginatedQueryResult.spec.ts
+++ b/packages/client/src/helpers/buildPaginatedQueryResult.spec.ts
@@ -1,7 +1,7 @@
 import { GraphQLClient } from 'graphql-request';
 
 import { mumbaiSandbox } from '../consts/environments';
-import { getSdk } from '../search/graphql/search.generated';
+import { getSdk } from '../profile/graphql/profile.generated';
 import { buildPaginatedQueryResult } from './buildPaginatedQueryResult';
 
 describe('Given a paginated query function and the paginated query result helper', () => {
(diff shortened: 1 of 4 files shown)


packages/client/src/search/Search.ts
```diff
@@ -21,19 +21,26 @@ export class Search {
...
 
   async publications(
     request: SearchPublicationsQueryVariables,
   ): Promise<PaginatedResult<CommentFragment | PostFragment>> {
     return buildPaginatedQueryResult(async (variables) => {
-      const result = await this.sdk.SearchPublications(variables);
+      const response = await this.sdk.SearchPublications(variables);
+      const result = response.data.result;
+
+      if (result.__typename === 'ProfileSearchResult') throw new Error('This should never happen');
```
krzysu, line 41: what do you think about a solution like this?

packages/client/src/search/Search.ts
```diff
@@ -21,19 +22,29 @@ export class Search {
...
   async publications(
     request: SearchPublicationsQueryVariables,
   ): Promise<PaginatedResult<CommentFragment | PostFragment>> {
     return buildPaginatedQueryResult(async (variables) => {
-      const result = await this.sdk.SearchPublications(variables);
+      const response = await this.sdk.SearchPublications(variables);
+      const result = response.data.result;
+
+      if (result.__typename === 'ProfileSearchResult') {
+        never('ProfileSearchResult are not expected in this query');
```
krzysu, line 44: what do you think about a solution like this?

packages/client/src/helpers/buildPaginatedQueryResult.spec.ts
```diff
@@ -1,33 +1,37 @@
...
 
 describe('Given a paginated query function and the paginated query result helper', () => {
   const sdk = getSdk(new GraphQLClient(mumbaiSandbox.gqlEndpoint));
 
   it('should be able to paginate forward', async () => {
     const result = await buildPaginatedQueryResult(
-      async (variables) => {
-        const res = await sdk.SearchProfiles(variables);
+      async (currRequest) => {
+        const res = await sdk.Profiles({
```
krzysu, line 13: I changed to less "troublesome" query in this test

packages/client/src/search/Search.ts
desfero: looks good, we can also go with `invariant(result.__typename === "PublicationSearchResult", ...)` to have just one line but that's nit

desfero (approved):

PR #192, opened 2023-02-27
feat: creates `@lens-protocol/gated-content` internal package using new Lit TS SDK

cesarenaldi:
At the time of this writing this PR contains a new `@lens-protocol/gated-content` internal package which will be used to provide token-gated support via Lit Protocol to the `@lens-protocol/react` and `@lens-protocol/client` public SDK packages.

Once approved I will use this branch as integration branch for the list of things below:
- [x] storage wiring -  in this PR
- [x] `useCreateEncryptedPost` - https://github.com/lens-protocol/lens-sdk/pull/203
- [x] `useCreateEncryptedPost` example - https://github.com/lens-protocol/lens-sdk/pull/204
- [x] `useEncryptedPublication` w/ example - https://github.com/lens-protocol/lens-sdk/pull/208
- [x] Compute Publication ID for Collect This Publication criterion - https://github.com/lens-protocol/lens-sdk/pull/209
- [x] Cover failed upload via `EncryptedPublicationMetadataUploader` - https://github.com/lens-protocol/lens-sdk/pull/211
- [x] Integrate new fields from [T-6567](https://avara.height.app/T-6567) - https://github.com/lens-protocol/lens-sdk/pull/222
- [x] `useCreateEncryptedComment` - https://github.com/lens-protocol/lens-sdk/pull/212
- [x] ~~`useCreateEncryptedComment` example~~
- [x] ~~formalize error handling (determine domain errors vs. invariants)~~
- [x] Fix peer deps issue from Lit SDK: https://github.com/lens-protocol/lens-sdk/pull/221, https://github.com/lens-protocol/lens-sdk/pull/224
```bash
packages/gated-content
└─┬ @lit-protocol/lit-node-client 2.1.42
  └─┬ @lit-protocol/auth-browser 2.1.42
    ├─┬ @walletconnect/ethereum-provider 2.4.6
    │ └─┬ @web3modal/standalone 2.1.1
    │   └─┬ @web3modal/core 2.1.1
    │     └─┬ valtio 1.9.0
    │       ├── ✕ missing peer react@>=16.8
    │       └─┬ use-sync-external-store 1.2.0
    │         └── ✕ missing peer react@"^16.8.0 || ^17.0.0 || ^18.0.0"
    └─┬ lit-siwe 1.1.8
      ├── ✕ missing peer @ethersproject/contracts@^5.2.0
      ├── ✕ missing peer @ethersproject/hash@^5.4.0
      ├── ✕ missing peer @ethersproject/providers@^5.2.0
      └── ✕ missing peer @ethersproject/wallet@^5.2.0
Peer dependencies that should be installed:
  @ethersproject/contracts@^5.2.0
  @ethersproject/hash@^5.4.0
  @ethersproject/providers@^5.2.0
  @ethersproject/wallet@^5.2.0
  react@">=16.8.0 <17.0.0 || >=17.0.0 <18.0.0 || >=18.0.0 <19.0.0"
```


packages/gated-content/src/__tests__/GatedClient.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+        ownerId,
+        others: [eoaOwnershipAccessCondition],
+        encryptionKey: encrypted.encryptionParams.providerSpecificParams.encryptionKey,
+        encryptedFields: mockEncryptedFieldsFragment(
+          encrypted.encryptionParams.encryptedFields as EncryptedFieldsFragment,
+        ),
+      });
+
+      // exercise
+      const decryptionResult = await client.decryptPublication(metadataFragment, encryptionParams);
```
cesarenaldi, line 104: For now it seems there is not really a use case to support decryption of raw Publication Metadata (i.e. the `GatePublicationMetadata` returned by the `encryptPublication` method).

From the `@lens-protocol/react` SDK perspective the usage of `GatedClient` is completely segregated and makes use of the same data structure used by `CommentFragment` and `PostFragment` (i.e. any fragment shape of `MetadataOutput`). This has  the nice advantage of not requiring a bespoke render path (nor transformation within the SDK) for rendering decrypted gated content. It also does not require to make an extra request to download raw metadata from `contentURI` (see how Lenster implemented token-gated to get an idea of these pain points).

`GatedClient` will also receive a small adapter to work within the low-level client SDK (i.e. `@lens-protocol/client`). This will be type safe in the sense that the integrator will be guided by the TS typedef to feed the entire `CommentFragment` and `PostFragment` into the adapter interface. The adapter will then use this `GatedClient#decryptPublication` as needed.

If the need arise to be able to decrypt raw metadata (as per JSON file uploaded on `contentURI`), I strongly suggest to create a bespoke method that knows the details of that data structure. As comments left in the code, there are some discrepancies between raw data (name, camelCases vs. snake_cased properties, missing properties). All these are implementation details, a consumer using the Lens SDK does not even need to know about these, so we can improve the Lens API over time and fix those discrepancies (if need to).

packages/gated-content/src/conditions/index.ts
```diff
@@ -0,0 +1,137 @@
...
+    rootEntries.map((entry) =>
+      transformCompoundCondition(entry as Entry<AnyCriterionFragment>, env),
+    ),
+  );
+
+  // the type assertion is needed because the Lit SDK typedef suggests nested conditions are not allowed but they are
+  return insertObjectInBetweenArrayElements(flat, {
+    operator: AccessConditionType.Or,
+  }) as UnifiedAccessControlConditions;
+}
```
cesarenaldi, line 137: 100% of the logic in this folder has been adapted from the `@lens-protocol/sdk-gated`. The validation logic has been merged into the traversing done during the `transform` logic, this reduced the times we need to 

Collect Access Condition when `thisPublication` is set assumes the `publicationId` has been provided by the caller. The Publication ID prediction logic needs to be done by the caller (`@lens-protocol/react` and `@lens-protocol/client`) cause they have significant differences on the way the next publication ID can be inferred.
Namely `@lens-protocol/react` requires to take into account what is in the current `TransactionQueue` while the `@lens-protocol/client` can take the traditional approach used in the `@lens-protocol/sdk-gated` package (just fetch last publication Id for the given profile and "add 1" to it).

packages/gated-content/package.json
```diff
@@ -0,0 +1,87 @@
...
+    "lint:fix": "pnpm run prettier:fix && pnpm run eslint:fix && pnpm run tsc",
+    "prettier:fix": "prettier --write .",
+    "prettier": "prettier --check .",
+    "test": "jest",
+    "test:watch": "jest --watch",
+    "tsc": "tsc --noEmit"
+  },
+  "license": "MIT",
+  "dependencies": {
+    "@ethersproject/address": "^5.7.0",
```
desfero, line 37: curious, given that we also install full `ethers`, does that mean it's possible to end up with duplicated code when both `@ethersproject/address` and `ethers` address have the same code?

packages/gated-content/src/conditions/index.ts
```diff
@@ -0,0 +1,137 @@
...
+import { insertObjectInBetweenArrayElements } from './utils';
+import {
+  assertAtLeastTwoCriteria,
+  assertNoMoreThanFiveCriteria,
+  InvalidAccessCriteriaError,
+} from './validators';
+
+function extractConditionEntry<T extends UnknownObject>(condition: T): Entry<T> {
+  for (const key in condition) {
+    if (isAccessConditionType(key) && isNonNullable(condition[key as keyof T])) {
```
desfero, line 32: if we extract the condition key and save it into the variable `const value = condtion[key]` will that allow us to remove `never` on line 33?

packages/gated-content/src/conditions/index.ts
```diff
@@ -0,0 +1,137 @@
...
+      return [key, condition[key] ?? never()];
+    }
+  }
+  throw new InvariantError('Invalid access condition');
+}
+
+function flatten<T>(conditions: T[]): T[] {
```
desfero, line 39: can we just use array `flatten`?

desfero (approved):

packages/gated-content/src/conditions/index.ts
cesarenaldi: This one flattens it only if the nested array has `lenght==1`, not exactly the same as lodash-style 'flatten` function.

PR #198, opened 2023-03-01
chore: prepare release v0.5.0 and client v0.1.0

krzysu:


desfero (approved):
LGTM, do we need to update changeset config to disconnect `@lens-protocol/client`  versioning from the react packages (https://github.com/changesets/changesets/blob/main/docs/config-file-options.md)?

krzysu:
> LGTM, do we need to update changeset config to disconnect `@lens-protocol/client` versioning from the react packages (https://github.com/changesets/changesets/blob/main/docs/config-file-options.md)?

yes, good point, as all our packages follow the same naming convention then we will have to write the names explicitly now, instead of the `""@lens-protocol/*""` shortcut, yes?

PR #200, opened 2023-03-02
chore: fix example apps

krzysu:
I updated the login flow implementation in nextjs app to use the same components as in web-wagmi, refactored also a bit the components structure in web-wagmi


examples/nextjs/pages/index.tsx
```diff
@@ -1,27 +1,8 @@
...
-
-  const { isDisconnected } = useAccount();
-  const { connectAsync } = useConnect({
-    connector: new InjectedConnector(),
-  });
-  const onLoginClick = async () => {
-    if (isDisconnected) {
-      const { connector } = await connectAsync();
-
-      if (connector instanceof InjectedConnector) {
```
krzysu, line 18: I updated the login flow implementation in nextjs app to use the same components as in web-wagmi, refactored also a bit component structure in web-wagmi

examples/nextjs/components/auth/LoginButton.tsx
```diff
@@ -0,0 +1,60 @@
...
+      </WhenLoggedInWithProfile>
+
+      <WhenLoggedOut>
+        <button onClick={onLoginClick} disabled={isLoginPending}>
+          <strong>Log in</strong>
+        </button>
+      </WhenLoggedOut>
+    </>
+  );
+}
```
cesarenaldi, line 60: We have `examples/shared` if need to.

krzysu: I know we repeat `LoginButton`, `WhenLoggedInWithProfile` and `WhenLoggedOut` components, they depend on `wagmi`, `react` and `@lens-protocol/react`.
I'm not sure about expanding `examples/shared` for now, from the other side that could become a package of useful Lens components in the future, if we keep adding to it

cesarenaldi (approved):

PR #202, opened 2023-03-03
chore: add more examples to the node app T-7571

krzysu:


packages/client/src/LensClient.ts
```diff
@@ -33,7 +33,7 @@ export class LensClient {
   }
 
   get feed(): Feed {
-    return new Feed(this.config);
+    return new Feed(this.config, this._authentication);
```
krzysu, line 36: I found a very bad bug here, Feed won't work without it

packages/client/src/nonces/Nonces.ts
```diff
@@ -12,7 +12,7 @@ export class Nonces {
   private readonly authentication: Authentication | undefined;
   private readonly sdk: Sdk;
 
-  constructor(config: LensConfig, authentication?: Authentication) {
+  constructor(config: LensConfig, authentication: Authentication) {
```
krzysu, line 15: for some submodules authentication is not optional (!)

cesarenaldi (approved):

PR #208, opened 2023-03-07
feat: adds `useEncryptedPublication` hook

cesarenaldi:


packages/react/src/__helpers__/jest.setup.ts
```diff
@@ -0,0 +1,18 @@
...
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+  value: Object.setPrototypeOf({ subtle: crypto.subtle }, crypto),
+});
+
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
+global.Blob = Blob;
+
+// until https://github.com/jsdom/jsdom/issues/2524 is resolved
+global.TextEncoder = TextEncoder;
```
krzysu, line 15: where are these used?

package.json
```diff
@@ -42,7 +42,8 @@
       ],
       "allowedVersions": {
         "@solana/web3.js": "1.52.0",
-        "ethers": "5.7.2"
+        "ethers": "5.7.2",
+        "jest": "29.4.3"
```
krzysu, line 46: why is that?

krzysu (approved):

packages/react/src/__helpers__/jest.setup.ts
cesarenaldi: Inside https://github.com/lens-protocol/lens-sdk/pull/192/files#diff-fa678de0ca5e45b91b0e063a33b6fd1db971e0a4ac7d43059ac3adfb9b41513e from a previous PR.

As in this PR i am testing a component thar uses `GatedClient`, it now needs the same polyfills.

package.json
cesarenaldi: To silence a warning of mismatch of peer dependency from a 3rd party lib. Which is nosense cause we don't even run the tests of the lib, but 🤷

PR #211, opened 2023-03-08
feat: handles failed upload error from user's provided upload handler

cesarenaldi:


krzysu (approved):

krzysu:
I will test all of it from a user perspective when the final branch (PR) is ready

PR #215, opened 2023-03-09
Add notification type filter to useNotifications hook

paulxuca:
Looks like we need this functionality elsewhere!


cesarenaldi (approved):
👌

PR #221, opened 2023-03-10
feat: fixes dependency tree so to improve tree-shakeablity of the React SDK

cesarenaldi:
- Creates `@lens-protocol/react-web` that re-exports all `@lens-protocol/react` hooks and specialize configuration that is relevant on web apps so to minimize the amount of configuration one needs to put in place to get going with the Lens SDK.
- Amends web app examples to use new `@lens-protocol/react-web` package
- introduces a new internal `@lens-protocol/tsconfig` package to start consolidating TSConfig across packages. In this PR only used by `@lens-protocol/react-web` but will be extended to all packages in a follow-up PR.
- Fixes discrepancy in Lit Protocol packages version which was causing duplicated packages in the final bundles.
- Fixes issue with `webCryptoProvider` being instantiated at module level which in turn on Rollup builds (like ViteJS does) was causing tree shaking to not properly working. Although this was not a direct issue with the lens SDK the code changes reduces chances for the consumer app to have larger than needed bundles out of the box.


.vscode/settings.json
```diff
@@ -10,11 +10,12 @@
     "packages/domain",
     "packages/eslint-config",
     "packages/prettier-config",
+    "packages/react-web",
     "packages/react",
     "packages/sdk-gated",
     "packages/shared-kernel",
     "packages/storage",
-    "packages/wagmi",
-    "packages/gated-content"
```
krzysu, line 18: is the removal of `"packages/gated-content"` intended here?

krzysu (approved):

.vscode/settings.json
cesarenaldi: nope...realized that and fixed in coming PR (latest of the series hopefully)

PR #224, opened 2023-03-13
chore: polishes `@lens-protocol/gated-content` dependencies

cesarenaldi:


packages/gated-content/.gitignore
```diff
@@ -0,0 +1,26 @@
+# Logs
+logs
+*.log
+npm-debug.log*
+yarn-debug.log*
+yarn-error.log*
```
desfero, line 6: curious, was this generated by some tool given has a lot of files that we can't have?

packages/gated-content/package.json
```diff
@@ -71,10 +71,16 @@
     "prettier": "^2.8.4",
     "ts-jest": "^29.0.5",
     "typescript": "^4.9.5",
-    "wait-for-expect": "^3.0.2"
+    "wait-for-expect": "^3.0.2",
+    "zod": "^3.20.6"
   },
   "peerDependencies": {
-    "ethers": "^5.7.2"
+    "@ethersproject/contracts": "^5.7.0",
```
desfero, line 78: just for me to get it right, what's the reason we keep part of ethers in `peerDependencies` and part in `dependencies`?

packages/gated-content/tsconfig.base.json
```diff
@@ -1,24 +1,10 @@
 {
+  "extends": "@lens-protocol/tsconfig/base.json",
```
desfero, line 2: 🔥

packages/gated-content/tsconfig.base.json
```diff
@@ -1,24 +1,10 @@
...
-    "resolveJsonModule": true,
-    "skipLibCheck": true,
-    "sourceMap": true,
-    "strict": true,
-    "target": "esnext"
+    "outDir": "dist"
   },
-  "include": ["./src"]
+  "include": ["./src"],
+  "exclude": ["./node_modules"]
```
desfero, line 9: do you need to exclude `./node_modules` if we are already not including them on `include` side?

desfero (approved):

packages/gated-content/package.json
cesarenaldi: The ones in `dependencies` are used directly by us. The ones in `peerDependencies` are used by sub-dependencies. The version number should help install the correct version that goes with the rest of the repo dependencies on the `ethersproject` package ecosystem.

packages/gated-content/tsconfig.base.json
cesarenaldi: let me clean this, just comes from an iteration while I was experimenting custom build with rollup.

PR #225, opened 2023-03-13
feat: client SDK - wrap all API requests with auth header if available

krzysu:
while working on `scripts/followProfile.ts` in the example-node app, I noticed that the recommended profiles result is different if the auth header is provided, despite that the endpoint doesn't require auth. 
I wrapped all API requests in a `provideAuthHeaders` helper, which optionally adds auth header if the access token is available.
I also renamed `execute` to `requireAuthHeaders` to make it obvious what is the difference.


package.json
```diff
@@ -11,7 +11,6 @@
     "build": "pnpm clean && turbo build",
     "postinstall": "preconstruct dev",
     "clean": "rimraf packages/*/*/dist packages/*/dist",
-    "build:packages": "turbo run build:packages",
```
krzysu, line 14: this is not working correctly since introduction of `preconstruct` I guess

packages/client/src/helpers/provideAuthHeaders.ts
```diff
@@ -0,0 +1,27 @@
...
+  return headerResult.value;
+}
+
+export async function provideAuthHeaders<Val>(
+  authentication: Authentication | undefined,
+  handler: Handler<Val>,
+): Promise<Val> {
+  const header = await getHeader(authentication);
+
+  return handler(header);
```
krzysu, line 26: I could export only `getHeader` function (with a better name) but I wanted to keep the usage the same as in `requireAuthHeader` for consistency, but we can discuss that

cesarenaldi (approved):

PR #226, opened 2023-03-13
feat: auto-wires Active Profile as Observer Id

cesarenaldi:
### In this PR:
- Creates `useActiveProfileIdentifier` out of `useActiveProfile` implementation. This will allow to sync all hooks as soon as we know if there is an Active Profile ID (no need to wait for the whole profile data `useActiveProfile` fetches afterwards).
- Defines new helper hooks that can be composed to achieve the desired behaviour:
  - `useLensApolloClient` injects `LensApolloClient` from shared deps as `client` prop of `QueryHookOptions`
  - `useActiveProfileAsDefaultObserver` halt current query using the `QueryHookOptions.skip` prop and injects ActiveProfile ID as `observerId`. It also allows to override `observerId`.
  - `useConfigSourcesVariable` (I migtht rename this one) - injects `sources` variable from shared deps.
- Amends the following hook so that they use Active Profile ID as default `observerId`:
  - `useCollectedPublications`
  - `useExploreProfiles`
  - `useProfile`
  - `useProfileFollowers`
  - `useProfileFollowing`
  - `useProfilesOwnedBy`
  - `useProfilesOwnedByMe` - also fixes it, didn't manage logged out state and would have make requests with `''` (empty string) as owner address.
  - `useProfilesToFollow` - also fixes it, didn't have `sources`
  - `useSearchProfiles`

### Also:
- Splits hook helpers into several modules so to avoid circular imports
- Refactors `useEnabledModules`, `useNotification`, `useUnreadNotificationCount`, `useActiveProfile`, `useMutualFollower`  to use new helpers

### What's missing?

All publications hooks


packages/react/src/notifications/useNotifications.ts
```diff
@@ -30,17 +31,15 @@ export function useNotifications({
...
-        limit: limit ?? 10,
-        notificationTypes,
-        sources,
-      },
-      client: apolloClient,
-    }),
+    useNotificationsQuery(
+      useLensApolloClient({
+        variables: useConfigSourcesVariable({
+          observerId: profileId,
```
krzysu, line 38: why not use `useActiveProfileAsDefaultObserver` here?

krzysu (approved):

packages/react/src/notifications/useNotifications.ts
cesarenaldi: This is mostly in relation to the way this query was conceived. The assumption was that notifications are very personal. It's very unlikely that one asks someone else notifications from current active provide perspective. Probably even more so for asking your own notifications with someone else perspective.

So the query has already the request `profileId` also as the observer. Like saying, if you ask notification for profile A only profile A perspective matters.

PR #228, opened 2023-03-14
feat: allow to define appId in the LensConfig

krzysu:
example metadata with appId created with web-wagmi example app
https://co2e35tifh3htlkrjewg77qsfurkvdopte2lcwtztkksaolzileq.arweave.net/E7RN9mgp9nmtUUksb_4SLSKqjc-ZNLFaeZqVIDl5Qsk

### DIFF
--- a/packages/domain/src/use-cases/publications/CreatePost.ts
+++ b/packages/domain/src/use-cases/publications/CreatePost.ts
@@ -1,6 +1,6 @@
 import { invariant } from '@lens-protocol/shared-kernel';
 
-import { DecryptionCriteria, TransactionKind } from '../../entities';
+import { AppId, DecryptionCriteria, TransactionKind } from '../../entities';
 import {
   DelegableProtocolCallUseCase,
   IDelegableProtocolCallGateway,
@@ -11,6 +11,7 @@ import { ReferencePolicyConfig } from './ReferencePolicyConfig';
 import { CollectPolicyConfig, ContentFocus, Locale, Media } from './types';
 
 export type CreatePostRequest = {
+  appId?: AppId;
   content?: string;
   contentFocus: ContentFocus;
   media?: Media[];
--- a/packages/react/src/config.ts
+++ b/packages/react/src/config.ts
@@ -1,6 +1,7 @@
 import { Sources } from '@lens-protocol/api-bindings';
+import { AppId } from '@lens-protocol/domain/entities';
 import { AuthenticationConfig, IEncryptionProvider } from '@lens-protocol/gated-content';
-import { ILogger } from '@lens-protocol/shared-kernel';
+import { ILogger, invariant } from '@lens-protocol/shared-kernel';
 import { IStorageProvider, IObservableStorageProvider } from '@lens-protocol/storage';
 
 import { EnvironmentConfig } from './environments';
@@ -20,9 +21,17 @@ export type LensConfig = {
   logger?: ILogger;
   storage: IStorageProvider | IObservableStorageProvider;
   sources?: Sources;
+  appId?: AppId;
 };
 
 export type EncryptionConfig = {
   authentication: AuthenticationConfig;
   provider: IEncryptionProvider;
 };
+
+export function validateConfig(config: LensConfig) {
+  invariant(
+    !(config.appId && config.sources && !config.sources?.includes(config.appId)),
+    `LensProvider config: "sources" don't include your "appId"`,
+  );
+}
--- a/packages/react/src/transactions/infrastructure/createPublicationMetadata.ts
+++ b/packages/react/src/transactions/infrastructure/createPublicationMetadata.ts
@@ -36,7 +36,7 @@ export function createPublicationMetadata(
   const sharedMetadata = {
     version: '2.0.0',
     metadata_id: v4(),
-
+    ...(request.appId && { appId: request.appId }),
     content: request.content,
     media: request.media?.map(mapMedia),
     locale: request.locale,
--- a/.changeset/twenty-icons-buy.md
+++ b/.changeset/twenty-icons-buy.md
@@ -1,0 +1,6 @@
+---
+"@lens-protocol/domain": minor
+"@lens-protocol/react": minor
+---
+
+Added appId to LensConfig. It is used when creating publications.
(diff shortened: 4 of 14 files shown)


packages/domain/src/use-cases/publications/CreatePost.ts
```diff
@@ -11,6 +11,7 @@ import { ReferencePolicyConfig } from './ReferencePolicyConfig';
 import { CollectPolicyConfig, ContentFocus, Locale, Media } from './types';
 
 export type CreatePostRequest = {
+  appId?: AppId;
```
cesarenaldi, line 14: I didn't think of passing this as part of the request models, I though of this as an implementation detail. But ultimately `AppId` is an domain concept and the more I look at it the more it's growing on me. Good call!

To fully support this you want to also make sure that when `SupportedTransactionRequest` (of which both `CreatePostRequest` and `CreateCommentRequest` belongs to), you need to make sure it can be serialized and deserialized from storage when part of a "pending transaction".

It boils down to adding zod validation here: https://github.com/lens-protocol/lens-sdk/blob/main/packages/react/src/transactions/adapters/PendingTransactionGateway/schema/publications.ts#L93 and https://github.com/lens-protocol/lens-sdk/blob/main/packages/react/src/transactions/adapters/PendingTransactionGateway/schema/publications.ts#L105

packages/react/src/config.ts
```diff
@@ -19,4 +20,5 @@ export type LensConfig = {
   logger?: ILogger;
   storage: IStorageProvider | IObservableStorageProvider;
   sources?: Sources;
+  appId?: AppId;
```
cesarenaldi, line 23: After this change, should we ensure that if one provides `appId` there is also `sources` where one of the item is the same `appId`?  It could lead to an odd UX if `sources` is configured for example with `X` and `Y` but `appId` publishes on, let's say `Z`.

packages/react/src/transactions/infrastructure/createPublicationMetadata.ts
```diff
@@ -36,7 +36,7 @@ export function createPublicationMetadata(
   const sharedMetadata = {
     version: '2.0.0',
     metadata_id: v4(),
-
+    ...(request.appId && { appId: request.appId }),
```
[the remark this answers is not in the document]
krzysu: test added

packages/domain/src/use-cases/publications/CreatePost.ts
krzysu: it doesn't work for me, how to add AppId type with zod?

This is complaining that a string is not the same type as AppId

```
appId: z.string().optional(),
```

packages/react/src/config.ts
cesarenaldi: I personally think it should be an invariant check, I cannot think of a normal scenario where one can get this configuration wrong... @desfero what do you think?

Also bear in mind that providing `appId` and not provide `sources` is a legitimate configuration. It means one does want to flag publication as originated from `appId` but it wants to show data from the entire ecosystem (🤔 ).

I would fail as soon as possible. A while back I was entertaining the idea if we need some sort of config helper like wagmi does.

The helper could accept raw JS types and return properly branded `sources` or `appId` for example. Over time it might be a place to do validation (actually the check we are talking about seems validation indeed). Thoughts?

packages/domain/src/use-cases/publications/CreatePost.ts
cesarenaldi: This is probably cause `AppId` is branded. It might require a TS assertion or something.

cesarenaldi: This long article explore zod with a library that makes use of branded types: https://www.sandromaglione.com/techblog/zod-and-newtype-ts-full-type-safety-with-typescript

**TL;DR;**
In our case `.transform(appId)` should be enough (not sure if before or after `optional`, you will need to try).

packages/react/src/config.ts
krzysu: I added `validateConfig` helper that is called in `LensProvider`, this way I can test it independently.
Regarding config helper, I think it would be an overkill for now, I understand you have sth like the below on your mind.

```
const { sources, appId } = buildConfig('my-app-id')

const lensConfig = {
  bindings: wagmiBindings(),
  environment: staging,
  storage: localStorage(),
  sources, 
  appId
}

```

packages/domain/src/use-cases/publications/CreatePost.ts
krzysu: I had to replace our implementation of Branded with zod one, to make them compatible, I hope that's fine

```
const AppId = z.string().brand<'AppId'>();
export type AppId = z.infer<typeof AppId>;
```

cesarenaldi: Unfortunately this dependency https://github.com/lens-protocol/lens-sdk/pull/228/files#diff-20cd445a0f23c7bcac95f9d29cdc1bfdb772a0642e6d190367e0d2b0e288dfeaR64

is problematic in terms for dependency rule. Now Domain depends on a 3rd party library.

krzysu: yeah and the build is broken, I can't figure out how to solve it so I guess I will revert adding `zod` to the `domain`

some explanation: https://github.com/microsoft/TypeScript/issues/29808#issuecomment-1320713018

packages/react/src/config.ts
cesarenaldi: I was thinking more akin to wagmi `createClient`, so in our case:
```ts

import { createConfig } from '@lens-protocol/react`;

const config = createConfig({
  source: ['foo', 'bar'], // from a type perspective this is Array<string>
  appId: 'foo' // string
  ...
})
```
The generated `config`  can then be opaque to the consumer but for the purpose of this conversation we can assume:
```ts
config.appId // AppId
condif.sources // Array<AppId> OR Sources which is the same
```



```

packages/domain/src/use-cases/publications/CreatePost.ts
krzysu: ok, solved with `appId: z.string().transform(appId).optional()`

packages/react/src/config.ts
krzysu: maybe let's wait and see how the actual appId format/generation will look like and then we will decide if a helper is needed?

PR #229, opened 2023-03-15
feat: extends use of `observerId` to relevant adapter calls

cesarenaldi:


packages/react/src/environments.ts
```diff
@@ -20,6 +23,7 @@ export const production: EnvironmentConfig = {
     maxIndexingWaitTime: 120000,
     maxMiningWaitTime: 60000,
   },
+  handleResolver: (handle) => `${handle}.lens`,
```
krzysu, line 26: 👍

packages/react/src/profile/adapters/ProfileGateway.ts
```diff
@@ -6,21 +6,21 @@ import {
...
 
   async getAllProfilesByOwnerAddress(address: string): Promise<Profile[]> {
     const { data } = await this.apolloClient.query<
       GetAllProfilesByOwnerAddressQuery,
       GetAllProfilesByOwnerAddressQueryVariables
     >({
       query: GetAllProfilesByOwnerAddressDocument,
-      variables: { address, limit: 10, sources: this.sources },
+      // 'sources' and 'observerId' are not needed. We just use 'id' and 'handle' for now.
+      variables: { address, limit: 10 },
```
krzysu, line 23: I think sources are not used in any profile queries anyway, are they?

packages/react/src/transactions/infrastructure/ProfileCacheManager.ts
```diff
@@ -0,0 +1,70 @@
...
+
+    const { data } = await this.client.query<GetProfileQuery, GetProfileQueryVariables>({
+      query: GetProfileDocument,
+      variables: {
+        request: {
+          profileId: args.id,
+          handle: args.handle,
+        },
+        observerId: activeProfile?.id,
+        sources: this.sources,
```
krzysu, line 40: do we really need sources to fetch a profile?

cesarenaldi: It's to get these stats counting correctly: https://github.com/lens-protocol/lens-sdk/pull/229#discussion_r1137114156

PR #235, opened 2023-03-17
feat: Expose `collectNftAddress` from publication fragments

krzysu:


packages/api-bindings/src/graphql/__helpers__/fragments.ts
```diff
@@ -271,6 +271,7 @@ export function mockPostFragment(
     stats: mockPublicationStatsFragment(),
     metadata: mockMetadataFragment(),
     profile: mockProfileFragment(),
+    collectNftAddress: mockEthereumAddress(),
```
cesarenaldi, line 274: Given that a collect address is only meaningful for collectable publications, should we expose it via the `collectPolicy: CollectPolicy` aggregate so that it's available only on the relevant part of the `CollectPolicy` union?

krzysu: it would be a new field on most of the types in this union, yes?

```ts
type CollectPolicy =
  | FeeCollectPolicy
  | NoFeeCollectPolicy
  | LimitedFeeCollectPolicy
  | TimedFeeCollectPolicy
  | LimitedTimedFeeCollectPolicy
  | NoCollectPolicy;
  
// ie.
type FeeCollectPolicy = {
  type: CollectPolicyType.CHARGE;
  state: CollectState.CAN_BE_COLLECTED | CollectState.NOT_A_FOLLOWER;
  amount: ClientErc20Amount;
  referralFee: number;
  followerOnly: boolean;
  collectNftAddress: string; // new one
};
```
but NOT on `NoCollectPolicy`

I could do that.

@desfero what do you think about it?

cesarenaldi (approved):

PR #236, opened 2023-03-17
chore: release react 0.6.0 & client 0.3.0

krzysu:


packages/gated-content/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@lens-protocol/gated-content",
-  "version": "0.0.1",
+  "version": "0.0.2",
```
cesarenaldi, line 3: @desfero what do you think we should do here? It seems ok to me to have this package live its own versioning until we reach v1.0.0, which will re-sync all of them.

packages/react-web/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@lens-protocol/react-web",
-  "version": "0.4.1",
+  "version": "0.6.0",
```
cesarenaldi, line 3: This one is kinda takes over from `@lens-protoco/react` for all users sung the Lens SDK on a React Web app. Seems ok to have it v0.6.0. Less opportunity for confusions.

Thoughts?

cc @desfero , @krzysu

krzysu:
packages released, can I get approval to merge this to the `main`? @cesarenaldi @desfero

packages/gated-content/package.json
cesarenaldi: it's a dependency of `react` and `react-web`.

cesarenaldi (approved):

PR #237, opened 2023-03-17
feat: removes declaration maps from published packages

cesarenaldi:


packages/storage/tsconfig.json
```diff
@@ -1,5 +1,5 @@
 {
-  "extends": "./tsconfig.base.json",
+  "extends": ".@lens-protocol/tsconfig/tsconfig.base.json",
```
desfero, line 2: Seems typo
```suggestion
  "extends": "@lens-protocol/tsconfig/tsconfig.base.json",
```

desfero (approved):

PR #238, opened 2023-03-17
feat: exports some unreachable types

cesarenaldi:


desfero (approved):

krzysu:
just wonder, is there any tool/process to detect that?

cesarenaldi:
> just wonder, is there any tool/process to detect that?

I noticed that `typedoc` does help here. It flags types that are not documented that are "dragged in" by other types. In doing so it helps to figure out what is not exported.

PR #241, opened 2023-03-20
feat: allow to define sortCriteria for useExploreProfiles

krzysu:


packages/api-bindings/src/graphql/feed.graphql
```diff
@@ -90,13 +90,14 @@ query Feed(
 }
 
 query ExploreProfiles(
-  $observerId: ProfileId
```
cesarenaldi, line 93: Is removing of `@observerId` intentional?

cesarenaldi (approved):

PR #245, opened 2023-03-21
feat: supports Lens API versioning

cesarenaldi:


packages/api-bindings/src/apollo/links.ts
```diff
@@ -0,0 +1,107 @@
...
+        if (apiMajor && clientMajor && apiMajor > clientMajor) {
+          logger.warn(
+            `The Lens API ${apiVersion} is NOT supported by the Lens SDK support range ^${supportedVersion}. Update your Lens SDK client to the latest version.`,
+          );
+          return response;
+        }
+
+        if (apiMinor && clientMinor && apiMinor > clientMinor) {
+          logger.info(
+            `The Lens API ${apiVersion} is ahead of the Lens SDK support range ^${supportedVersion}. Check for a new version of the Lens SDK client, if available.`,
```
krzysu, line 77: so it will already log as the current API version is `x-api-version: 1.31.0` and we start here with `1.0.0`, yes?

krzysu (approved):
nice, I didn't even know that we can set it up on the apollo link level, I see how little I know about apollo.

PR #246, opened 2023-03-21
feat: adds `useProfiles` hook

cesarenaldi:


krzysu (approved):
I like that we use a single query now for 3 different hooks, I just wonder if it's better to offer 3 hooks to get profiles or unify them into one, so `useProfiles({ byIds, byHandles, byOwner, byWhoMirrored })` all in one hook.

PR #254, opened 2023-03-25
feat: removes `*Fragment` suffix and a lot of other nice things

cesarenaldi:


packages/api-bindings/codegen-api.yml
```diff
@@ -75,6 +78,24 @@ generates:
           skipTypename: true
           immutableTypes: false
           inlineFragmentTypes: combine
-      - 'typescript-react-apollo'
+          omitOperationSuffix: true
+          operationResultSuffix: 'Data'
+
+  src/graphql/hooks.ts:
```
krzysu, line 84: can we also add a comment on the very top of these two new files that they are auto-generated and to not make direct changes to these files (`src/graphql/operations.ts` and `src/graphql/hooks.ts`) as it is not obvious, compared to the previous `generated.tsx`.

packages/api-bindings/codegen-api.yml
```diff
@@ -75,6 +78,24 @@ generates:
...
+              '/* eslint-disable import/no-default-export */', # generatedTypes default export
+              '/* eslint-disable @typescript-eslint/no-explicit-any */', # see extensive use of FieldPolicy<any> | FieldReadFunction<any>
+              '/* eslint-disable @typescript-eslint/naming-convention */', # interface PossibleTypesResultData not starting with I
+              '/* eslint-disable no-restricted-imports */', # import * from @apollo/client
+              "import * as Operations from './operations';",
+            ]
+      - 'typescript-react-apollo':
+          pureMagicComment: true # keep
+          importOperationTypesFrom: 'Operations'
+          omitOperationSuffix: true # MUST goes with typescript-operations.omitOperationSuffix in src/graphql/operations.ts
```
krzysu, line 97: I guess this is the main change in this PR outside of renaming.
anything else I should be aware of?

cesarenaldi: Correct, this is the core of the change. Everything else is in support or a consequence of the change.

packages/api-bindings/codegen-api.yml
cesarenaldi: good call, done in the following PR: https://github.com/lens-protocol/lens-sdk/pull/255/commits/3b8f49847e69734923156ced24f2672a0a305b46

krzysu (approved):

PR #255, opened 2023-03-26
feat: get ready for TypeDoc documentation

cesarenaldi:


packages/api-bindings/src/graphql/common.graphql
```diff
@@ -22,25 +22,51 @@ fragment ModuleFeeAmount on ModuleFeeAmount {
   value
 }
 
-fragment ReferenceModule on ReferenceModule {
-  ... on FollowOnlyReferenceModuleSettings {
-    __typename
-    contractAddress
+fragment AaveFeeCollectModuleSettings on AaveFeeCollectModuleSettings {
```
krzysu, line 25: I see you already included new fragments from my other PR here 😄

krzysu (approved):

packages/api-bindings/src/graphql/common.graphql
cesarenaldi: Yes, I had to cause it was generating types with `| {}` otherwise. I will address the merge conflict if need to.

PR #269, opened 2023-03-30
chore: remove `resultCaching` in apollo config because its true by default

bigint:
https://www.apollographql.com/docs/react/caching/cache-configuration/#resultcaching


bigint:
cc @cesarenaldi 🙇🏼

cesarenaldi (approved):

cesarenaldi (approved):

PR #270, opened 2023-03-30
feat: segregates storage entries under env specific namespace

cesarenaldi:


cesarenaldi:
An interesting observation, in this PR I touched `@lens-protocol/client`, `@lens-protocol/gated-content` and `@lens-protocol/react`. But in the changeset I added `@lens-protocol/react-web` cause I think it would be appropriate for the changelog to mention this change. For many users that use `@lens-protocol/react-web` the `@lens-protocol/react` package is an implementation detail. Thoughts?

image

packages/gated-content/src/environments.ts
```diff
@@ -3,18 +3,21 @@ import { EthereumAddress } from '@lens-protocol/shared-kernel';
 import { SupportedChains } from './conditions/types';
 
 export type EnvironmentConfig = {
+  name: string;
```
bigint, line 6: Instead of free string type we can define the values right? something like this

`'prod' | 'stg' | 'sandbox'`

packages/client/src/consts/environments.ts
```diff
@@ -1,13 +1,13 @@
...
 
   get gqlEndpoint() {
     return this.url;
   }
 }
 
-export const polygon = new Environment('https://api.lens.dev');
-export const mumbai = new Environment('https://api-mumbai.lens.dev');
+export const polygon = new Environment('production', 'https://api.lens.dev');
+export const mumbai = new Environment('staging', 'https://api-mumbai.lens.dev');
```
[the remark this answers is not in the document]
krzysu: another thing to consider, how would it look like if lens was deployed to more chains?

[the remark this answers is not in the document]
cesarenaldi: I understand the desire to refer to the environment by using the blockchain name (polygon, mumbai) and so on. But the Lens Protocol is growing beyond the single blockchain.

Even by just looking at the features... there is a concept of groups of configurations that are legitimate and groups that are not.

Consider NFT PFP. On a "live" environment one can use an Ethereum Mainnet NFT as PFP for a Protocol deployed on Polygon Mainnet. Another valid group is Ethereum Testnet and Mumbai. But there is no meaning to the combination Ethereum Mainnet and Mumbai OR Ethereum Testnet and Polygon Mainnet.

The same applies to token-gated content. There are rules that can be evaluated cross-chain but only mainnet-mainnet and testnet-testnet pairs are valid combination.

The intention with the client environment config is not just about the API deployment, but also about understanding these groups.

Production:
- Polygon: Mainenet
- Ethereum: Mainnet
- API: https://api.lens.dev

Staging:
- Polygon: Mumbai
- Ethereum: Testnet
-API: https://api-mumbai.lens.dev

Sandbox:
- Polygon: Mumbai
- Ethereum: Testnet
- API: https://api-sandbox-mumbai.lens.dev

krzysu (approved):

packages/react/src/environments.ts
```diff
@@ -66,3 +74,20 @@ export const staging: EnvironmentConfig = {
...
+ * After extensive considerations, we have decided to rename the `staging` variable into `development`.
+ * We found that the term `staging` is inflated and not very clear to the all users.
+ *
+ * Together with {@link `production`} the changes is meant to be more explicit about the intended usage of these variables.
+ *
+ * - `production` is the environment to be used in the live instance of your application (real users, real profiles, real data).
+ * - `development` is the environment to be used when you develop and test your application (test users, test profiles, test data).
+ *
+ * We also aligned the naming in the `@lens-protocol/client` package to enable interoperability between the two packages.
+ */
```
bigint, line 92:
```suggestion

/**
 * The development environment configuration
 *
 * @deprecated Please use the {@link development} variable instead
 *
 * After extensive considerations, we have decided to rename the `staging` variable into `development`.
 * We found that the term `staging` is inflated and unclear to all users.
 *
 * Together with {@link `production`} the changes is meant to be more explicit about the intended usage of these variables.
 *
 * - `production` is the environment to be used in the live instance of your application (real users, real profiles, real data).
 * - `development` is the environment to be used when you develop and test your application (test users, test profiles, test data).
 *
 * We also aligned the naming in the `@lens-protocol/client` package to enable interoperability between the two packages.
 */
```

packages/client/src/consts/environments.ts
```diff
@@ -1,13 +1,45 @@
...
+export const development = new Environment('development', 'https://api-mumbai.lens.dev');
+
+/**
+ * @deprecated Please use the {@link production} variable instead
+ *
+ * After extensive considerations, we have decided to rename the `polygon` variable into `production`.
+ * See also the deprecated {@link `mumbai`} variable.
+ *
+ * The changes is meant to be more explicit about the intended usage of the variables.
+ * It also helps to widen the meaning of these as we add features that are not limited to Polygon blockchain.
```
bigint, line 26:
```suggestion
 * The changes are meant to be more explicit about the intended usage of the variables.
 * It also helps to widen the meaning of these as we add features that are not limited to Polygon blockchain.
```

packages/client/src/consts/environments.ts
```diff
@@ -1,13 +1,45 @@
...
+ */
+export const polygon = production;
+
+/**
+ * @deprecated Please use the {@link development} variable instead
+ *
+ * After extensive considerations, we have decided to rename the `mumbai` variable into `development`.
+ * See also the deprecated {@link `polygon`} variable.
+ *
+ * The changes is meant to be more explicit about the intended usage of the variable.
```
bigint, line 36:
```suggestion
 * The changes are meant to be more explicit about the intended usage of the variable.
```

bigint (changes requested):

packages/react/src/environments.ts
cesarenaldi: I see, I'll reword it "unclear to all users." is not what I was aiming to say. Thank for flagging it as not clear.

packages/client/src/consts/environments.ts
cesarenaldi: Reworded and incorporated your suggestion.

packages/client/src/consts/environments.ts
cesarenaldi: Reworded and incorporated your suggestion.

packages/client/src/consts/environments.ts
```diff
@@ -1,13 +1,45 @@
...
 
-export const polygon = new Environment('https://api.lens.dev');
-export const mumbai = new Environment('https://api-mumbai.lens.dev');
+export type { Environment };
 
-// not exposed from the package, used in tests
-export const mumbaiSandbox = new Environment('https://api-sandbox-mumbai.lens.dev');
+export const production = new Environment('production', 'https://api.lens.dev');
+
+export const development = new Environment('development', 'https://api-mumbai.lens.dev');
```
bigint, line 17: I'd always suggest not to hardcode the values, better if we can have an exported Environment enum and consume it everywhere.

WDYT?

packages/gated-content/src/environments.ts
```diff
@@ -3,18 +3,21 @@ import { EthereumAddress } from '@lens-protocol/shared-kernel';
 import { SupportedChains } from './conditions/types';
 
 export type EnvironmentConfig = {
+  name: 'production' | 'development';
```
bigint, line 6: Also we can consume the exported enum here aswell

bigint (changes requested):

packages/gated-content/src/environments.ts
krzysu: that can cause unwanted dependencies between packages, I don't know now, but I think it's not a good time to make changes like that just before releasing v1, let's create a follow-up ticket to consider these comments, wdyt?

cesarenaldi: `EnvironmentConfig` is an internal anywhere else and this whole package `@lens-protocol/gated-content` is an internal (not meant to be used directly).

This simply because the Lens API is not open source at the moment. So nobody can produce a complete environment deployment of the protocol+api+custom modules+gated contracts. So making it configurable seems not necessary at this stage.

Environment config (here but also in other parts of the SDK) are groups of cohesive configuration that needs to account for all these aspect of they won't simply work. So at this point in time the public environment we are going to offer to the public are just 2: `production` and `development`.

PR #273, opened 2023-03-31
chore(ci): use node version from .nvmrc and enable cache

bigint:
```yml
node-version-file: '.nvmrc'
cache: 'pnpm'
```

This will improve the CI speed

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -16,7 +16,8 @@ jobs:
       - name: Use Node.js
         uses: actions/setup-node@v3
         with:
-          node-version: "18.x"
+          node-version-file: '.nvmrc'
+          cache: 'pnpm'
 
       - name: Setup pnpm
         uses: pnpm/action-setup@v2.2.4
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -13,16 +13,16 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
+      - name: Setup pnpm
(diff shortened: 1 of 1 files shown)


krzysu:
the build failed with `Error: Unable to locate executable file: pnpm.` so something is wrong here

bigint:
@krzysu fixed ✅ 

Error because we need to move pnpm before node setup

Also here is the cache where it gets stored https://github.com/lens-protocol/lens-sdk/actions/caches

Screenshot 2023-03-31 at 1 17 47 PM

I'll create another PR to remove the cache when a PR gets merged

.github/workflows/ci.yaml
```diff
@@ -13,15 +13,16 @@ jobs:
...
         with:
-          node-version: "18.x"
-
-      - name: Setup pnpm
-        uses: pnpm/action-setup@v2.2.4
+          node-version-file: '.nvmrc'
+          cache: 'pnpm'
 
-      - run: pnpm install
+      - run: pnpm install --frozen-lockfile
```
krzysu, line 25: I hope this will work with Renovate - our automated dependency updates bot.

bigint: `frozen-lockfile` here in ci is just not to touch the lock file after pnpm install, so it will remains the same as in the repo.

This not going to affect Renovate in anyway 🙇🏼

krzysu: ok, let's give it a try :)

krzysu (approved):

PR #279, opened 2023-03-31
feat: client sdk - Add TSDoc for all available methods of submodules of LensClient

krzysu:
Add TSDoc for all exposed methods of submodules of LensClient

Link T-9220

### DIFF
--- a/packages/client/src/authentication/Authentication.ts
+++ b/packages/client/src/authentication/Authentication.ts
@@ -6,12 +6,39 @@ import { CredentialsExpiredError, NotAuthenticatedError } from '../consts/errors
 import { AuthenticationApi } from './adapters/AuthenticationApi';
 import { CredentialsStorage } from './adapters/CredentialsStorage';
 
+/**
+ * Authentication for Lens API.
+ *
+ * @group LensClient Modules
+ */
 export interface IAuthentication {
+  /**
+   * Generate a challenge string for the wallet to sign.
+   *
+   * @param address - The wallet address
+   * @returns A challenge string
+   */
   generateChallenge(address: string): Promise<string>;
+
+  /**
+   * Authenticate the user with the wallet address and signature of the challenge.
+   *
+   * @param address - The wallet address
+   * @param signature - The signature of the challenge
+   */
   authenticate(address: string, signature: string): Promise<void>;
+
+  /**
+   * Check if the user is authenticated. If the credentials are expired, try to refresh them.
+   *
+   * @returns Whether the user is authenticated
+   */
   isAuthenticated(): Promise<boolean>;
 }
 
+/**
+ * Authentication for Lens API. Request challenge, authenticate, manage credentials.
+ */
 export class Authentication implements IAuthentication {
   private readonly api: AuthenticationApi;
   private readonly storage: CredentialsStorage;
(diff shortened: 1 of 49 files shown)


packages/client/src/feed/Feed.ts
```diff
@@ -42,6 +63,22 @@ export class Feed {
     });
   }
 
+  /**
+   * Fetch feed highlights.
+   *
+   * ⚠️ Requires authenticated LensClient.
```
cesarenaldi, line 69: Nice touch with the ⚠️!

packages/client/src/authentication/Authentication.ts
```diff
@@ -6,12 +6,39 @@ import { CredentialsExpiredError, NotAuthenticatedError } from '../consts/errors
...
+ * @group LensClient Modules
+ */
 export interface IAuthentication {
+  /**
+   * Generate a challenge string for the wallet to sign.
+   *
+   * @param address - The wallet address
+   * @returns A challenge string
+   */
   generateChallenge(address: string): Promise<string>;
```
cesarenaldi, line 21: Maybe for a future iteration. We do have a alias called `EthereumAddress` in `shared-kernel` which can help make signatures more clear. In the future we might also brand the `EthereumAddress` so that, with nominal typing derived from the branding, it's not possible to make mistakes.

packages/client/src/explore/Explore.ts
```diff
@@ -18,6 +23,22 @@ export class Explore {
     this.authentication = authentication;
   }
 
+  /**
+   * Explore publications
+   *
+   * @param request - Request object for the query
+   * @param observerId - Optional id of a profile that is the observer for this request
+   * @returns Array of {@link PublicationFragment} wrapped in the {@link PaginatedResult} helper
```
cesarenaldi, line 31: nit: 
```suggestion
   * @returns Array of {@link PublicationFragment} wrapped in {@link PaginatedResult}
```

IMO "helper" suggests a instance of something or a function but it's just a type.

If you decide to action this, there are several occurrences of this text.

packages/client/src/helpers/buildPaginatedQueryResult.ts
```diff
@@ -10,12 +8,10 @@ export type PaginatedQueryData<Item> = {
   items: Item[];
 };
 
-export type PaginatedResult<T> = Prettify<
-  PaginatedQueryData<T> & {
-    next(): Promise<PaginatedResult<T> | null>;
-    prev(): Promise<PaginatedResult<T> | null>;
-  }
->;
+export type PaginatedResult<T> = PaginatedQueryData<T> & {
```
cesarenaldi, line 11: Does the `&` achieve the same result as `Prettify` in the resulting typedef in VS Code suggestions?

packages/client/src/index.ts
cesarenaldi, line 1: It's a long PR, where are these exported from now?

cesarenaldi (commented):
Looks pretty good, I have a couple of questions and small nit comment.

packages/client/src/index.ts
krzysu: I've found that these are not required to be exported at all, before the typedoc was complaining about them, but not anymore. I guess, it's because of the improved return types (defined fragments and not `InferResultType` output).

cesarenaldi (approved):

PR #282, opened 2023-04-03
feat: self-fund protocol calls when subsidized approaches fails

cesarenaldi:
Integration branch for T-3086


packages/react/src/transactions/infrastructure/__tests__/TransactionFactory.spec.ts
```diff
@@ -373,148 +373,129 @@ describe(`Given an instance of the ${TransactionFactory.name}`, () => {
...
+          indexingEventsSequence: [proxyActionStatusEvent],
+        });
+        const factory = setupTransactionFactory({ observer });
+
+        const transaction = factory.createProxyTransaction(init);
+        const result = await transaction.waitNextEvent();
+
+        expect(result.unwrap()).toBe(TransactionEvent.UPGRADED);
+        expect(transaction.hash).toEqual(proxyActionStatusEvent.txHash);
+        expect(transaction.status).toBe(ProxyActionStatus.TRANSFERRING);
```
cesarenaldi, line 498: Refactored these tests cause it was very unclear and confusing and from the test alone I couldn't figure out if the relevant logic was affected by the changes I need to do or not.

packages/react/src/profile/useCreateProfile.ts
```diff
@@ -29,31 +32,63 @@ export type CreateProfileOperation = Operation<
...
+ *         Create profile
+ *       </button>
+ *     </div>
+ *   );
+ * }
+ * ```
+ *
+ * @example Programmatic error handling
+ * ```tsx
+ * import { useCreateProfile, DuplicatedHandleError } from '@lens-protocol/react-web';
```
krzysu, line 65: should example in `react` package mention import from `react-web` package?

krzysu (approved):

packages/react/src/profile/useCreateProfile.ts
krzysu: I see, maybe let's just leave it as it is then, till we have a separate package for react-native, then we will have to figure it out

PR #286, opened 2023-04-06
feat: client sdk - add support for DA publications

krzysu:


packages/client/src/publication/index.ts
```diff
@@ -1,8 +1,8 @@
 export type {
-  CreateCollectBroadcastItemResultFragment,
-  CreateCommentBroadcastItemResultFragment,
-  CreateMirrorBroadcastItemResultFragment,
-  CreatePostBroadcastItemResultFragment,
+  CreateCollectTypedDataFragment,
+  CreateCommentTypedDataFragment,
+  CreateMirrorTypedDataFragment,
+  CreatePostTypedDataFragment,
```
krzysu, line 5: there is no need for keeping backward compatibility as these types are only introduced in #279

examples/node/scripts/createDAPostViaDispatcher.ts
```diff
@@ -0,0 +1,67 @@
...
+    await lensClient.publication.createDataAvailabilityCommentViaDispatcher({
+      from: profileId,
+      contentURI,
+      commentOn: createPostResultValue.id,
+    });
+
+  const createCommentResultValue = createCommentResult.unwrap();
+
+  if (!isCreateDataAvailabilityPublicationResult(createCommentResultValue)) {
+    console.log(`Something went wrong`, createCommentResultValue);
```
bigint, line 60: Do we need to consider this as template strings?

or we can make use of it

```suggestion
    console.log(`Something went wrong ${createCommentResultValue}`);
```

examples/node/scripts/createDAPostViaDispatcher.ts
```diff
@@ -0,0 +1,67 @@
...
+    return;
+  }
+
+  console.log(`DA comment was created too: `, createCommentResultValue);
```
bigint, line 64:
```suggestion
  console.log(`DA comment was created too: ${createCommentResultValue}`);
```

examples/node/scripts/createDAPostViaTypedData.ts
```diff
@@ -0,0 +1,69 @@
...
+  const signedTypedData = await wallet._signTypedData(
+    createPostTypedDataValue.typedData.domain,
+    createPostTypedDataValue.typedData.types,
+    createPostTypedDataValue.typedData.value
+  );
+
+  console.log(`Broadcasting signed createDataAvailabilityPostTypedData...`);
+
+  const broadcastResult = await lensClient.transaction.broadcastDataAvailability({
+    id: createPostTypedDataValue.id,
```
bigint, line 54: I think we can destructure all variables 🙇🏼

Something like 

```
const { id, typedData } = createPostTypedDataResult.unwrap();
```

examples/node/scripts/createDAPostViaTypedData.ts
```diff
@@ -0,0 +1,69 @@
...
+    signature: signedTypedData,
+  });
+
+  // broadcastResult is a Result object
+  const broadcastValue = broadcastResult.unwrap();
+
+  if (!isCreateDataAvailabilityPublicationResult(broadcastValue)) {
+    console.log(`Something went wrong`, broadcastValue);
```
bigint, line 62:
```suggestion
    console.log(`Something went wrong ${broadcastValue}`);
```

examples/node/scripts/createDAPostViaTypedData.ts
```diff
@@ -0,0 +1,69 @@
...
+    return;
+  }
+
+  console.log(`DA post was created: `, broadcastValue);
```
bigint, line 66:
```suggestion
  console.log(`DA post was created: ${broadcastValue}`);
```

bigint (changes requested):

examples/node/scripts/createDAPostViaDispatcher.ts
krzysu: no, it won't work, as I want to print out an object, not a string, so with your suggestion it ends up as `[object Object]`.  
Also please don't block a PR with "request changes" if you only want to suggest something insignificant in the scope of the PR.

compare
```
console.log(`Transaction status: `, wasIndexedFirstCheck.unwrap());
// vs.
console.log(`Transaction status: ${wasIndexedFirstCheck.unwrap()}`);
```

result
```
Transaction status:  {
  __typename: 'TransactionIndexedResult',
  indexed: false,
  txHash: '0xdfbf48b549813fc1c4654122132223d4ceccc9b1b6e71dbbec609560b4838fc0'
}
// vs.
Transaction status: [object Object]
```

examples/node/scripts/createDAPostViaTypedData.ts
krzysu: this is an example script for integrators to understand how to use the SDK in different scenarios, I assumed the more meaningful names of consts/vars, the more self-explanatory the example is. I don't think restructuring makes any difference here, for sure it doesn't improve the clarity of the example. In the commented line of code, it is clear where the `id` is supposed to come from.

cesarenaldi (approved):
LGTM

Issue #285, opened 2023-04-06
@lens-protocol/client - XMLHttpRequest error on Cloudflare functions (Remix loader)


korzewski:
Hey everyone!
Today I started to play around with Lens API's.
GraphQL queries works great, although I wanted to get advantage of types by using this package **@lens-protocol/client**. 

Unfortunately it seems to not work at Cloudflare functions:
_ReferenceError: XMLHttpRequest is not defined_
I'm using Remix loader here^^

**To Reproduce**
- Create new Remix app **npx create-remix@latest**
- Deployment configuration **Cloudflare Pages**
- At any route create loader function and run any query
 ```
const result = await lensClient.explore.profiles({
    sortCriteria: ProfileSortCriteria.CreatedOn,
});
 ```
- I got an error `ReferenceError: XMLHttpRequest is not defined` - As Cloudflare functions supports only Fetch API's.

Worth to mention same code used at frontend **useEffect()** works as expected.

**Expected behaviuor**
Lens client would work at Remix Cloudflare Pages without errors

**Version of the packages you are using**
- "@lens-protocol/client": "^0.3.0",
- "@remix-run/cloudflare": "^1.15.0",
- "@remix-run/cloudflare-pages": "^1.15.0",

**Additional context**
Are there any plans to switch over to Fetch API? Or some other known workaround for it?

Thanks

desfero:
Thanks for reporting the issue @korzewski.
It seems under the hood lib that we use is using by default XMLHTTPRequest (quite weird to still use it given how broadly available across ecosystems the `fetch API` is 🤔 ).
see: https://github.com/jasonkuhrt/graphql-request/issues/399

Anyway, the fix is very simple, will try to talk with the team to push it ASAP.

PR #287, opened 2023-04-11
feat: client sdk - use fetch with graphql

krzysu:
fixes https://github.com/lens-protocol/lens-sdk/issues/285

### DIFF
--- a/packages/client/src/helpers/FetchGraphQLClient.ts
+++ b/packages/client/src/helpers/FetchGraphQLClient.ts
@@ -1,0 +1,11 @@
+import { GraphQLClient } from 'graphql-request';
+import { PatchedRequestInit } from 'graphql-request/dist/types';
+
+export class FetchGraphQLClient extends GraphQLClient {
+  constructor(url: string, options?: PatchedRequestInit) {
+    super(url, {
+      ...options,
+      fetch: fetch,
+    });
+  }
+}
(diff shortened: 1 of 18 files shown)


packages/client/src/helpers/FetchGraphQLClient.ts
```diff
@@ -0,0 +1,11 @@
+import { GraphQLClient } from 'graphql-request';
+import { PatchedRequestInit } from 'graphql-request/dist/types';
```
desfero, line 2: could that be possible to import types from `graphql-request` (not from `dist`)?

If the params are not exported we could try to get one on our own

```
type ClientOptions = ConstructorParameters<GraphQLClient>[1];
type Options = Omit<ClientOptions, "fetch">;
```

packages/client/tsconfig.json
```diff
@@ -2,6 +2,7 @@
   "extends": "@lens-protocol/tsconfig/base.json",
   "compilerOptions": {
     "module": "ESNext",
+    "lib": ["DOM", "ESNext"],
```
desfero, line 5: nit: personally would just copy the `fetch` type definitions into some `typings` folder with a link to the issue as `DOM` opens the possibility to use some DOM-specific api and break node.js integrations down the line.

desfero (approved):

krzysu:
closing this in favor of https://github.com/lens-protocol/lens-sdk/pull/290

PR #294, opened 2023-04-13
feat: adds `fallback: RequestFallback` to `BroadcastingError`

cesarenaldi:
Most of these details are drastically changes in https://github.com/lens-protocol/lens-sdk/pull/297

I'll merge this into the integration branch to not cause odd conflict, but the version in the latest PR of this series will be the one to go live.

### DIFF
--- a/packages/api-bindings/codegen-api.yml
+++ b/packages/api-bindings/codegen-api.yml
@@ -41,6 +41,7 @@ config:
     ProfileId: ProfileId
     ProfileInterest: string
     ProxyActionId: string
+    PublicationUrl: Url
     ReferenceModuleData: string
     ReferencePolicy: ReferencePolicy
     Search: string
--- a/packages/api-bindings/codegen-api.yml
+++ b/packages/api-bindings/codegen-api.yml
@@ -24,6 +24,7 @@ config:
     DecryptionCriteria: DecryptionCriteria
     EncryptedValueScalar: string
     EthereumAddress: EthereumAddress
+    FollowModuleData: string
     FollowPolicy: FollowPolicy
     FollowStatus: FollowStatus
     Handle: string
--- a/packages/api-bindings/codegen-api.yml
+++ b/packages/api-bindings/codegen-api.yml
@@ -12,6 +12,7 @@ config:
   scalars:
     Attach: string
     BroadcastId: string
+    BlockchainData: string
     ChainId: number
     ClientErc20Amount: ClientErc20Amount
     CollectModuleData: string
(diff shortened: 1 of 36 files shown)


packages/react/src/transactions/adapters/TypedData.ts
```diff
@@ -1,7 +1,8 @@
-import { TypedDataDomain, TypedDataField } from 'ethers';
+import { Eip712TypedDataDomain, OmitTypename } from '@lens-protocol/api-bindings';
+import { TypedDataField } from 'ethers';
 
 export type TypedData = {
-  domain: TypedDataDomain;
+  domain: OmitTypename<Eip712TypedDataDomain>;
```
cesarenaldi, line 5: Better type. Not strictly needed by the implementation but helped with the test helpers and it's better overall.

packages/api-bindings/codegen-api.yml
```diff
@@ -41,6 +41,7 @@ config:
     ProfileId: ProfileId
     ProfileInterest: string
     ProxyActionId: string
+    PublicationUrl: Url
```
cesarenaldi, line 44: The only reason I regenerated the GQL types was this resolving w/ `unknown` so I am avoiding several TS assertion to `string` with this.

packages/domain/src/use-cases/transactions/BroadcastingError.ts
```diff
@@ -18,7 +27,7 @@ export class BroadcastingError extends Error {
   /**
    * @internal
    */
-  constructor(readonly reason: BroadcastingErrorReason) {
+  constructor(readonly reason: BroadcastingErrorReason, readonly fallback?: RequestFallback) {
```
cesarenaldi, line 30: `fallback` is optional for now that not all codepaths that generates it are changed to produce one. After all changes I'll review if it's possible to make it non-optional.

packages/react/src/transactions/adapters/__helpers__/mocks.ts
```diff
@@ -118,3 +124,27 @@ export function mockIMetadataUploader(urlOrError: Url | Error): IMetadataUploade
...
+  expectedBroadcastingErrorReason: BroadcastingErrorReason,
+  typedData: TypedData,
+) {
+  assertFailure(result);
+  expect(result.error.reason).toEqual(expectedBroadcastingErrorReason);
+  expect(result.error.fallback).toMatchObject({
+    contractAddress: typedData.domain.verifyingContract,
+    encodedData: expect.any(String),
+  });
+}
```
cesarenaldi, line 150: I created these helpers to consolidate and simplify test assertions in the coming changes.

PR #297, opened 2023-04-14
feat: implement pay transaction fallback

cesarenaldi:


packages/react/src/transactions/useSelfFundedFallback.ts
```diff
@@ -0,0 +1,86 @@
...
+}
+
+export type SelfFundedOperation = Operation<
+  void,
+  PendingSigningRequestError | UserRejectedError | WalletConnectionError,
+  [SelfFundedOperationRequest]
+>;
+
+/**
+ * `useExploreProfiles` is an hook that let you retry a failed operation that could be self-funded.
```
krzysu, line 49: typo: `useSelfFundedFallback` is an hook that let you retry a failed operation that could be self-funded.

packages/react/src/transactions/useSelfFundedFallback.ts
```diff
@@ -0,0 +1,86 @@
...
+ * @example Create a post
+ * ```ts
+ * import { ContentFocus, ProfileOwnedByMe, supportsSelfFundedFallback, useCreatePost, useSelfFundedFallback } from '@lens-protocol/react-web';
+ *
+ * function PostComposer({ publisher }: { publisher: ProfileOwnedByMe }) {
+ *   const { execute: createPost, error, isPending } = useCreatePost({ publisher, upload: uploadToIpfs });
+ *   const { execute: createPostFromWallet, error: selfFundedError, isPending: selfFundedPending } = useSelfFundedFallback();
+ *
+ *   const submit = async (content: string) => {
+ *     const result = await create({
```
krzysu, line 63: typo: await createPost

packages/domain/src/use-cases/transactions/BroadcastingError.ts
```diff
@@ -1,22 +1,6 @@
...
-export enum BroadcastingErrorReason {
-  /**
-   * The gas-less broadcasting of the tx was rejected, probably due to reaching a quota limit
-   */
-  REJECTED = 'REJECTED',
-  /**
-   * A not specified reason
-   */
-  UNSPECIFIED = 'UNSPECIFIED',
-}
```
cesarenaldi, line 14: Turns out this is not really necessary at this level. All usage of this detail were quite a speculation, so removed it for now.

packages/react/src/transactions/adapters/SelfFundedProtocolCallGateway.ts
```diff
@@ -0,0 +1,68 @@
...
+> implements IPayTransactionGateway<TSelfFundedRequest>
+{
+  constructor(private readonly providerFactory: IProviderFactory) {}
+
+  async prepareSelfFundedTransaction(
+    request: TSelfFundedRequest,
+    wallet: Wallet,
+  ): Promise<UnsignedTransaction<TSelfFundedRequest>> {
+    const provider = await this.providerFactory.createProvider({
+      chainType: ChainType.POLYGON,
```
krzysu, line 43: why sometimes we have Polygon hardcoded and sometimes we use chainType from the arguments? (`waitForExecuted` or `createApproveTransaction`)
would it be possible to submit self funded tx from Mumbai then?

krzysu (approved):

packages/react/src/transactions/adapters/SelfFundedProtocolCallGateway.ts
cesarenaldi: On the `ChainType` specifically worth to clarify something. Bear in mind that `ChainType` is not the same as `ChainId`.

`ChainType` is a way for us to distinguish between Ethereum blockchains and Polygon blockchains within the current environment configuration. It's an abstraction on the nature of the blockchain not it's current specific network.

The Lens Protocol + API currently has some capabilities that are cross chain. There are domain policies that DO hold true no matter if the environment is `production` or `development`. Those domain policies dictates for example that on the "polygon" network (not matter if the Polygon Mainnet or Mumbai, which depends on the environment config) we have the Lens Protocol contracts, so that's where we are supposed to send our Protocol related interaction.

This is fundamentally why the `ChainType.POLYGON` is a times hard-coded. In the case above there is no much meaning to talk about "self funded protocol call" outside Polygon, the Lens Protocol contract are deployed on polygon networks only.

To continue on the domain policies that are true regardless of the chain IDs, we also have some regarding the "ethereum" network (again not matter if Mainnet or Goerli, the important bit is that if the environment is configured correctly so that we have Mainnet-Polygon[^1] OR Goerli-Mumbai pairs [^2]). Some examples are NFT ownership for the PFP (which we don't fully support yet at the hooks level) and token-gated criteria.

So, yes, it's possible to do a self funded tx on Mumbai. It's actually quite a normal thing.

[^1]: In `production` environment `ChainType.ETHEREUM` correspond to chain with chain ID: 1, while `ChainType.POLYGON` is chain with chain ID: 137.

[^2]: In `development` environment `ChainType.ETHEREUM` correspond to chain with chain ID: 5 (goerli), while `ChainType.POLYGON` is chain with chain ID: 80001 (mumbai).

PR #305, opened 2023-04-19
feat: support more metadata fields in create publication hooks (and publication queries)

krzysu:


packages/react/src/transactions/infrastructure/__tests__/createPublicationMetadata.spec.ts
```diff
@@ -121,36 +126,65 @@ describe(`Given the ${createPublicationMetadata.name} helper`, () => {
...
 
-          expect(metadata.media?.[0]).not.toHaveProperty('altTag');
-          expect(metadata.media?.[0]).not.toHaveProperty('cover');
+          expect(metadata.animation_url).toBeDefined();
+          expect(metadata.appId).toBeDefined();
+          expect(metadata.contentWarning).toEqual('NSFW');
+          expect(metadata.external_url).toBeDefined();
+          expect(metadata.image).toBeDefined();
+          expect(metadata.imageMimeType).toBeDefined();
+          expect(metadata.tags).toBeDefined();
```
cesarenaldi, line 187: I think we need to support `cover` as well. I would assume this is just a URL of the original image (like for `media.cover` and `media.url`). I asked cause it's not documented in the docs.lens.xyz.

cesarenaldi: Ok, the origin of this seems not clear, but it's basically the `image` field exposed as `MediaSet` (which currently is not that more useful) but with `optimized` image URL being added to `MediaSet` it can become useful.

So there is technically nothing to do on the write path.

krzysu: afaik the `cover` is not even mentioned in the metadata specs, I didn't test it but I guess it would not pass the verification.

PR #306, opened 2023-04-19
feat: ensures self-funded protocol call tx are waited for until indexed

cesarenaldi:


packages/domain/src/entities/__helpers__/mocks.ts
```diff
@@ -229,7 +229,6 @@ type MockedNativeTransactionInit<T extends TransactionRequestModel> = {
   chainType?: ChainType;
   hash?: string;
   id?: string;
-  indexingId?: string;
```
cesarenaldi, line 232: just some leftover from ages ago.

packages/domain/src/use-cases/transactions/SupportedTransactionRequest.ts
```diff
@@ -24,3 +25,17 @@ export type SupportedTransactionRequest =
...
+  TransactionKind.CREATE_POST,
+  TransactionKind.CREATE_PROFILE,
+  TransactionKind.FOLLOW_PROFILES,
+  TransactionKind.MIRROR_PUBLICATION,
+  TransactionKind.UPDATE_PROFILE_IMAGE,
+  TransactionKind.UNFOLLOW_PROFILE,
+  TransactionKind.UPDATE_PROFILE_DETAILS,
+  TransactionKind.UPDATE_FOLLOW_POLICY,
+  TransactionKind.UPDATE_DISPATCHER_CONFIG,
+];
```
cesarenaldi, line 41: I would like to generate a union out of this so to type more correctly the places that currently accepts `SupportedTransactionRequest` giving the wrong message that all Lens SDK supported tx can be: meta-tx, native-tx, proxy-actions, etc).

I might do this as initial refactoring for DA, cause that's another subset of the supported tx.

krzysu (approved):

PR #309, opened 2023-04-21
Include cover image

mvanhalen:
This will add the cover image used for audio collects and maybe others. So we can make it look nice.

### DIFF
--- a/packages/client/src/graphql/fragments.graphql
+++ b/packages/client/src/graphql/fragments.graphql
@@ -268,6 +268,9 @@ fragment Metadata on MetadataOutput {
   attributes {
     ...MetadataAttributeOutput
   }
+  cover {
+    ...MediaSet
+  }
 }
 
 fragment MetadataAttributeOutput on MetadataAttributeOutput {


krzysu:
as I understand `cover` is a remaining of some old implementation, it's not supported in PublicationMetadata standard, and the `image` attribute has exactly the same value. I've just merged another PR with many more publication fields, maybe that would be enough for your use case @mvanhalen ?
See https://github.com/lens-protocol/lens-sdk/pull/305/files#diff-bac6d0767ba3c63c12ff5fefe05bada8b54b701c4848a62b8b6e28839f5e8082

krzysu:
@mvanhalen could you give me an example publication id with `cover` you want to use and I'll double check if the same values are available as `image`. Just to be sure.

mvanhalen:
No problem @krzysu here an example of audio with cover https://lenster.xyz/posts/0x01c54f-0x39 (0x01c54f-0x39) and a video with cover https://lenster.xyz/posts/0xadb7-0x40 (0xadb7-0x40) idea is the image is a placeholder (video) or album cover (audio).

mvanhalen:
Great we are using it for [@collectz.lens](https://lenster.xyz/u/collectz) going live any day now.

krzysu:
sorry, yes, we started releasing `@next` versions, the latest is `1.1.0-next.3` so you install it like 
```
npm install @lens-protocol/react-web@next
```
I hope this will work for you

mvanhalen:
@krzysu thansk. We use https://www.npmjs.com/package/@lens-protocol/client package which has next 0 I tried that one but it did not include the image for audio main content focus it seems. Maybe it did not get the update yet?

> sorry, yes, we started releasing `@next` versions, the latest is `1.1.0-next.3` so you install it like
> 
> ```
> npm install @lens-protocol/react-web@next
> ```
> 
> I hope this will work for you

krzysu:
our bad, but we've just released a new `next` version for LensClient (`1.1.0-next.1`), this time it contains changes you need

sorry that it took so long time 🙏

PR #315, opened 2023-05-03
Draft: Add proposed gated content changes

paulxuca:


cesarenaldi (changes requested):
I think the approach of using the `IEncryptionProvider` interface to override such an internal breaks the encapsulation of the `GatedClient` in a way that skewed towards Lit Protocol. bear in mind our integration with Lit Protocol is designed like one-of-possibly-many encryption providers. This is enabled at the API level and it would go against the Lens direction if we were to lock it down to one provider or suggest an extension approach that does not scale (we cannot assume the `IEncryptionProvider` can be used for any random 3rd party provider implementation detail).

I brainstormed a different approach with @desfero and we cam to the conclusion that it is better to amend the react package [`EncryptionConfig`](https://github.com/lens-protocol/lens-sdk/blob/ccbe1c9efb32f92065ded0834f39b3ace52ac4b3/packages/react/src/config.ts#L64) so that is accepts an optional `GatedClient` factory function:

```ts
export type EncryptionConfig = {
  authentication: AuthenticationConfig;
  provider: IEncryptionProvider;
  
  /**
   * @internal
   */
  createGatedClient: typeof createGatedClient
};
```

And in our use case we can then provide a extended version of the `GatedClient` that provides a bespoke Lit `NodeClient`.

PR #327, opened 2023-05-08
fix: clarify create publication request models 

cesarenaldi:


packages/domain/src/use-cases/publications/CreatePost.ts
```diff
@@ -8,69 +8,116 @@ import {
...
+
+export type CreateEmbedPostRequest = BasePostRequest & {
   /**
-   * This is the URL that will appear below the asset's image on OpenSea and others
-   * and will allow users to leave OpenSea and view the item on the site.
+   * A URL to a multi-media attachment for the item. The file extensions GLTF, GLB, WEBM, MP4, M4V, OGV,
+   * and OGG are supported, along with the audio-only extensions MP3, WAV, and OGA.
+   * Animation_url also supports HTML pages, allowing you to build rich experiences and interactive NFTs using JavaScript canvas,
+   * WebGL, and more. Scripts and relative paths within the HTML page are now supported. However, access to browser extensions is not supported.
    */
```
cesarenaldi, line 100: I need to polish this copy, could lead to odd audio post if we keep it as it is.

packages/domain/src/use-cases/publications/CreateComment.ts
```diff
@@ -14,70 +14,120 @@ import {
...
+   * These are not the same as #hashtag in the publication content. Use these if you don't want to clutter the publication content with tags.
    */
-  content?: string;
+  tags?: string[];
+};
+
+export type CreateTextualCommentRequest = BaseCommentRequest & {
   /**
-   * Main content focus for this publication
+   * The publication focus is the to text only
```
krzysu, line 76:
```suggestion
   * The publication focus is text only
```

krzysu (approved):
just one comment, as CreatePost and CreateComment requests are very much the same (outside of 2 fields, `kind` and `publicationId`) would it make sense to share the same type? then you don't need to duplicate all comments/docs for each field.

cesarenaldi:
> just one comment, as CreatePost and CreateComment requests are very much the same (outside of 2 fields, `kind` and `publicationId`) would it make sense to share the same type? then you don't need to duplicate all comments/docs for each field.

That is a good point, I am mostly doing it due to our current TypeDoc setup, documentation wise it would produce an annoying to navigate typedef.

So keeping it separate for now. I am developing on the side a TypeDoc plugin which should be useful in this context. It allows to expand a type to it's resolved union by adding a `@expand` custom tag. A that point we can have common type without affecting the readability of the generated doc.

desfero (approved):
I do like it, it's now clear from the type system that you have different publication types.

PR #337, opened 2023-05-15
Draft: Gated client changes

paulxuca:


packages/react/src/config.ts
```diff
@@ -64,6 +65,7 @@ export type LensConfig = {
 export type EncryptionConfig = {
   authentication: AuthenticationConfig;
   provider: IEncryptionProvider;
+  createGatedClient?: typeof createGatedClient;
```
cesarenaldi, line 68: Can you please add `@internal` TSDoc? For now it's not meant for public use (we reserve the right to remove it/change it without notice without breaking semver).

PR #338, opened 2023-05-16
feat: paginated hooks now have option to fetch items before results set

cesarenaldi:


packages/api-bindings/src/apollo/cache/createApolloCache.ts
```diff
@@ -63,6 +63,9 @@ function createTypePolicies({ activeWalletVar }: TypePoliciesArgs): StrictTypedT
 
     RevenueAggregate: createRevenueAggregateTypePolicy(),
 
+    // ensures that no matter what fields we add to it in the future, it will NOT be normalized
+    PaginatedResultInfo: notNormalizedType(),
```
desfero, line 67: 👍

packages/api-bindings/src/apollo/cache/utils/cursorBasedPagination.ts
```diff
@@ -27,34 +23,71 @@ export function cursorBasedPagination<T extends CursorBasedPagination>(
...
+      if (variables.cursor === existing.pageInfo.prev) {
+        return {
+          ...incoming,
+          items: incomingItems.concat(existingItems),
+          pageInfo: {
+            ...incoming.pageInfo, // future-proofing in case we add more fields to pageInfo
+            beforeCount: incoming.pageInfo.beforeCount,
+            moreAfter: existing.pageInfo.next !== null,
+            next: existing.pageInfo.next,
+            prev: incoming.pageInfo.prev ?? existing.pageInfo.prev,
```
desfero, line 71: Could it end up with duplicated items in the list? 
Let's assume the following flow
1. `prev` is called, 2 items returned and appended, `incoming.pageInfo.prev` is undefined so we reuse existing prev
2. X minutes later we have 2 new publications 
3. `prev` is called, 4 items returned (two we already had, two new), now you end up with duplicated items in the `items` list.

Is that a possible scenario? Maybe I'm missing some part that's doing deduplication

packages/api-bindings/src/apollo/cache/utils/cursorBasedPagination.ts
```diff
@@ -27,34 +23,71 @@ export function cursorBasedPagination<T extends CursorBasedPagination>(
...
+            ...incoming.pageInfo, // future-proofing in case we add more fields to pageInfo
+            beforeCount: incoming.pageInfo.beforeCount,
+            moreAfter: incoming.pageInfo.next !== null,
+            next: incoming.pageInfo.next ?? existing.pageInfo.next,
+            prev: existing.pageInfo.prev,
+          },
+        };
+      }
+
+      throw new InvariantError('Unable to merge incoming cursor-based pagination result');
```
desfero, line 90: nit, we have also a `never` helper that we use across the mono repo. Not strong as both are ok

desfero (changes requested):

packages/api-bindings/src/apollo/cache/utils/cursorBasedPagination.ts
cesarenaldi: I see the possible confusion, and at first I thought after reading your scenario I thought it was possible. Then remembered.

The thing is that if `items` is populated then `prev` and `next` are also populated. `prev` and `next` are null only when `items: []`.

I totally get this is not obvious from the code above. It relies on an implicit behaviour of the API. I think we can make it more obvious.

packages/api-bindings/src/apollo/cache/utils/cursorBasedPagination.ts
```diff
@@ -27,34 +23,71 @@ export function cursorBasedPagination<T extends CursorBasedPagination>(
...
-        items: existingItems.concat(incomingItems),
-        pageInfo: incoming.pageInfo,
-      } as SafeReadonly<T>;
+      if (variables.cursor === existing.pageInfo.prev) {
+        return {
+          ...incoming,
+          items: incomingItems.concat(existingItems),
+          pageInfo: {
+            ...incoming.pageInfo, // future-proofing in case we add more fields to pageInfo
+            beforeCount: incoming.pageInfo.beforeCount,
```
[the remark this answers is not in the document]
cesarenaldi: In an previous iteration of this PR it was called `moreBefore`. Then I started to entertain the idea of building a functionality that would (via periodic polling) check for new items and propagate this information to the UI so that "Show X newer tweets" feature could be implemented. This is how `moreBefore` evolved into `beforeCount`. I then decided to descope the observing logic into [T-11159](https://avara.height.app/T-11159) cause I had more urgent tickets to focus on.

The reason i left most of the `beforeCount` (even if stays to `0` in this PR) is that there are some test cases that will become useful for that follow-up ticket. I just didn't want to delete them knowing I would pick up that ticket at some point next week.

packages/api-bindings/src/apollo/cache/utils/cursorBasedPagination.ts
```diff
@@ -27,34 +23,71 @@ export function cursorBasedPagination<T extends CursorBasedPagination>(
 
       // items that are not in the cache anymore (for .e.g deleted publication)
       const danglingItems = items.filter((item) => !canRead(item));
-      const readRes = {
+      const readRes: SafeReadonly<TResult> = {
         ...existing,
         items,
         pageInfo: {
           ...pageInfo,
+          beforeCount: existing.pageInfo.beforeCount ?? 0,
```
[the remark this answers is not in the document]
cesarenaldi: `read` and `merge` are not called always in the same order. Very first render is different from subsequent updates. Not sure why Apollo Client has this not symmetric behaviour, I discovered this empirically. One of the tests cover this.

packages/api-bindings/src/apollo/cache/utils/cursorBasedPagination.ts
cesarenaldi: I checked `never` throws a simple `Error`.

Did you mean `invariant` helper?

packages/react/src/helpers/reads.ts
```diff
@@ -145,5 +151,15 @@ export function usePaginatedReadResult<
         });
       }
     },
+
+    prev: async () => {
+      if (data?.result.pageInfo.prev) {
+        await fetchMore({
+          variables: {
+            cursor: data?.result.pageInfo.prev,
```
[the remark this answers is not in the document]
cesarenaldi: good catch, I'll check.

packages/api-bindings/src/apollo/cache/utils/cursorBasedPagination.ts
desfero: got it, now i remember too that we had to call the api one last time to get empty cursor and empty array.
Would that be possible to make some branching logic that clearly indicates that behaviour?

packages/api-bindings/src/apollo/cache/utils/cursorBasedPagination.ts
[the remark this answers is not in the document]
desfero: got it, makes sense

packages/api-bindings/src/apollo/cache/utils/cursorBasedPagination.ts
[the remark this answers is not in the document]
desfero: ah, interesting.

packages/api-bindings/src/apollo/cache/utils/cursorBasedPagination.ts
desfero: `invariant` would require you to pass some value and seeing `invariant(false, "...")` IMO is confusing.
But looking at the `never` implementation i think it should return `InvariantError` as i looked quickly through the all usages of the helper and it's always used as an invariant that should never happen.

packages/api-bindings/src/apollo/cache/utils/cursorBasedPagination.ts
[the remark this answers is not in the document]
desfero: love that we would expose `beforeCount`, makes the UI a lot better on the consumer side.

desfero (approved):

Issue #252, opened 2023-03-24
Expose Access Token in Lens SDK for server-side verification of user


wingedrhino:
**Is your feature request related to a problem? Please describe.**

The lens SDK does not expose the access token directly. However, the
[verify-jwt](https://docs.lens.xyz/docs/verify-jwt) endpoint requires it.
This seems to be the only way for a lens protocol implementer's API
server to verify that the user has logged in through lens.

**Describe the solution you'd like**

Add a `getAccessToken(): Promise<string>` method to `IAuthentication`.

Re-use the code currently in `Authentication.getRequestHeader` and pull it
out into a publicly exposed method on `IAuthentication`.

**Describe alternatives you've considered**

Adding `getRequestHeader` to the public interface. But this seems a little too
use-case specific. You don't always need the access token for adding to your
request headers.

**Additional context**

Current workaround from a package that uses the official SDK:

```typescript
async function getCredentials (lens: LensClient): Promise<string> {
const lensAuth = lens.authentication as Authentication
const getHeadersResp = await lensAuth.getRequestHeader()
if (getHeadersResp.isFailure()) {
  return ''
}
const headers = getHeadersResp.unwrap()
const authHeaderValue = headers.authorization
if (authHeaderValue.length < 7) {
  return ''
}
const accessToken = authHeaderValue.slice(7)
return accessToken
}
```

krzysu:
would a simple implementation like this be enough or do we also need to include "the refresh" flow?

```
async getAccessToken(): Promise<string | undefined> {
  const credentials = await this.storage.get();

  return credentials?.accessToken;
}
```

wingedrhino:
> would a simple implementation like this be enough or do we also need to include "the refresh" flow?

By "the refresh flow", do you mean the ability to automatically check if the token is expired and then fetch a fresh one from the server? Yep, in a high level SDK that is VERY useful.

I cannot think of many usage scenarios where I would want an old, expired lens access token :smile:

cesarenaldi:
@wingedrhino this seems a pretty good first issue. Would you like to submit a PR?

wingedrhino:
@cesarenaldi I can take this up over the weekend!

ashutosh887:
Let me work on this now @cesarenaldi Sir

wingedrhino:
I got a little caught up. Wasn't in the best of health.

[...]

PR #349, opened 2023-05-22
feat: add verify and getAccessToken methods

krzysu:
solves https://github.com/lens-protocol/lens-sdk/issues/252

### DIFF
--- a/packages/client/src/authentication/Authentication.ts
+++ b/packages/client/src/authentication/Authentication.ts
@@ -1,4 +1,10 @@
-import { EthereumAddress, failure, PromiseResult, success } from '@lens-protocol/shared-kernel';
+import {
+  EthereumAddress,
+  failure,
+  invariant,
+  PromiseResult,
+  success,
+} from '@lens-protocol/shared-kernel';
 import { InMemoryStorageProvider } from '@lens-protocol/storage';
 
 import { LensConfig } from '../consts/config';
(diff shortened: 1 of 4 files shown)


packages/client/src/authentication/Authentication.ts
```diff
@@ -81,6 +102,33 @@ export class Authentication implements IAuthentication {
...
+      const newCredentials = await this.api.refresh(credentials.refreshToken);
+      await this.storage.set(newCredentials);
+
+      invariant(newCredentials.accessToken, 'Access token should be defined');
+      return success(newCredentials.accessToken);
+    }
+
+    // otherwise
+    return failure(new CredentialsExpiredError());
+  }
```
cesarenaldi, line 130: Nice!

packages/client/src/authentication/Authentication.ts
```diff
@@ -81,6 +102,33 @@ export class Authentication implements IAuthentication {
...
+
+    if (!credentials.shouldRefresh() && credentials.accessToken) {
+      return success(credentials.accessToken);
+    }
+
+    if (credentials.canRefresh()) {
+      const newCredentials = await this.api.refresh(credentials.refreshToken);
+      await this.storage.set(newCredentials);
+
+      invariant(newCredentials.accessToken, 'Access token should be defined');
```
cesarenaldi, line 124: It's well possible the refresh token does not get renewed. I would return a
```
return failure(new CredentialsExpiredError())
```
in case of failed refresh, cause it implies the refresh token is expired as well.

krzysu: done!

cesarenaldi (approved):

PR #350, opened 2023-05-22
feat: `usePollDetails` and `usePollVote` hooks

cesarenaldi:
I isolated this refactoring so to remove noise from the rest of the PR. In this PR:
- segregated codegen config under a single generated entry so that I will be able to add Snapshot GQL config next to it
- regenerate the files as per changes above
- move `api-bindings/src/graphql/*` into ``api-bindings/src/lens/*` with the idea to add `api-bindings/src/snapshot` in a following PR


reecejohnson (approved):

reecejohnson:
nice, thanks for this. makes the following PR much easier to look through and provides a good separation of concerns

PR #351, opened 2023-05-22
feat: inbox domain

krzysu:


packages/domain/src/use-cases/inbox/GetConversation.ts
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    const messages = await this.messageGateway.fetchMessagesFor(conversation);
+
+    const data = this._assembleConversationData(conversation, messages);
+
+    this.presenter.presentConversation(data);
+  }
+
+  private _assembleConversationData(
```
cesarenaldi, line 61: nit: do we need the `_` given we have language support for visibility modifier?

packages/domain/src/entities/Conversation.ts
```diff
@@ -0,0 +1,27 @@
+import { EthereumAddress } from '@lens-protocol/shared-kernel';
+
+import { ProfileId } from './Profile';
+
+export type ConversationId = string;
+export type Markdown = string;
+export type Emoji = string;
```
cesarenaldi, line 7: Shall we brand these now? There is a `Brand<T, symbol>` helper for that

packages/domain/src/use-cases/inbox/GetConversation.ts
krzysu: you are right, done

PR #352, opened 2023-05-22
feat: handle proxy action errors

reecejohnson:
Proxy action limit reach on follow and collect, revert back to self funded signed flow using `useSelfFundedFallback`

https://github.com/lens-protocol/lens-sdk/assets/34844044/11ecd458-6e26-43e6-a547-38c4066aa6ef


packages/react/src/transactions/adapters/CollectPublicationCallGateway.ts
```diff
@@ -4,37 +4,70 @@ import {
...
+} from '@lens-protocol/shared-kernel';
 
 import { UnsignedProtocolCall } from '../../wallet/adapters/ConcreteWallet';
+import { ITransactionFactory } from './ITransactionFactory';
+import { ProxyReceipt } from './ProxyReceipt';
 import { Data, SelfFundedProtocolTransactionRequest } from './SelfFundedProtocolTransactionRequest';
 
-export class CollectPublicationCallGateway implements IOnChainProtocolCallGateway<CollectRequest> {
-  constructor(private apolloClient: LensApolloClient) {}
+export class CollectPublicationCallGateway
```
cesarenaldi, line 36: Naming wise I would probably not retain the word "call" as it's mean to signify "protocol call" entity.

Maybe just `CollectPublicationGateway`?

packages/react/src/transactions/adapters/FollowProfilesCallGateway.ts
```diff
@@ -47,13 +67,91 @@ function resolveProfileFollow(request: FollowRequest): Follow[] {
   return [{ profile: request.profileId }];
 }
 
-export class FollowProfilesCallGateway implements IOnChainProtocolCallGateway<FollowRequest> {
-  constructor(private apolloClient: LensApolloClient) {}
+export class FollowProfilesCallGateway
```
cesarenaldi, line 70: Same here re: "call". Maybe just `FollowProfilesGateway`?

packages/react/src/transactions/adapters/__tests__/CollectPublicationCallGateway.spec.ts
```diff
@@ -81,4 +133,107 @@ describe(`Given an instance of the ${CollectPublicationCallGateway.name}`, () =>
       expect(unsignedCall.nonce).toEqual(nonce);
     });
   });
+
+  describe(`when calling the "${CollectPublicationCallGateway.prototype.createProxyTransaction.name}" method`, () => {
+    describe('when relaying a FreeCollectRequest', () => {
```
cesarenaldi, line 138: These tests read odds: "when ... when" in a behaviour focused tests style, also the extra "and" makes it difficult to distinguish what is the event (the when) and what are the preconditions (the givens):

I think the `when relaying a FreeCollectRequest` already implies calling the `createProxyTransaction`, it's just expressed in more domain language. I don't have a preference between one or the other as far as now that we merged the 2 gateways together we keep the wording consistent with the other method tests.

packages/react/src/transactions/adapters/__tests__/CollectPublicationCallGateway.spec.ts
```diff
@@ -81,4 +133,107 @@ describe(`Given an instance of the ${CollectPublicationCallGateway.name}`, () =>
...
+            }),
+          );
+
+          // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-explicit-any
+          expect((transaction as any).state.proxyId).toEqual(indexingId);
+        });
+      });
+      describe('and receiving an error response', () => {
+        const request = mockFreeCollectRequest();
+        it(`should resolve with ${ProxyTransaction.name} on Polygon`, async () => {
```
cesarenaldi, line 190: This wording here does not reflect the actual expectations in the test code.

Accounting for the previous comments let's try to rewrite these so that we have less nesting (seems not needed in this case) and the whole test reads like a specification for this unit:

```
when calling the "createProxyTransaction" method
  [it] should succeed with a "ProxyTransaction" on Polygon
  [it] should fail with a "BroadcastError" in case of broadcast failure
```

another way would be to shift the "when" into a precondition (part of the "givens") and nest one more time:
```
and a "FreeCollectRequest" via the "createProxyTransaction" method
  when the broadcasting succeeds
    [it] should succeed with a "ProxyTransaction" on Polygon
  when the broadcasting fails
    [it] should fail with a "BroadcastError" in case of broadcast failure
```

"succeed" and "fail" aims at suggesting this it returning `Success<ProxyTransaction>` and and `Failure<BroadcastError>` (I took some shortcuts in writing these types to get to the point). We use these in other places. The intent it's to make a distinction between succeeds/fails vs. resolves/rejects vs. returns/throws. Not sure if very intelligible, but we use this elsewhere.

packages/react/src/transactions/adapters/__tests__/FollowProfilesCallGateway.spec.ts
```diff
@@ -153,14 +201,111 @@ describe(`Given an instance of the ${FollowProfilesCallGateway.name}`, () => {
...
+            }),
+          );
+
+          // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-explicit-any
+          expect((transaction as any).state.proxyId).toEqual(indexingId);
+        });
+      });
+
+      describe('and receiving an error response', () => {
+        it.only('should return broadcast error with fallback', async () => {
```
cesarenaldi, line 264: `.only` + similar considerations done in collect gateway regarding tests organization/wording.

packages/domain/src/use-cases/transactions/SignlessSubsidizeOnChain.ts
```diff
@@ -24,7 +25,11 @@ export class SignlessSubsidizeOnChain<T extends ProtocolTransactionRequestModel>
   async execute(request: T) {
     const transaction = await this.relayer.createProxyTransaction(request);
 
-    await this.transactionQueue.push(transaction);
+    if (transaction.isFailure()) {
+      this.presenter.present(failure(transaction.error));
+      return;
+    }
```
cesarenaldi, line 31: Neat!

cesarenaldi (changes requested):
Overall flow LGTM!

cesarenaldi (changes requested):
Overall flow LGTM!

packages/react/src/transactions/adapters/__tests__/CollectPublicationCallGateway.spec.ts
cesarenaldi: (please don't mind by broken english... take my wording as an example but improve it if needs to)

reecejohnson: okay nice, thanks for detailed response here, amended 👍

packages/domain/src/use-cases/transactions/SignlessSubsidizeOnChain.ts
```diff
@@ -1,15 +1,16 @@
...
+import { BroadcastingError } from './BroadcastingError';
 import { IGenericResultPresenter } from './IGenericResultPresenter';
 import { TransactionQueue } from './TransactionQueue';
 
 export interface ISignlessSubsidizedCallRelayer<T extends ProtocolTransactionRequestModel> {
-  createProxyTransaction(request: T): Promise<ProxyTransaction<T>>;
+  createProxyTransaction(request: T): PromiseResult<ProxyTransaction<T>, BroadcastingError>;
 }
 
 export type ISignlessSubsidizeOnChainPresenter = IGenericResultPresenter<void, Error>;
```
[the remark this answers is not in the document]
reecejohnson: sure, `BroadcastError` seems to be the only possible error right now

examples/web-wagmi/src/profiles/UseFollowAndUnfollow.tsx
```diff
@@ -23,11 +26,35 @@ function FollowButton({ followee, follower }: FollowButtonProps) {
...
+        'We cannot cover the transaction costs at this time. Do you want to retry with your own MATIC?',
+      );
+
+      setHasRetriedWithSelfFunded(true);
+
+      if (retry) {
+        await selfFundedFollow(attempt.error.fallback);
+      }
+    }
+  };
```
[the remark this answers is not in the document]
reecejohnson: sorted 👍

examples/web-wagmi/src/profiles/UseFollowAndUnfollow.tsx
```diff
@@ -59,7 +63,7 @@ type UseFollowInnerProps = {
 };
 
 function UseFollowInner({ activeProfile }: UseFollowInnerProps) {
-  const { data, error, loading } = useExploreProfiles({ limit: 50 });
+  const { data, error, loading } = useExploreProfiles({ limit: 50, observerId: activeProfile.id });
```
cesarenaldi, line 66: `observerId` is not longer needed to be explicit if you want it to be about the Active Profile. Did you find issues with this working as expected?

examples/web-wagmi/src/publications/UseCollect.tsx
```diff
@@ -22,9 +25,33 @@ type CollectButtonProps = {
...
+    }
+    if (supportsSelfFundedFallback(attempt.error)) {
+      const retry = window.confirm(
+        'We cannot cover the transaction costs at this time. Do you want to retry with your own MATIC?',
+      );
+
+      setHasRetriedWithSelfFunded(true);
+
+      if (retry) {
+        await selfFundedCollect(attempt.error.fallback);
```
cesarenaldi, line 50: Do you think we can move this to use the self-contained hook approach?

cesarenaldi (commented):
Solution and tests looks good to me. Only few comments on the example.

examples/web-wagmi/src/profiles/UseFollowAndUnfollow.tsx
reecejohnson: okay great, thanks. no issues - this was leftover from prior implementation

examples/web-wagmi/src/publications/UseCollect.tsx
reecejohnson: of course, done 👍

cesarenaldi (approved):

PR #361, opened 2023-05-26
feat: inbox - `enableConversations` hook with simplified `useConversations` hook

krzysu:
2 hooks to make sure it all work together

- enableConversations hook with react-web WebConversationProvider using xmtp-js 
- disable conversation on wallet logout
- useConversations simple hook (to be extended later) to test the usage flow


examples/web-wagmi/src/authentication/AuthenticationPage.tsx
cesarenaldi, line 1: FYI I repurposed this page in this branch https://github.com/lens-protocol/lens-sdk/pull/350/files#diff-3d1c2c515d05863d2db642732a4cc5840f0fdbc8711837681b8ce058b5a9d91f

packages/domain/src/use-cases/inbox/EnableConversations.ts
```diff
@@ -0,0 +1,48 @@
...
+    private readonly presenter: IEnableConversationsPresenter,
+  ) {}
+
+  async execute(request: EnableConversationsRequest): Promise<void> {
+    const wallet = await this.activeWallet.requireActiveWallet();
+
+    invariant(
+      wallet.address === request.address,
+      'Active wallet address does not match request address',
+    );
```
cesarenaldi, line 43: I think we can omit the `address` from the `request`.

packages/domain/src/use-cases/inbox/GetAllConversations.ts
```diff
@@ -0,0 +1,31 @@
+import { EthereumAddress, Result } from '@lens-protocol/shared-kernel';
+
+import { ProfileId } from '../../entities';
+import { Conversation, ConversationsDisabledError } from '../../entities/Conversation';
+
+export type GetAllConversationsRequest = {
+  profileId?: ProfileId;
+  address: EthereumAddress;
```
cesarenaldi, line 8: Do you think we still need `address` here with this new flow? Should it use the Active Wallet?

To understand this I am trying to entertain the idea of what happens if the user logs out (with `useWalletLogin`) and logs back in with another Signer/Wallet. 🤔

packages/react-web/src/inbox/KeyBundleStorage.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export type KeyBundleData = z.infer<typeof KeyBundleData>;
+
+export function createKeyBundleStorage(
+  storageProvider: IStorageProvider,
+  namespace: string,
+  address: EthereumAddress,
+) {
+  const notificationStorageDataSchema = new BaseStorageSchema(
+    `lens.${namespace}.${address}.inbox.keyBundle`,
```
cesarenaldi, line 15: ah... interesting the use of `address`, I can see a narrative with this, wondering if you could spend 5 minutes to explain to me the rationale you had in mind (also in relation to the comment I left on the use case interactors).

packages/react-web/src/inbox/SignerAdapter.ts
```diff
@@ -0,0 +1,20 @@
+import { WalletEntity } from '@lens-protocol/react';
+
+interface ISigner {
+  getAddress(): Promise<string>;
+  signMessage(message: ArrayLike<number> | string): Promise<string>;
+}
+
+export class SignerAdapter implements ISigner {
```
cesarenaldi, line 8: Shouldn't this implement the XMTP interface rather than a local interface?

packages/react-web/src/inbox/WebConversationProvider.ts
```diff
@@ -0,0 +1,144 @@
...
+import { extractPeerProfileId } from './helpers';
+
+function getXmtpEnvironment(name: EnvironmentConfig['name']) {
+  switch (name) {
+    case 'production':
+      return 'production';
+    case 'development':
+      return 'dev';
+  }
+}
```
cesarenaldi, line 27: I know you got inspired by token-gated content for this.... but it was a sloppy shortcut I took... better do proper dependency inversion of the config and feed it from the "above". Just last week I had to change token-gated to receive config from the main `environment` config cause it was getting in the way of some integration tests we are conducting. 🤷

packages/react-web/src/inbox/helpers.spec.ts
cesarenaldi, line 1: Move under `src/inbox/__tests__` folder.

packages/react-web/src/inbox/KeyBundleStorage.ts
cesarenaldi, line 1: In general can we move files that belongs more strictly to the adapter layer(s) into ` src/inbox/adapters` (like for the other features) and leave here maybe just the outer facing hooks.

packages/react-web/src/inbox/WebConversationProvider.ts
```diff
@@ -0,0 +1,144 @@
...
+export class WebConversationProvider implements IConversationProvider {
+  private client?: Client;
+
+  constructor(
+    private readonly environment: EnvironmentConfig['name'],
+    private readonly storage: IStorage<string>,
+  ) {}
+
+  private storeKeys(keys: Uint8Array) {
+    void this.storage.set(Uint8Array.from(keys).toString());
```
cesarenaldi, line 38: is the `void` intentional? why not wait for the write to finish?

packages/react/src/experimental/useEnvironmentConfig.ts
```diff
@@ -0,0 +1,15 @@
+import type { EnvironmentConfig } from '../environments';
+import { useSharedDependencies } from '../shared';
+
+/**
+ * Returns the internal {@link EnvironmentConfig}.
+ 
+ * @experimental This hook is experimental and may change in the future.
```
cesarenaldi, line 7: I would even flag it `@internal` for now.

packages/react/src/inbox/adapters/GetAllConversationsPresenter.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export class GetAllConversationsPresenter implements IGetConversationsPresenter {
+  private result: GetAllConversationsResult | null = null;
+
+  presentConversations(r: GetAllConversationsResult) {
+    this.result = r;
+  }
+
+  asResult(): GetAllConversationsResult {
+    invariant(this.result, 'Result should be present');
```
cesarenaldi, line 15: This invariant surface the fact that the caller of the use case using this hook, must know when the use case is finished before calling this method.

If you notice elsewhere this is use more as a "I call it but don't know when it ends... it will end at some point".

packages/react/src/inbox/useConversations.ts
```diff
@@ -0,0 +1,77 @@
...
+  config: InboxConfig;
+};
+
+/**
+ * @internal
+ * @experimental
+ */
+export function useConversations(
+  args: UseConversationsArgs,
+): ReadResult<Conversation[], ConversationsDisabledError> {
```
cesarenaldi, line 25: What happened to the Conversation DTO we used to have a previous iteration?

Better not expose the entities directly... they might evolve into object with behaviour attached to them and here we just need data.

The fact they are the same today... at the beginning of building a new feature... should be considered an apparent duplication. These types changes at different speed for different reason. Better not couple them even if it saves few keystrokes today.

We can leverage TS structural typing and for now just define a new type that has same properties of the entities.

cesarenaldi (changes requested):

packages/react-web/src/inbox/KeyBundleStorage.ts
krzysu: I copied this approach from how xmtp stores keys, as a key bundle is created with an address (signer), then I think it is important to store a key per address, if a user changes wallet, then we can retrieve the correct key from the storage for the current signer, the same with saving, to avoid overwriting the previous key.

packages/react-web/src/inbox/SignerAdapter.ts
krzysu: true, I thought they don't export it, but I was wrong

packages/domain/src/use-cases/inbox/GetAllConversations.ts
krzysu: in some other hooks, we always require activeProfile (i.e. in useFollow as `follower`) but we could also just read it internally in the hook.
so I just followed the same pattern. when the Signer changes, then the address provided to `useConversations` hook will also change as it is taken from `useActiveWallet` hook. 
at least that's what happens in the example in the web-wagmi app.

packages/domain/src/use-cases/inbox/DisableConversations.ts
```diff
@@ -0,0 +1,11 @@
+export interface IDisableConversationsGateway {
+  disableConversations(): Promise<void>;
+}
+
+export class DisableConversations {
+  constructor(private readonly gateway: IDisableConversationsGateway) {}
+
+  async execute(): Promise<void> {
+    await this.gateway.disableConversations();
```
cesarenaldi, line 9: It seems the separate use case could be error prone leaving the consumer app in a dangerous state from user's privacy perspective.

I would simply include this call in the existing `WalletLogout` use case and remove the need for extra hook, documentation, example, etc.

packages/domain/src/use-cases/inbox/GetAllConversations.ts
cesarenaldi: But Active Wallet and Active Profile are significantly different things.

Active Wallet determine the EOA used to login. Active Profile is one among several Lens Profile that can be owned by the Active Wallet EOA.

We do not bind ourself to the user's wallet like many crypto-oriented dApps do. This is on purpose so that after initial signing the social app based on the Lens SDK feels like a web2 experience. User can switch the network and/or account on their wallet app and there should not be consequence on their social app based  on the Lens SDK.

Only when a signature is needed the SDK requires a network switch and verifies the EOA address to match the Active Wallet address.

This is a design decision.

So the idea of Signer changing dynamically is not really a problem for the SDK. For the SDK the Active Wallet changes as consequence of logout or login with another wallet.

packages/react-web/src/inbox/KeyBundleStorage.ts
cesarenaldi: I am not sure if we should leave XMTP credentials laying around after login. I think this ties up with the idea of clearing the connection from `WalletLogout` and this https://github.com/lens-protocol/lens-sdk/pull/361#discussion_r1222654977 convesation.

packages/react/src/inbox/adapters/GetAllConversationsPresenter.ts
krzysu: what can I do about that? if I don't include `invariant` then TS complains that `this.result` could be `null`

packages/domain/src/use-cases/inbox/DisableConversations.ts
krzysu: I agree, and I started implementing this, the problem is that conversationGateway requires `IConversationProvider` which we create on the fly in each `use<*>Conversation` hook, but we don't have it in `useWalletLogout` 😟

packages/domain/src/use-cases/inbox/EnableConversations.ts
krzysu: done

packages/domain/src/use-cases/inbox/GetAllConversations.ts
krzysu: ok, done

packages/react/src/inbox/adapters/GetAllConversationsPresenter.ts
krzysu: replaced with `PromiseResultPresenter`

krzysu:
closing in favor of simpler integration with xmtp-react-sdk lib, see https://github.com/lens-protocol/lens-sdk/pull/401

PR #362, opened 2023-05-26
Gated Client changes for react native environment

paulxuca:


packages/react/src/publication/useEncryptedPublication.ts
```diff
@@ -75,7 +75,9 @@ export function useEncryptedPublication<T extends ContentPublication>({
         );
       }
 
-      const client = createGatedClient({
+      const createGatedClientFn = encryption.createGatedClient || createGatedClient;
```
desfero, line 78: wondering if we could make the `encryption.createGatedClient` default to `createGatedClient` to avoid branching logic here?

packages/gated-content/src/GatedClient.ts
```diff
@@ -72,6 +85,7 @@ export class GatedClient {
     this.signer = signer;
     this.storage = createAuthStorage(storageProvider, environment.name);
     this.encryptionProvider = encryptionProvider;
+    this.litClient = new NodeClient({ debug: false });
```
desfero, line 88: wondering if we could allow it to be passed from the constructor as looking into the code on the app side 
```
 constructor(config: GatedClientConfig) {
        super(config);

        this.litClient = litClient;
      }
```
we create `NodeClient` just to replace it on the next line with out custom react-native implmentation.

What if we go with:
```
constructor({
    authentication,
    environment,
    signer,
    storageProvider,
    encryptionProvider,
    litClient = new NodeClient({debug: false}),
  }: GatedClientConfig) {
```
@cesarenaldi @paulxuca  wdyt?

desfero (changes requested):
2 questions

packages/gated-content/src/GatedClient.ts
cesarenaldi: Let's try to wrap this up for now, compared to using the `IEncryptionProvider` and the cloned `GatedClient` solution this is already an order of magnitude better IMO.

desfero (approved):

desfero (approved):

krzysu:
build is green, do we need a changeset for this?

PR #369, opened 2023-05-30
feat: update wagmi version range

reecejohnson:


packages/wagmi/package.json
```diff
@@ -45,12 +45,12 @@
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "typescript": "^4.9.5",
-    "wagmi": "0.12.7"
+    "wagmi": "0.12.7 || <=0.12.13"
```
desfero, line 48: could we consistency to keep versions in `devDependency` in a format without using the `or`?
```
"wagmi": "0.12.7 || <=0.12.13"
```

packages/wagmi/package.json
```diff
@@ -45,12 +45,12 @@
...
   },
   "peerDependencies": {
     "@lens-protocol/react-web": "workspace:*",
     "ethers": "^5.7.2",
-    "wagmi": "0.12.7"
+    "wagmi": "0.12.7 || <=0.12.13"
```
desfero, line 53:
```
    "wagmi": "0.12.7 - 0.12.13"
```
```
    version1 - version2 Same as >=version1 <=version2.
```

PR #371, opened 2023-06-01
Present wallet and profile at the same time during login flow.

desfero:
This PR shuffles the logic a little bit to present the wallet and profile at the same time otherwise ui state in inconsistent and implementation of the claiming flow is trickier/buggy.
Also `useWalletLogin` now returns `WalletLoginResult` with the wallet and profile to allow imperative redirect call when someone needs it.

### DIFF
--- a/packages/react/src/wallet/useWalletLogin.ts
+++ b/packages/react/src/wallet/useWalletLogin.ts
@@ -3,13 +3,14 @@ import {
   UserRejectedError,
   WalletConnectionError,
 } from '@lens-protocol/domain/entities';
+import { WalletLoginResult } from '@lens-protocol/domain/use-cases/wallets';
 import { Signer } from 'ethers';
 
 import { Operation, useOperation } from '../helpers/operations';
 import { useWalletLoginController } from './adapters/useWalletLoginController';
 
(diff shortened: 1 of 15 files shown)


packages/react/src/wallet/useWalletLogin.ts
```diff
@@ -3,13 +3,14 @@ import {
...
 } from '@lens-protocol/domain/entities';
+import { WalletLoginResult } from '@lens-protocol/domain/use-cases/wallets';
 import { Signer } from 'ethers';
 
 import { Operation, useOperation } from '../helpers/operations';
 import { useWalletLoginController } from './adapters/useWalletLoginController';
 
 export type WalletLoginOperation = Operation<
-  void,
+  WalletLoginResult,
```
desfero, line 13: This one changes the return type of user-facing hook, do I need to update any docs?

cesarenaldi: Reference documentation can be regenerated once this is released on a stable version. Manual documentation wise we could start adding some details to [this guide](https://docs.lens.xyz/docs/hooks-authentication) once this is merged and released under `next` tag.

cesarenaldi (approved):

PR #372, opened 2023-06-01
feat: adds sandbox env config

cesarenaldi:


packages/react/src/environments.ts
```diff
@@ -75,6 +75,32 @@ export const development: EnvironmentConfig = {
   handleResolver: (handle) => `${handle}.test`,
 };
 
+/**
+ * The sandbox environment configuration
+ *
+ * This is the environment to be used when you develop and you need to experiment with custom collect/follow modules.
```
desfero, line 81: Do the lens-sdk support custom collect/follow modules as i quickly found inside `createCollectRequest` the invariant that's throw for unknown collect modules

desfero (approved):

desfero:
And seems changeset is missing

packages/react/src/environments.ts
cesarenaldi: It does support returning unknown collect and follow settings data on publications/profiles.... but not configuration of them.

desfero: got it, so there is no way latter to collect that publication through lens-sdk? Should go though some custom logic, right?

desfero: Maybe we could think for lens-sdk v2 about some way of collecting these sort of posts as could be a blocker for someone to use lens-sdk?

cesarenaldi: We could definitely enable more of these. Right now creating custom follow/collect module is not a very common occurrence and requires some coordination in terms of review, audit, whitelisting them. So it's not something happening very often.

PR #380, opened 2023-06-08
feat: upgrade wagmi bindings to wagmi v1

reecejohnson:
- update `wagmi` package exposed bindings to use wagmi v1 and viem 
- maintains Lens SDK interfaces as they were 
- provides adapters for viem/wagmi users to interop with our ethers requirements


packages/wagmi/src/index.ts
```diff
@@ -23,11 +28,18 @@ export function bindings(): IBindings {
...
+      return ethersProviderFromPublicClient(publicClient).getSigner();
     },
   };
 }
+
+export function ethersProviderFromPublicClient(publicClient: PublicClient): providers.Web3Provider {
+  invariant(window.ethereum, 'window.ethereum is not defined');
+
+  return new providers.Web3Provider(
+    window.ethereum as providers.ExternalProvider,
```
krzysu, line 42: is this actually using anything from `viem`? is it safe to depend on `window.ethereum` or viem is setting that up?

krzysu: I also see sth like this in the discussion here https://github.com/wagmi-dev/wagmi/discussions/2340#discussioncomment-5935358
(instead of using `window.ethereum`)

```
const { connector } = useAccount()

// get the connected provider instead of window.ethereum
connector.getProvider().then((provider) => {
  return new Web3Provider(provider).getSigner()
})
```

examples/web-wagmi/src/misc/LoginSpecificProfile.tsx
```diff
@@ -56,16 +58,15 @@ function AuthenticatedContent({
...
       await disconnectAsync();
     }
 
     const { connector } = await connectAsync();
 
     if (connector instanceof InjectedConnector) {
-      const signer = await connector.getSigner();
-      await login(signer, handle);
+      const publicClient = getPublicClient();
+      await login(ethersProviderFromPublicClient(publicClient).getSigner(), handle);
```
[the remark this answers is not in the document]
reecejohnson: we still use this check to determine whether there is a connector or not, as it can be `undefined`

although we only support the `InjectedConnector` in this example app so maybe we could just check it is defined

packages/wagmi/src/index.ts
reecejohnson: update this to use `walletClient.transport` which removes dependency on `window.ethereum` 👍

packages/wagmi/src/index.ts
```diff
@@ -23,11 +27,31 @@ export function bindings(): IBindings {
...
-      invariant(signer, 'Cannot get signer, is the wallet connected?');
+      invariant(walletClient, 'Wallet client not found');
 
-      return signer;
+      return signerFromWalletClient({ walletClient, chainId });
     },
   };
 }
+
+export function providerFromPublicClient({
```
reecejohnson, line 39: how does this look @krzysu @cesarenaldi ?

packages/wagmi/src/index.ts
```diff
@@ -1,12 +1,16 @@
-import { IBindings, RequiredSigner } from '@lens-protocol/react-web';
+import { IBindings } from '@lens-protocol/react-web';
 import { invariant } from '@lens-protocol/shared-kernel';
 import { providers } from 'ethers';
-import { SwitchChainNotSupportedError } from 'wagmi';
-import { fetchSigner, getNetwork, getProvider, switchNetwork } from 'wagmi/actions';
+import { PublicClient, SwitchChainNotSupportedError, WalletClient } from 'wagmi';
+import { getNetwork, getPublicClient, getWalletClient, switchNetwork } from 'wagmi/actions';
+import 'wagmi/window';
```
krzysu, line 6: what is this import for? won't it break in react-native or on server-side?

packages/wagmi/src/index.ts
```diff
@@ -23,11 +27,31 @@ export function bindings(): IBindings {
...
     },
   };
 }
+
+export function providerFromPublicClient({
+  publicClient,
+  chainId,
+}: {
+  publicClient: PublicClient;
+  chainId?: number;
```
desfero, line 44: nit:, you could get the `chainId` from the `publicClient.chain`, same with `signerFromWalletClient`

packages/wagmi/package.json
```diff
@@ -45,15 +45,18 @@
...
   "peerDependencies": {
     "@lens-protocol/react-web": "workspace:*",
     "ethers": "^5.7.2",
-    "wagmi": "0.12.7 - 0.12.13"
+    "viem": "^1.0.0",
+    "wagmi": "^1.1.0"
   },
   "prettier": "@lens-protocol/prettier-config",
   "dependencies": {
+    "@ethersproject/abstract-signer": "^5.7.0",
```
desfero, line 59: we probably don't need it as as of now the whole `ethers` is a `peerDependency` anyway

.changeset/wet-mice-impress.md
```diff
@@ -0,0 +1,5 @@
+---
+"@lens-protocol/wagmi": major
+---
+
+upgrade wagmi bindings to wagmi v1
```
desfero, line 5:
```suggestion
**Upgraded** wagmi to v1
```

desfero: just for consistency as i see we recently got out of sync with changeset namings

.changeset/wet-mice-impress.md
```diff
@@ -0,0 +1,5 @@
+---
+"@lens-protocol/wagmi": major
```
desfero, line 2: @cesarenaldi Are the versions of `react` and `wagmi` still linked? Given it's a major change it would upgrade the number to `2.x.x` and don't know if we want to also upgrade `react`/`react-web` to `2.x.x`.

desfero (changes requested):
Small nitpicks, good work 🌿

packages/wagmi/src/index.ts
```diff
@@ -23,11 +27,31 @@ export function bindings(): IBindings {
...
+
+export async function signerFromWalletClient({
+  walletClient,
+  chainId,
+}: {
+  walletClient: WalletClient;
+  chainId?: number;
+}): Promise<providers.JsonRpcSigner> {
+  return new providers.Web3Provider(walletClient.transport, chainId).getSigner();
+}
```
cesarenaldi, line 57: What is the mindset in exporting these 2 functions `signerFromWalletClient` and `providerFromPublicClient`?

packages/wagmi/src/index.ts
reecejohnson: good spot, this was for previous implementation when using `window.etherum` - it is now not needed 

i didn't get any feedback it wasn't used in my editor because it's not a var 🫣

will remove, thanks!

packages/wagmi/src/index.ts
reecejohnson: tried this and it sometimes came back as undefined?

packages/wagmi/package.json
reecejohnson: cool, thanks

packages/wagmi/src/index.ts
cesarenaldi: Yeah, I've noticed...  not sure if it's fully Lens SDK responsibility. The example app uses `WebBundlr` almost for convenience, but by design not tied to any specific upload strategy.

Is it possible to rewrite the uploader so to use wagmi/view directly?

PR #384, opened 2023-06-14
fix: remove an existing reaction before adding a new one

krzysu:


examples/web-wagmi/src/publications/UseReaction.tsx
```diff
@@ -48,7 +50,7 @@ function ReactionButton({ publication, profileId, reactionType }: ReactionButton
 
   return (
     <>
-      <button onClick={toggleReaction} disabled={isPending}>
+      <button onClick={toggleReaction} disabled={isPending || (hasAnyReaction && !hasReactionType)}>
```
krzysu, line 53: only single type of reaction is allowed from the API side, so disable possibility of adding a reaction if another one is already present - pure UI solution

packages/domain/src/use-cases/publications/Reaction.ts
```diff
@@ -23,24 +23,12 @@ export class Reaction {
   ) {}
 
   async add(request: ReactionRequest) {
+    await this.gateway.add(request);
     await this.presenter.add(request);
-
-    try {
-      await this.gateway.add(request);
-    } catch (e) {
-      await this.presenter.remove(request);
```
krzysu, line 31: the problem with this approach was, that when another type of reaction already exists, and we try to add a new one, then gateway call fails, and we don't know the previous state to revert to (could be `reaction:null` or `reaction:another type than what we tried to add`) 

so, as reactions are stored in API only, the request is fast, I decided we don't need optimistic update here, otherwise we need to store somewhere the previous value.

desfero: This could be tricky on slower networks, imagine someone on 3g who could visibly see the lag when pressed.
I live it for @cesarenaldi to decide if we want that to be fully optimistical or allow some sort of delay and force frontends to deal with `loading` state

krzysu: yes, good point, I think it's better to keep it as it was then, and just disable button in the UI, reactions need to be redone from the API side anyways

packages/react/src/publication/adapters/ReactionPresenter.ts
```diff
@@ -50,6 +55,33 @@ export class ReactionPresenter implements IReactionPresenter {
     });
   }
 
+  async revert(request: ReactionRequest): Promise<void> {
```
krzysu, line 58: @desfero what do you think about reverting a reaction to the previous state in case of error?

packages/react/src/publication/adapters/ReactionPresenter.ts
```diff
@@ -50,6 +55,33 @@ export class ReactionPresenter implements IReactionPresenter {
...
+    const statKey = getReactionStatKey(request.reactionType);
+
+    this.publicationCacheManager.update(request.publicationId, (current) => {
+      invariant(
+        isContentPublication(current),
+        `Reactions are not supported on ${current.__typename}`,
+      );
+
+      const removedStatKey = this.previousReactionType
```
desfero, line 67: would that be possible for `this.previousReactionType` to be not defined? 
Wondering if there is some way we could guard ourselves to catch broken wiring faster.
As right now if you revert back the changes in `useReactionControllrer` (aka remove `useMemo`) you won't get any error and the code would just work but at the same time be broken.

packages/react/src/publication/adapters/ReactionGateway.ts
```diff
@@ -12,10 +12,27 @@ import {
 import { ReactionRequest, IReactionGateway } from '@lens-protocol/domain/use-cases/publications';
 import { assertError } from '@lens-protocol/shared-kernel';
 
-export class ReactionGateway implements IReactionGateway {
+export type ExtendedReactionRequest = ReactionRequest & {
+  existingReactionType?: ReactionRequest['reactionType'];
```
desfero, line 16: would avoid using `?` as it's easy to miss that, if `undefined` is correct value let's state it clearly in union `ReactionRequest['reactionType'] | undefined`

desfero (approved):

PR #390, opened 2023-06-16
fix: ensure apollo type policies are type safe

krzysu:
to prevent similar issues like here https://github.com/lens-protocol/lens-sdk/pull/387/files in the future


krzysu:
the improvement would be to patch an `apollo-client-helpers` plugin to include correct field types when generating `StrictTypedTypePolicies`, I created a ticket for that: https://avara.height.app/lens-sdk/T-12187

closing this PR for now

PR #393, opened 2023-06-19
Require an active profile to have active wallet

desfero:
This is just a patch fix for now, created the ticket to track the proper change on the use case levels https://avara.height.app/T-12199.

### DIFF
--- a/packages/domain/src/use-cases/wallets/WalletLogin.ts
+++ b/packages/domain/src/use-cases/wallets/WalletLogin.ts
@@ -74,8 +74,8 @@ export class WalletLogin {
       request.handle,
     );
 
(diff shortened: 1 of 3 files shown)


packages/domain/src/use-cases/wallets/WalletLogin.ts
```diff
@@ -74,8 +74,8 @@ export class WalletLogin {
       request.handle,
     );
 
-    this.activeProfilePresenter.presentActiveProfile(profile);
```
desfero, line 77: This line here was causing some state inconsistencies inside `isOwnedByMe` which requires the wallet to be already present. This indicates also why it's important to keep that state together as it's strongly coupled together to some extend. See https://avara.height.app/T-12199

PR #397, opened 2023-06-22
add example of integration with xmtp react sdk

krzysu:
show how simple is the integration and where the actual lens bits are needed


examples/web-wagmi/src/inbox/helpers.ts
```diff
@@ -0,0 +1,40 @@
+import { ProfileId, profileId as brandProfileId } from '@lens-protocol/react-web';
+
+const CONVERSATION_ID_PREFIX = 'lens.dev/dm';
```
krzysu, line 3: this file is actually the Lens part of the integration with xmtp

PR #401, opened 2023-06-26
feat: simple lens<>xmtp web integration

krzysu:
- `useXmtpClient` hook
- `useStartLensConversation` hook
- `useEnhanceConversations` and `useEnhanceConversation` hook
- example inbox implementation using xmtp-react-sdk and new lens/react-web hooks


packages/react-web/src/inbox/adapters/SignerAdapter.ts
```diff
@@ -0,0 +1,39 @@
...
+
+    await useCase.execute(request);
+
+    const result = await presenter.asResult();
+
+    if (result.isSuccess()) {
+      return result.value;
+    } else {
+      throw result.error;
+    }
```
cesarenaldi, line 37:
```suggestion
    return result.unwrap();
```

packages/react-web/src/inbox/entities/Conversation.ts
cesarenaldi, line 1: I would not use the word entity here given its meaning. Maybe just a file `types.ts` instead of `entities` folder is enough in this case.

packages/domain/src/use-cases/inbox/SignArbitraryMessage.ts
```diff
@@ -0,0 +1,31 @@
...
+    private readonly activeWallet: ActiveWallet,
+    private readonly presenter: ISignArbitraryMessagePresenter,
+  ) {}
+
+  async execute(request: string): Promise<void> {
+    const wallet = await this.activeWallet.requireActiveWallet();
+    const result = await wallet.signMessage(request);
+    this.presenter.present(result);
+  }
+}
```
cesarenaldi, line 31: solid!

packages/react/src/experimental/useActiveWalletController.ts
```diff
@@ -0,0 +1,14 @@
+import { ActiveWallet } from '@lens-protocol/domain/use-cases/wallets';
+
+import { useSharedDependencies } from '../shared';
+
+/**
+ * Returns the internal active wallet controller.
```
cesarenaldi, line 6:
```suggestion
 * Returns the internal active wallet interactor.
```

packages/react/src/profile/useProfile.ts
```diff
@@ -20,7 +20,12 @@ export type UseProfileByHandleArgs = {
 };
 
 export type UseProfileArgs = WithObserverIdOverride<
-  XOR<UseProfileByIdArgs, UseProfileByHandleArgs>
+  XOR<
+    {
+      skip: true;
```
cesarenaldi, line 25: Do we need this? I think it's part of a wider conversation we need to have on how and if we need to have conditionally triggerable queries (either via a `skip` of by a laxy-query approach).

But I do like your take using `XOR`, it does address my reservations on type safety as one cannot provide data until it's skipped.

cesarenaldi: In this PR I would still use conditional rendering to render downstream hooks. Then we can open a separate conversation about this `skip` approach, which I reckon could be done across the board to all read hooks.

packages/react-web/src/inbox/useXmtpClient.ts
```diff
@@ -0,0 +1,93 @@
...
+const defaultOptions: InitXmtpClientOptions = {
+  persistConversations: true,
+};
+
+export function useXmtpClient(): UseXmtpClientResult {
+  const { client, disconnect, isLoading: clientIsLoading, initialize } = useClient();
+  const [isLoading, setIsLoading] = useState(false);
+  const [error, setError] = useState<UseXmtpClientResult['error']>();
+
+  const activeWaletController = useActiveWalletController();
```
cesarenaldi, line 42:
```suggestion
  const activeWalletController = useActiveWalletController();
```

packages/react-web/src/inbox/useXmtpClient.ts
```diff
@@ -0,0 +1,93 @@
...
+          options,
+        });
+
+        const newKeys = await Client.getKeys(signer);
+        await storeKeys(storage, newKeys);
+
+        return newClient;
+      } catch (e) {
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore Error not yet properly typed
```
cesarenaldi, line 75: you can use `assertError`

packages/react-web/src/inbox/useXmtpClient.ts
```diff
@@ -0,0 +1,93 @@
...
+        setError(e);
+
+        return undefined;
```
cesarenaldi, line 78: Is this `undefined`  expected from a XMTP `useClient` signature perspective?

krzysu: yes, this is the original signature, I only modified the type of `error`

```
declare const useClient: () => {
    client: _xmtp_xmtp_js.Client | undefined;
    disconnect: () => void;
    error: unknown;
    initialize: (arg0: InitClientArgs) => Promise<_xmtp_xmtp_js.Client | undefined>;
    isLoading: boolean;
    signer: _xmtp_xmtp_js.Signer | undefined;
};
```

packages/react/src/profile/useProfile.ts
krzysu: the problem is that I'm fetching xmtp conversation(s), extracting `profileIds` from the conversations (if lens convos), and then fetching full profiles with `useProfile(s)` hooks. It is possible that no conversation is a lens conversation, which means I'm calling `useProfile(s)` without any profileId, which returns an error from the lens API. I would like to not trigger `useProfile(s)` hooks if no profileIds found in conversations. any ideas how to do it?

examples/web-wagmi/src/inbox/UseConversations.tsx
```diff
@@ -0,0 +1,81 @@
...
+    conversations: tempConversations,
+    error: tempError,
+    isLoading: isLoadingTemp,
+  } = useConversations();
+
+  const {
+    data: conversations,
+    error,
+    loading,
+  } = useEnhanceConversations({ conversations: tempConversations, profile });
```
desfero, line 25: wondering if we could model the api in this way:
```
const {} = useEnhanceConversations(useConversations());
```
Than as a consumer you only have to care about single `error/loging/data` variables, what do you think?

examples/web-wagmi/src/inbox/components/ConversationComposer.tsx
```diff
@@ -0,0 +1,51 @@
...
+    event.preventDefault();
+
+    const form = event.currentTarget;
+
+    const formData = new FormData(form);
+    const content = (formData.get('content') as string | null) ?? never();
+
+    await startConversation(peerProfile.ownedBy, content);
+
+    if (!error) {
```
desfero, line 26: curious about the flow here and where does the `error` change the value. 
If i get correctly it's technically not possible for `error` to be non nullable here as even when `startConversation` will update the `error` it would happen at the next render cycle and this function would already finish the execution?

examples/web-wagmi/src/inbox/components/MessagesCard.tsx
```diff
@@ -0,0 +1,39 @@
+import { Conversation, DecodedMessage, useMessages, useStreamMessages } from '@xmtp/react-sdk';
+import { useCallback, useEffect, useState } from 'react';
+
+type MessagesCardProps = {
+  conversation: Conversation;
+};
+
+export function MessagesCard({ conversation }: MessagesCardProps) {
+  const { messages } = useMessages(conversation);
```
desfero, line 9: Curious why xmtp doesn't expose a hook that always returns the messages by subscribing to the steam to reduce the API surface for the consumers.

packages/react-web/package.json
```diff
@@ -67,6 +69,7 @@
     "typescript": "^4.9.5"
   },
   "peerDependencies": {
+    "@xmtp/react-sdk": "1.0.0-preview.39",
```
desfero, line 72: could be annoying as maybe you don't plan to use xmpt in your frontend but it would force you to anyway install it as a dependency in your project.

packages/react-web/src/inbox/helpers.ts
```diff
@@ -0,0 +1,58 @@
...
+
+export function notEmpty<TValue>(value: TValue | null | undefined): value is TValue {
+  return value !== null && value !== undefined;
+}
+
+/**
+ * Create a unique conversation ID based on sender/receiver addresses and
+ * context values
+ */
+export const createUniqueConversationId = (conversation?: Conversation): string =>
```
desfero, line 55: wondering if we could avoid calling at all that function when `conversation` is not present?

packages/react-web/src/inbox/helpers.ts
```diff
@@ -0,0 +1,58 @@
+import { ProfileId, profileId as brandProfileId } from '@lens-protocol/react';
+import { invariant } from '@lens-protocol/shared-kernel';
+import { Conversation } from '@xmtp/react-sdk';
+
+const CONVERSATION_ID_PREFIX = 'lens.dev/dm';
```
desfero, line 5: An open question more here, wondering if all projects would want to use the same conversation id 🤔

packages/react-web/src/inbox/helpers.ts
krzysu: I copied it from here without thinking too much 🤦 it shouldn't be optional, thx for noticing
https://github.com/xmtp/xmtp-web/blob/main/packages/react-sdk/src/helpers/getConversationId.ts

packages/react-web/package.json
krzysu: shall I make it optional then?

```
"peerDependenciesMeta": {
  "@xmtp/react-sdk": {
    "optional": true
  },
},
```

examples/web-wagmi/src/inbox/components/EnableConversationsButton.tsx
```diff
@@ -0,0 +1,27 @@
+import { useXmtpClient } from '@lens-protocol/react-web';
+import { useCallback } from 'react';
+
+import { ErrorMessage } from '../../components/error/ErrorMessage';
+import { Loading } from '../../components/loading/Loading';
+
+export function EnableConversationsButton() {
+  const { client, error, isLoading, initialize } = useXmtpClient();
+
+  const handleConnect = useCallback(async () => {
```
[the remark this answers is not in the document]
krzysu: in our example app the hook is re-rendered a few times, I was looking for ways to reduce it, but now I'm not sure if useCallback helped

examples/web-wagmi/src/inbox/UseConversations.tsx
krzysu: yes, this is a very good idea! I will try it.

packages/react-web/src/inbox/helpers.ts
krzysu: this is coming from our own docs https://docs.lens.xyz/docs/encrypted-dms and seems the standard for all lens apps integrating xmtp. but as a builder you can also display all xmtp conversations wallet-to-wallet (where with prefix they are considered lens profile-to-profile ones). That is what xmtp calls a "interoperable inbox". https://xmtp.org/docs/concepts/interoperable-inbox

examples/web-wagmi/src/inbox/components/ConversationComposer.tsx
krzysu: here is xmtp implementation, error is set when calling the startConversation function https://github.com/xmtp/xmtp-web/blob/main/packages/react-sdk/src/hooks/useStartConversation.ts

but yes, you are right, the handling of error in submit here doesn't make sense

desfero: i see, but there is just no chance that `error` gonna be set in the same task queue so probably can be removed

examples/web-wagmi/src/inbox/components/EnableConversationsButton.tsx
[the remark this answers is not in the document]
desfero: i'm not sure it would help in any way given button doesn't care if it's stable reference or not

packages/react-web/package.json
desfero: optional could work, let's try.

packages/react-web/src/inbox/helpers.ts
desfero: got it, make sense

desfero (approved):

packages/react-web/src/__helpers__/jest.setup.ts
```diff
@@ -0,0 +1,18 @@
...
+
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
+global.Blob = Blob;
+
+// until https://github.com/jsdom/jsdom/issues/2524 is resolved
+global.TextEncoder = TextEncoder;
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
+global.TextDecoder = TextDecoder;
```
cesarenaldi, line 18: This content is familiar... is this needed by XMTP SDK?

packages/react-web/src/inbox/adapters/SignerAdapter.ts
```diff
@@ -0,0 +1,35 @@
...
+import { SignArbitraryMessage } from '@lens-protocol/domain/use-cases/inbox';
+import { ActiveWallet } from '@lens-protocol/domain/use-cases/wallets';
+import { PromiseResultPresenter } from '@lens-protocol/react';
+import { Signer } from '@xmtp/react-sdk';
+
+export class SignerAdapter implements Signer {
+  constructor(private activeWallet: ActiveWallet) {}
+
+  async getAddress() {
+    const wallet = await this.activeWallet.requireActiveWallet();
```
cesarenaldi, line 16: I would leave a comment here that highlight this being a shortcut and we would not normally have use cases returning data synchronously with their return statement but we would have decoupled the delivery mechanism via a presenter interface.

packages/react-web/src/inbox/helpers/__tests__/conversationIdUtils.spec.ts
```diff
@@ -0,0 +1,37 @@
+import {
+  buildConversationId,
+  extractPeerProfileId,
+  parseConversationId,
+} from '../conversationIdUtils';
+
+describe('Given a collection of inbox helper functions', () => {
+  describe('when calling buildConversationId', () => {
```
cesarenaldi, line 8:
```suggestion
  describe(`when calling ${buildConversationId.name}`, () => {
```

makes future renaming refactoring less error prone.

Other tests in this file might need this change.

packages/react-web/src/inbox/helpers/conversationIdUtils.ts
cesarenaldi, line 1: Both `parseConversationId` and `extractPeerProfileId` are implementation details but are exported to be tested. Should we test them via proper scenarios of the `extractPeerProfileId`. This makes future refactoring much easier as the tests are less couple with the implementation details and the tests can assist during refactoring instead of breaking.

packages/react-web/src/inbox/useEnhanceConversations.ts
```diff
@@ -0,0 +1,101 @@
...
+        ...acc,
+        [createUniqueConversationId(c)]: peerProfileId,
+      };
+    }, {});
+  }, [conversations, profile.id]);
+
+  const uniqueProfileIds: ProfileId[] = useMemo(() => {
+    const ids = Object.values(conversationToProfileIdMap).filter(notEmpty);
+    return [...new Set(ids)];
+  }, [conversationToProfileIdMap]);
```
cesarenaldi, line 40: Why not compute `conversationToProfileIdMap` and `uniqueProfileIds` in the same `useMemo` instead of chaining them?

packages/react-web/src/inbox/useEnhanceConversations.ts
```diff
@@ -0,0 +1,101 @@
...
+
+      return {
+        ...acc,
+        [createUniqueConversationId(c)]: peerProfileId,
+      };
+    }, {});
+  }, [conversations, profile.id]);
+
+  const uniqueProfileIds: ProfileId[] = useMemo(() => {
+    const ids = Object.values(conversationToProfileIdMap).filter(notEmpty);
```
cesarenaldi, line 38: I think here we can simply but computing `conversationToProfileIdMap: Record<string, ProfileId>` (i.e. not including `undefined` values.

packages/react-web/src/inbox/useEnhanceConversations.ts
```diff
@@ -0,0 +1,101 @@
...
+        }
+
+        // Clone the xmtp Conversation instance with all its methods and add peerProfile
+        // eslint-disable-next-line
+        return Object.assign(Object.create(Object.getPrototypeOf(c)), c, { peerProfile });
+      });
+
+      setEnhancedConversations(eConversations);
+    }
+  }, [profiles, conversationToProfileIdMap, conversations]);
```
cesarenaldi, line 69: I think this effect that ultimately leads to update the state could just be another `useMemo` and we can then remove the need for `useState`

```ts
const enhancedConversations = useMemo(() => {...}, [profiles, conversations])

```

packages/react-web/src/inbox/useStartLensConversation.ts
```diff
@@ -0,0 +1,21 @@
...
+import { buildConversationId } from './helpers';
+
+export type StartLensConversationRequest = {
+  ownedProfile: ProfileOwnedByMe;
+  peerProfile: Profile;
+};
+
+export type UseStartLensConversationResult = ReturnType<typeof useStartConversation>;
+
+export function useStartLensConversation({
```
cesarenaldi, line 13: Flag as `@experimental` including correlated types.

packages/react-web/src/inbox/useEnhanceConversations.ts
```diff
@@ -0,0 +1,101 @@
...
+import { useEffect, useMemo, useState } from 'react';
+
+import { extractPeerProfileId, createUniqueConversationId, notEmpty } from './helpers';
+import { EnhancedConversation } from './types';
+
+export type EnhanceConversationsRequest = {
+  profile: ProfileOwnedByMe;
+};
+
+export function useEnhanceConversations(
```
cesarenaldi, line 19: Flag as `@experimental` including correlated types.

packages/react-web/src/inbox/useXmtpClient.ts
```diff
@@ -0,0 +1,93 @@
...
+  error: PendingSigningRequestError | UserRejectedError | WalletConnectionError | Error | undefined;
+  initialize: (options?: InitXmtpClientOptions) => Promise<Client | undefined>;
+  isLoading: boolean;
+};
+
+const defaultOptions: InitXmtpClientOptions = {
+  persistConversations: true,
+};
+
+export function useXmtpClient(): UseXmtpClientResult {
```
cesarenaldi, line 38: Flag as `@experimental` including correlated types.

packages/react-web/src/inbox/useXmtpClient.ts
```diff
@@ -0,0 +1,93 @@
...
+async function storeKeys(storage: IStorage<string>, keys: Uint8Array) {
+  await storage.set(Uint8Array.from(keys).toString());
+}
+
+async function loadKeys(storage: IStorage<string>): Promise<Uint8Array | null> {
+  const val = await storage.get();
+  return val ? Uint8Array.from(val.split(',').map((c) => parseInt(c))) : null;
+}
+
+export type InitXmtpClientOptions = Partial<Omit<ClientOptions, 'env' | 'privateKeyOverride'>>;
```
cesarenaldi, line 24: Does `Partial<...>` here risk to mislead consumer to provide configuration that is from from XMTP client perspective?

packages/react-web/src/inbox/useEnhanceConversation.ts
```diff
@@ -0,0 +1,48 @@
...
+  useEffect(() => {
+    setEnhancedConversation((c) => {
+      if (peerProfile) {
+        // Clone the xmtp Conversation instance with all its methods and add peerProfile
+        // eslint-disable-next-line
+        return Object.assign(Object.create(Object.getPrototypeOf(c)), c, { peerProfile });
+      }
+      return c;
+    });
+  }, [peerProfile, conversation]);
```
cesarenaldi, line 35: Should we simplify with a `useMemo` instead of `useState` + `useEffect`?

```ts
const enhancedConversation = useMemo(() => { ... }, [conversation, peerProfile])
```

This way some computation like `extractPeerProfileId` (although very minimal) can be done only when needed.

packages/react-web/src/inbox/useEnhanceConversation.ts
```diff
@@ -0,0 +1,48 @@
...
+
+import { extractPeerProfileId } from './helpers';
+import { EnhancedConversation } from './types';
+
+export type EnhanceConversationRequest = {
+  profile: ProfileOwnedByMe;
+  conversation: Conversation;
+};
+
+export function useEnhanceConversation({
```
cesarenaldi, line 13: Flag as `@experimental`

packages/react/src/index.ts
```diff
@@ -123,6 +123,7 @@ export { UnspecifiedError } from '@lens-protocol/api-bindings';
 export { Amount, WellKnownSymbols, ether, matic, erc20, usd } from '@lens-protocol/shared-kernel';
 export * from './sources';
 export * from './utils';
+export { PromiseResultPresenter } from './transactions/adapters/PromiseResultPresenter';
```
cesarenaldi, line 126: Let's make sure all `PromiseResultPresenter.ts` exports are flagged as `@internal`

packages/react/src/profile/useProfile.ts
cesarenaldi: Seems a good opportunity to use `OneOf` (https://timhwang21.gitbook.io/index/programming/typescript/xor-type) generalization of `XOR`.

Can we flag the `{ skip: true }` option as `@experimental`?

packages/react/src/profile/useProfile.ts
```diff
@@ -37,16 +42,17 @@ export function useProfile({
...
+      ...useLensApolloClient(
         useActiveProfileAsDefaultObserver({
           variables: useSourcesFromConfig({
             request,
             observerId,
           }),
         }),
       ),
-    ),
+      skip: request.skip,
```
cesarenaldi, line 54: We should pass `skip` through `useActiveProfileAsDefaultObserver` or we risk to inhibit some of it's use cases.

```ts
      useLensApolloClient(
        useActiveProfileAsDefaultObserver({
          variables: useSourcesFromConfig({
            request,
            observerId,
          }),
          skip: request.skip,
        }),
      ),
```

packages/react/src/profile/useProfiles.ts
```diff
@@ -88,19 +90,21 @@ export function useProfiles({
...
-    useGetAllProfiles(
-      useLensApolloClient(
+    useGetAllProfiles({
+      ...useLensApolloClient(
         useActiveProfileAsDefaultObserver({
           variables: useSourcesFromConfig({ byHandles, byProfileIds, limit, observerId }),
         }),
       ),
-    ),
+      skip,
```
cesarenaldi, line 107: Same here, not bypassing `useActiveProfileAsDefaultObserver` but through it.

cesarenaldi (changes requested):

packages/react-web/src/__helpers__/jest.setup.ts
krzysu: I've just checked, and no, it's because in react-web tests I included import of `@lens-protocol/react` package and error comes somewhere from lit package. we have the same jest.setup in react package

Screenshot 2023-07-24 at 11 03 19

cesarenaldi: got it

examples/web-wagmi/src/inbox/components/MessagesCard.tsx
krzysu: I guess you don't always want to subscribe, sometimes just fetch messages and that's all

packages/react-web/src/inbox/useXmtpClient.ts
krzysu: this is the original type of xmtp options so I just omitted the two that we take care of

```ts
type InitClientArgs = {
    keys?: Uint8Array;
    options?: Partial<ClientOptions> & OnError;
    signer?: Signer | null;
};
```

packages/react-web/src/inbox/helpers/conversationIdUtils.ts
krzysu: I see, will do

packages/react-web/src/inbox/adapters/SignerAdapter.ts
krzysu: hm, but this is not a use-case, this is a Xmtp Signer adapter, in the method below `signMessage` there is a usecase with a presenter (using PromiseResultPresenter)

packages/domain/src/use-cases/wallets/__tests__/WalletLogout.spec.ts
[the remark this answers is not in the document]
krzysu: sorry, but the reference to the line got lost, do you remember what you meant here?

packages/react-web/src/inbox/useEnhanceConversation.ts
krzysu: done

packages/react-web/src/inbox/useEnhanceConversations.ts
krzysu: the thing is: from xmtp perspective, it's perfectly fine that a conversation has no profileId, so the mapping makes sense, and only inside computing `uniqueProfileIds` we don't are about conversations without profile, so it's easier to follow the logic this way

packages/react/src/index.ts
krzysu: I ended up removing the export of PromiseResultPresenter and implementing a version directly in react-web

packages/react/src/profile/useProfile.ts
```diff
@@ -38,18 +47,19 @@ export function useProfile({
   );
 
   const { data, error, loading } = useReadResult(
-    useGetProfile(
-      useLensApolloClient(
+    useGetProfile({
+      ...useLensApolloClient(
```
cesarenaldi, line 51: Object and spreading no longer needed.

packages/react/src/profile/useProfiles.ts
```diff
@@ -89,21 +91,23 @@ export function useProfiles({
...
   invariant(
     byHandles === undefined || byProfileIds === undefined,
     `Cannot provide both 'handles' and 'profileIds' to '${useProfiles.name}' hook`,
   );
 
   return usePaginatedReadResult(
-    useGetAllProfiles(
-      useLensApolloClient(
+    useGetAllProfiles({
+      ...useLensApolloClient(
```
cesarenaldi, line 103: Object and spreading no longer needed.

packages/react/src/experimental/useEnvironmentConfig.ts
```diff
@@ -0,0 +1,13 @@
+import type { EnvironmentConfig } from '../environments';
+import { useSharedDependencies } from '../shared';
+
+/**
+ * Returns the internal {@link EnvironmentConfig}.
+ *
+ * @internal
+ */
+export function useEnvironmentConfig(): EnvironmentConfig {
```
cesarenaldi, line 9: I this still needed, search on this PR does not lead to any usage? 🤔

packages/react/src/experimental/useActiveWalletController.ts
```diff
@@ -0,0 +1,14 @@
...
+
+import { useSharedDependencies } from '../shared';
+
+/**
+ * Returns the internal active wallet interactor.
+ *
+ * @internal
+ */
+export function useActiveWalletController(): ActiveWallet {
+  const { activeWallet } = useSharedDependencies();
```
cesarenaldi, line 11: Maybe not for now, but wondering if we should just export `useSharedDependencies` (as `@internal`).

krzysu: yeah, maybe not a bad idea

cesarenaldi (approved):

krzysu:
I've heard that XMTP folks are doing a total overhaul on sending messages with codecs and encoded content right now. And as their react SDK is not released as v1 yet, I will hold this PR from merging till XMTP is ready with a stable release of their sdk.

PR #405, opened 2023-06-29
feat: introduces experimental `useCurrentSession` hook

cesarenaldi:


packages/domain/src/use-cases/lifecycle/ISessionPresenter.ts
```diff
@@ -0,0 +1,42 @@
+import { ProfileId } from '../../entities';
+import type { LogoutReason } from '../wallets';
+
+/**
+ * The identifier of a Profile
+ */
+export type ProfileIdentifier = {
```
desfero, line 7: Any specific reason to not use `Profile` entity?

packages/domain/src/use-cases/lifecycle/ISessionPresenter.ts
```diff
@@ -0,0 +1,42 @@
...
+  id: ProfileId;
+  handle: string;
+};
+
+/**
+ * Framework-agnostic Wallet representation
+ */
+export type WalletData = {
+  address: string;
```
desfero, line 16: nit: could use `EthereumAddress` from `shared-kernel` which we could brand at some point in time

packages/api-bindings/src/apollo/cache/createProfileTypePolicy.ts
```diff
@@ -73,9 +71,9 @@ export function createProfileTypePolicy(
 
       followStatus: {
         read(_, { readField }) {
-          const activeWallet = activeWalletVar();
+          const session = getSession();
 
-          if (!activeWallet) {
+          if (!session || session.isNotAuthenticated()) {
```
desfero, line 76: what could be the use case where `getSession` return a null value? 
Should we make it an `invariant` as it should never happen if the app is properly wired together?

desfero (approved):

packages/api-bindings/src/apollo/cache/createProfileTypePolicy.ts
cesarenaldi: If everything is wired correctly requests requiring the session should never fire. Maybe it's a big assumption that all hooks (present and future) will be bug free from day-one.

These invariant are difficult to catch unless we increase our test coverage (not all read hooks do have an integration unit test). Also these field policies are triggered by profile/publication requests done in presenters. I would need to think about it if it's possible for those to fire before we know current session state.

packages/domain/src/use-cases/lifecycle/ISessionPresenter.ts
cesarenaldi: good call

packages/domain/src/use-cases/lifecycle/ISessionPresenter.ts
cesarenaldi: Passing a DTO to the presenter, not the entity directly. Leveraging TS structural typing to not create a new object until we need to.

PR #408, opened 2023-06-30
fix: login with handle

cesarenaldi:


examples/web-wagmi/src/authentication/LoginSpecificProfile.tsx
```diff
@@ -18,95 +10,31 @@ function UnauthenticatedContent() {
...
-        })}
-      </select>
-
-      <button onClick={onLoginClick}>
-        <strong>Log in</strong>
-      </button>
-    </>
-  );
-}
-
```
cesarenaldi, line 104: I simplified this example... when already logged-in, logging-in with another handle on the same wallet is not a use case. One should use `useActiveProfileSwitch` for that purpose.

packages/react/src/transactions/infrastructure/TransactionObserver.ts
```diff
@@ -33,7 +33,7 @@ const ONE_SECOND = 1000; // ms
  * @internal
  */
 export type TransactionObserverTimings = {
-  pollingInterval: number; // ms // TODO delete this
+  pollingInterval: number; // ms // TODO move this definition closer to EnvironmentConfig, not used by TransactionObserver
```
cesarenaldi, line 36: I was confused to had to investigate. Left my findings.

packages/react/src/wallet/useWalletLogin.ts
```diff
@@ -31,7 +64,7 @@ export function useWalletLogin(): WalletLoginOperation {
       if ('address' in walletLoginArgs) {
         return loginWallet({
           address: walletLoginArgs.address,
-          handle,
+          handle: walletLoginArgs.handle,
```
cesarenaldi, line 67: The fix.

reecejohnson: Ah, nice catch. Thanks

reecejohnson (approved):

PR #411, opened 2023-07-06
fix: split second double incrementing bug cache issue

reecejohnson:
There is an issue in the mirror and collect counters where we are managing state between the Apollo cache and internal Apollo reactive variables.

The problem was that we had a `network-only` call to get a fresh copy of the publication in the responders for both mirror and collect. This updated the Apollo cache with the incremented value a few MS before the pending txs were marked as settled. In the type policy we were combining the count in the publication stats for mirror/collect along with the count of pending txs still pending. This leaves us with a split second where there count is 1 more than it should be before the pending txs are marked as settled.

@desfero and I found two potential fixes:

1. Settle the txs before calling `commit` on the responders. Remove stat calculation from the type policy and handle all cache updates in the responders themselves. This removes any deviation from the correct total amount of mirrors/collects value. This is what is shown in this PR.
2. Use `queueMicrotask` in the responder to resolve the `commit` promise early and make the cache updates run so closely together that they cannot be seen as visual changes in the UI. We noted (on my machine) that there was a single MS diff between the two calls meaning that React did not render the first and incorrect update to the UI. This does not rule out the fact this may happen under other conditions.

I am leaving this PR open as a discussion point before we agree on the final fix.


packages/react/src/transactions/adapters/responders/CreateMirrorResponder.ts
```diff
@@ -1,25 +1,47 @@
...
+      id: identifier,
+      fields: {
+        stats(existing: PublicationStats) {
+          return {
+            ...existing,
+            totalAmountOfMirrors: existing.totalAmountOfMirrors + 1,
+          };
+        },
+      },
+    });
```
cesarenaldi, line 38: The challenge with these programmatic cache updates and the reason shy we started moving away from them and using more field policies is that they rely on the cache to be primed with the node we are aiming to update (a Publication in this case). This works in the immediate of a UI but does not work well in another thread tha tmight not have this exact item in cache (e.g. another tab) or during hard refreshes where the SDK bootstrap by being pure JS logic tries to update cache before any active query is able to fetch the relevant item.

cesarenaldi (approved):

PR #414, opened 2023-07-11
feat: expose publication types for usePublications query

krzysu:


examples/web-wagmi/src/publications/UsePublications.tsx
```diff
@@ -24,6 +40,28 @@ export function UsePublications() {
...
+          {allPublicationTypes.map((value) => (
+            <label key={value}>
+              <input
+                type="checkbox"
+                checked={publicationTypes.includes(value)}
+                name="publicationTypes"
+                value={value}
+                onChange={(e) => {
+                  if (e.target.checked) {
+                    setPublicationTypes([...publicationTypes, value]);
```
desfero, line 54: nit: could be a good use case for js `Set` given we can't have duplicates in the array (or at least we don't want :D)

packages/react/src/publication/usePublications.ts
```diff
@@ -15,6 +15,7 @@ export type UsePublicationsArgs = PaginatedArgs<
   WithObserverIdOverride<{
     metadataFilter?: PublicationMetadataFilters;
     profileId: ProfileId;
+    publicationTypes?: PublicationTypes | PublicationTypes[];
```
desfero, line 18: Wondering if we could force consumers to always provider an array?

desfero: Could you also review all field policies if they include `publicationTypes` as a cache key to have separate cache storages in case the query is called with different `publicationTypes` 🙏 ?

krzysu: regarding field policies, is this what you mean? https://github.com/lens-protocol/lens-sdk/blob/04647bbeb13fc18bcd030f4abfac6b8e15d82439/packages/api-bindings/src/apollo/cache/createPublicationsFieldPolicy.ts 
seems `publicationTypes` is already there

desfero: yes, that was exactly what I was not sure about and wanted to double check. Thanks.

examples/web-wagmi/src/publications/UsePublications.tsx
krzysu: do you mean sth like below? I don't think it's a case here as the value can only be set from a select box, and it can't be checked twice

```ts
setPublicationTypes(Array.from(new Set([...publicationTypes, value])));
```

PR #416, opened 2023-07-13
Fix `web-wagmi` build

desfero:


examples/nextjs/components/auth/LoginButton.tsx
```diff
@@ -1,39 +1,21 @@
 import { useWalletLogin, useWalletLogout } from '@lens-protocol/react-web';
 import { useEffect } from 'react';
 import toast from 'react-hot-toast';
-import { useAccount, useConnect, useDisconnect } from 'wagmi';
-import { InjectedConnector } from 'wagmi/connectors/injected';
```
desfero, line 5: Simplifying `next.js` example given it was anyway broken as was using the older `wagmi 0.x`. We anyway have wagmi integration inside `web-wagmi` so to simplify maintenance this is using now just in memory signer as the `react-native` example.

examples/web-wagmi/package.json
```diff
@@ -15,7 +16,7 @@
     "tsc": "tsc --noEmit"
   },
   "dependencies": {
-    "@bundlr-network/client": "^0.10.3",
+    "@bundlr-network/client": "^0.11.9",
```
desfero, line 19: Had to upgrade bundler due to some heresy in build pipline.
https://github.com/vitejs/vite/issues/9703
Older `@bundlr-network/client` was using `near-api-js` which is probably wrongly published and was causing build failures for production `web-wagmi` deployment.

examples/web-wagmi/vite.config.ts
```diff
@@ -11,19 +11,4 @@ export default defineConfig({
     nodePolyfills({ protocolImports: true }),
     visualizer({ open: true }) as PluginOption,
   ],
-
-  build: {
```
desfero, line 15: simplifying, for example app it's not needed

examples/nextjs/components/auth/LoginButton.tsx
reecejohnson: nice, thanks!

examples/nextjs/domain/wallet.ts
```diff
@@ -0,0 +1,21 @@
+// This is NOT PRODUCTION READY implementation of the EOA.
```
reecejohnson, line 1: good to show usage of the SDK without `wagmi` 👍

examples/web-wagmi/package.json
```diff
@@ -3,8 +3,9 @@
   "private": true,
   "version": "0.1.0",
   "scripts": {
+    "build:with-packages": "turbo run build --filter=...example-wagmi",
```
krzysu, line 6: what does `--filter=...example-wagmi` do?

krzysu (approved):
nice!

reecejohnson (approved):

examples/web-wagmi/package.json
desfero: Builds all dependencies of the `example-wagmi` inside monorepo, we could also build the whole monorepo but this one is better as only builds what `example-wagmi` is actually using.

examples/web-wagmi/vite.config.ts
```diff
@@ -11,19 +11,4 @@ export default defineConfig({
...
-
-    rollupOptions: {
-      treeshake: {
-        preset: 'recommended',
-      },
-    },
-  },
-
-  define: {
-    'process.env': {},
```
cesarenaldi, line 27: @desfero did we check metadata upload via WebBundlr works still? This line and the use of `nodePolyfills` plugin were needed to make `WebBundlr` works on browsers.

PR #420, opened 2023-07-14
feat: Profile Guardian support

cesarenaldi:


packages/client/src/profile/Profile.ts
```diff
@@ -916,4 +918,34 @@ export class Profile {
...
+   * Fetch profile guardian settings for the specified profile.
+   *
+   * ⚠️ Requires authenticated LensClient with the provided profileId.
+   *
+   * @param request - Request object for the mutation
+   * @returns {@link PromiseResult} with {@link ProfileGuardianResultFragment}
+   *
+   * @example
+   * ```ts
+   * await client.profile.profileGuardian({
```
krzysu, line 932: maybe let's call it just `guardian` to avoid repetition of "profile", so the call will look like

```
await client.profile.guardian({})
```

krzysu (approved):
don't forget about the changeset 😃 

LGTM!

packages/client/src/profile/Profile.ts
cesarenaldi: I like it

PR #424, opened 2023-07-17
feat: client sdk - support new media sets

krzysu:
similar as in react sdk, using config to setup defaults, but defaults are empty, so using API default values


packages/client/src/feed/Feed.ts
```diff
@@ -44,6 +50,7 @@ export class Feed {
   async fetch(
     request: FeedRequest,
     observerId?: string,
+    mediaTransformParams: MediaTransformParams = defaultMediaTransformParams,
```
desfero, line 53: @cesarenaldi Are we okay with just allowing you to have one transformed image?

krzysu: in react-sdk PR we decided to introduce separate image sets for Publication and two sets for Profile (picture and cover) 
see: https://github.com/lens-protocol/lens-sdk/pull/430/files#diff-b89498756162aec8b024c715e6744fceac12f4529c23d825c4baafba133e2f45

I wanted to keep it simple with client sdk, but we could do sth similar here too

krzysu: I rewrite the whole thing, could you take a look again?

packages/client/codegen-api.yml
```diff
@@ -24,12 +24,12 @@ config:
     FollowModuleData: string
     Handle: string
     HandleClaimIdScalar: string
-    IpfsCid: string
+    ImageSizeTransform: string
```
cesarenaldi, line 27: Can we copy over from react package the more type safe version of `ImageSizeTransform`?

cesarenaldi (commented):
Just one question

packages/client/codegen-api.yml
krzysu: ok, I added it for both api-bindings and client packages 

https://github.com/lens-protocol/lens-sdk/pull/424/commits/ebfa1011e39f64501eed074f2780792ed067d782

packages/client/src/revenue/Revenue.ts
```diff
@@ -115,6 +121,7 @@ export class Revenue {
         {
           request,
           observerId,
+          ...buildMediaTransformsFromConfig(this.config.mediaTransforms),
```
reecejohnson, line 124: I wonder if there's a nice way to abstract this enriching of the request into a function `...buildMediaTransformsFromConfig(this.config.mediaTransforms)` and keep this logic contained into one place?

reecejohnson: `enrichRequestWithMediaTransformations(...)` maybe?

reecejohnson (approved):

packages/client/src/revenue/Revenue.ts
reecejohnson: discussed with @krzysu, difficult to abstract any further due to media transformations not always being required

cesarenaldi (approved):
LGTM

Issue #419, opened 2023-07-13
Cache in Queries


jsonpreet:
**Describe the bug**
Is there a cache in queries, when I change the route or click on another link and go back, the Component is re-rendered but the hook is not called.

- [x] React Web App

**Version of the packages you are using**
1.2.1

cesarenaldi:
Hey @jsonpreet,

under the hood Lens SDK uses Apollo Client which has its own cache where the vast majority of the Lens UI state lives (also inside few Apollo Reactive Variables but these are transparently merged into cache items one way or the other).

What you are describing is an expected behaviour: changing routes should leverage cached content for a fast UX. 

Can you please clarify a little more about your scenario so we can understand the issue you are seeing and what you expect instead?

jsonpreet:
I am using react-router-dom for navigation and when I switch to another route the component is re-rendered but it won't request useFeed or any other lens-client/react-web hook

cesarenaldi:
If the hook arguments are not changed that is expected. Please advise if arguments are changing (e.g. a different profile ID).

This allows to accommodate the needs of apps that want to leverage the client cache for a snappy SPA experience. You will notice for example that if you come from a list of items (profiles or publications) and click through a page that shows the details of one of such items, the corresponding single item hook won't make another request but leverage the content of the cache.

One can fetch newer results on demand via the `prev` callback. You can see in [this example](https://github.com/lens-protocol/lens-sdk/blob/main/examples/web-wagmi/src/discovery/UseFeed.tsx#L63) used to create a "fetch newer" result feature (but can be called programmatically from a `useEffect` if need to).

jsonpreet:
That's not what I am saying, I mean When I click on Profile it loads profile data with publications and I click on Home or Go back from Browser on the home page it shows prev data, the hook is not called/re-rendered. And again if I click on the Same profile it also shows prev (cached) data. I have to hard reload the tab or window for new data.

jsonpreet:
Ok Thanks

cesarenaldi:
@jsonpreet as per linked PR this is now fixed and will be released soon. Will let you know when we have a preview release you can try.

cesarenaldi:
@jsonpreet to keep you in the loop on how we are solving this issue.

After evaluating different scenarios and usages of the Lens SDK we had to implement 2 different strategy.

[...]

PR #427, opened 2023-07-17
fix: read hooks not refetching data on remount cycle

cesarenaldi:
Addresses issue reported here: https://github.com/lens-protocol/lens-sdk/issues/419

### DIFF
--- a/examples/web-wagmi/src/App.tsx
+++ b/examples/web-wagmi/src/App.tsx
@@ -121,41 +121,31 @@ export function App() {
                   element={<LoginSpecificProfile />}
                 />
 
-                <Route path="/publications" element={<PublicationsPage />} />
-                <Route path="/publications/usePublication" element={<UsePublication />} />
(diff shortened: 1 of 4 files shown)


cesarenaldi:
After conversation with @desfero it turns out this approach does not work well with paginated query hooks.

Depending on the consumer use of the `next` callback and the time between the initial request and subsequent requests, the new dataset might not fit well within the cached items window causing a variety of unwanted side effects lik jumps in the UI, loss of restoring scroll position, wrong ordering of items.

We investigated and we think it would be better to limit this specific fetch policy handling to just single items query hooks (e.g. `usePublication`, `useProfile`, etc.) and instead for paginated query hooks incorporate the strategy of "probing for newer results" and give the consumer the option to know when new results are available and opt-in by calling `prev` callback (see other draft PR: https://github.com/lens-protocol/lens-sdk/pull/429).

cesarenaldi:
Solution integrate into https://github.com/lens-protocol/lens-sdk/pull/429

PR #429, opened 2023-07-18
feat: notifies presence of newer results

cesarenaldi:
This is a proof of concept to gather some consensus on the approach. If happy I'll make all the necessary changes to all read hooks.


examples/web-wagmi/src/publications/UsePublications.tsx
```diff
@@ -62,6 +64,12 @@ export function UsePublications() {
           ))}
         </fieldset>
 
+        {beforeCount > 0 && (
+          <button disabled={loading} onClick={prev}>
```
desfero, line 68: I'm not sure without running it locally and debugging, does `loading` flag switches to `true` when you call `prev`?

cesarenaldi: mmmhh.. good question... in my tests turns out it wasn't even doing the request... it was using the cache... will test

cesarenaldi: checked... `loading` is stable!

packages/react/src/helpers/reads.ts
```diff
@@ -132,18 +156,72 @@ type InferPaginatedItemsType<T extends PaginatedQueryData<unknown>> = T extends
...
+    async function (prev: Cursor) {
+      const response = await fetch({
+        variables: {
+          ...variables,
+          cursor: prev,
+        } as TVariables,
+      });
+
+      if (response.data) {
+        setBeforeCount(response.data.result.items.length);
```
krzysu, line 200: I don't get how `beforeCount` is set. I tested the new example in `UsePublications` and button never appeared for me when creating a new post from another app.
The `prev` call works as expected, fetching newer posts. but `beforeCount` never changes.

krzysu (approved):
in general looks good, left one question

packages/react/src/helpers/reads.ts
krzysu: ok, after reading the original issue thread (https://github.com/lens-protocol/lens-sdk/issues/419), all clear now, it's good to go 👍

PR #430, opened 2023-07-18
feat: support new media sets - optimized and transformed with size defined in config

krzysu:


packages/api-bindings/src/lens/common.graphql
```diff
@@ -204,11 +204,43 @@ fragment Media on Media {
...
+  }
+  optimized {
+    ...Media
+  }
+  thumbnail: transformed(params: $mediaTransformProfileThumbnail) {
+    ...Media
+  }
+}
+
+fragment ProfileCoverPicture on MediaSet {
```
krzysu, line 236: introduce new fragments for PublicationMediaSet, ProfilePicture and ProfileCoverPicture as they require different size transformations

packages/react/src/config.ts
```diff
@@ -57,6 +58,9 @@ export type LensConfig = {
    * @see {@link appId} helper
    */
   appId?: AppId;
+
+  /** TODO */
+  mediaTransforms?: MediaTransformsConfig;
```
krzysu, line 63: new config where user can define transformations per type of media they want to receive

packages/react/src/mediaTransforms.ts
```diff
@@ -0,0 +1,39 @@
...
+
+function buildMediaTransform(width: string, height = 'auto'): MediaTransform {
+  return {
+    width: imageSizeTransform(width),
+    height: imageSizeTransform(height),
+    keepAspectRatio: true,
+  };
+}
+
+export const defaultMediaTransformsConfig: MediaTransformsConfig = {
```
krzysu, line 23: if not provided by user, the default media transforms will be applied

packages/api-bindings/src/lens/utils/media.ts
```diff
@@ -0,0 +1,24 @@
+import { Brand } from '@lens-protocol/shared-kernel';
+
+export type ImageSizeTransform = Brand<string, 'ImageSizeTransform'>;
```
cesarenaldi, line 3: We could type this so that informs TS consumers of the valid options: 

```ts
type ImageSizeTransform = `${number}px` | `${number}%` | 'auto'
```

This is not super robust as `number` could be hex numbers, decimals, `1_000` notations, and many others.

A more robust and strict could be:

```ts
type Digit = '0' | '1' | '2' | '3' | '4' | '5' | '6' | '7' | '8' | '9'

type Percentage = `100%` | `${Digit}${Digit}%` | `${Digit}%`

type Pixel = `${Digit}${Digit}${Digit}${Digit}px` | `${Digit}${Digit}${Digit}px` | `${Digit}${Digit}px`

type ImageSizeTransform = Pixel | Percentage | 'auto'
```

cesarenaldi: I am making assumption that pixel values in the single digit or above 4 digits are not meaningful... probably safe for any reasonable application.

packages/api-bindings/src/lens/common.graphql
cesarenaldi: Naming wise there seems to be some inconsistencies in the use of the "set" suffix.

Maybe:

- `PublicationMediaSet` (unchanged) - I think `Media` correctly carry the fact it't not necessarily an image.
- `ProfilePictureSet` - `PictureSet` like if it were to be a specialized type of `MediaSet`, a "set of pictures"
- `ProfileCoverPictureSet` - same as above

packages/react/src/utils.ts
```diff
@@ -31,3 +32,13 @@ export function publicationId(id: string): PublicationId {
...
+
+/**
+ * Ensures that the given value is a valid ImageSizeTransform
+ *
+ * @group Helpers
+ */
+export function imageSizeTransform(value: string): ImageSizeTransform {
+  // for now just asserts the type, in future it will enforce a format
+  return value as ImageSizeTransform;
+}
```
cesarenaldi, line 44: With the template literal type... you won't need this to be a branded type anymore so the need for this helper kinda disappear.

krzysu:
@cesarenaldi unfortunately I still need to modify all affected tests, even with optional new mediaTransform params

examples/web-wagmi/src/publications/components/Medias.tsx
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  return null;
+}
+
+type MediaProps = {
+  publication: Post | Comment;
+};
+
+export function Medias({ publication }: MediaProps) {
```
cesarenaldi, line 24: nit: "media" is such an annoying word. I think it's already plural but in common english it has been used as singular in it's plural form. 🤷

packages/api-bindings/src/lens/utils/media.ts
```diff
@@ -0,0 +1,31 @@
...
+  | `${Digit}${Digit}${Digit}px`
+  | `${Digit}${Digit}px`;
+
+export type ImageSizeTransform = Pixel | Percentage | 'auto';
+
+export type MediaTransform = {
+  /**
+   * Set the transformed image's width. You can use specific size in
+   * pixels eg. 100px, a percentage eg. 50% or set as 'auto' to be set automatically.
+   * Default value is 'auto'.
```
cesarenaldi, line 16:
```suggestion
   *
   *  @defaultValue `auto`
```

packages/api-bindings/src/lens/utils/media.ts
```diff
@@ -0,0 +1,31 @@
...
+   */
+  width?: ImageSizeTransform;
+  /**
+   * Set the transformed image's height. You can use specific size in
+   * pixels eg. 100px, a percentage eg. 50% or set as 'auto' to be set automatically.
+   * Default value is 'auto'.
```
cesarenaldi, line 22:
```suggestion
   *
   *  @defaultValue `auto`
```

packages/api-bindings/src/lens/utils/media.ts
```diff
@@ -0,0 +1,31 @@
...
+   */
+  height?: ImageSizeTransform;
+  /**
+   * Set if you want to keep the image's original aspect ratio.
+   * True by default.
+   * If explicitly set to false, the image will stretch based on the width and height values.
+   */
```
cesarenaldi, line 29:
```suggestion
   * If explicitly set to false, the image will stretch based on the width and height values.
   *
   *  @defaultValue `true`
   */
```

examples/web-wagmi/src/publications/components/Medias.tsx
krzysu: right, thanks!

examples/web-wagmi/src/utils/media.ts
```diff
@@ -0,0 +1,65 @@
...
+
+export enum VideoType {
+  MP4 = 'video/mp4',
+}
+
+export enum AudioType {
+  WAV = 'audio/wav',
+  MP3 = 'audio/mpeg',
+  OGG = 'audio/ogg',
+}
```
cesarenaldi, line 28: These are exported by the `@lens-protocol/react` and `@lens-protocol/react-web` packages. Any reason to repeat these?

examples/web-wagmi/src/utils/media.ts
```diff
@@ -0,0 +1,65 @@
...
+
+export const FileType = { ...ImageType, ...VideoType, ...AudioType };
+export type FileType = ImageType | VideoType | AudioType;
+
+type Media = ContentPublication['metadata']['media'][number]['original'];
```
cesarenaldi, line 33: We should probably export `Media`

krzysu: it is exported, I didn't check, I copied it straight from our other project, now it is fixed.

examples/web-wagmi/src/utils/media.ts
krzysu: fixed

cesarenaldi (approved):

PR #433, opened 2023-07-18
fix: renames several mock helpers ahead of tests refactoring

cesarenaldi:


krzysu (approved):
you can merge this first, I will solve all merge issues with my PR after (https://github.com/lens-protocol/lens-sdk/pull/430) as it shouldn't be too hard as we touch different lines of code.

PR #437, opened 2023-07-20
Release 1.3-next.0

cesarenaldi:


packages/client/CHANGELOG.md
```diff
@@ -1,5 +1,18 @@
...
+### Minor Changes
+
+- c2a91ef4: **Added** `Profile.invitedBy` field
+- 4c4505d2: **Added** support for Profile Guardian
+
+### Patch Changes
+
+- 0f75f79d: **Added** experimental `useCurrentSession` hook
+  **Fixed** issue with `Profile` entity being leaked by the `useWalletLogin` hook
+  **Fixed** bug preventing logout on expired credentials detected at startup type
```
krzysu, line 14: what do we do with these entries? they are not related to client SDK, might be the result of changes in underlying packages, but I don't think so in this case

packages/wagmi/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@lens-protocol/wagmi",
-  "version": "2.0.1",
+  "version": "2.0.0-next.0",
```
krzysu, line 3: how did it go from 2.0.1 to 2.0.0-next?

desfero (approved):

krzysu (approved):

packages/client/CHANGELOG.md
cesarenaldi: Good catch, manually fixed.

packages/wagmi/package.json
cesarenaldi: My bad. Fixed.

PR #440, opened 2023-07-20
Fix `useUnreadNotificationsQuery` to work with fresh profile

desfero:


packages/react/src/notifications/useUnreadNotificationCount.ts
```diff
@@ -59,8 +59,10 @@ export function useUnreadNotificationCount({ profileId }: UseUnreadNotificationC
         return;
       }
 
+      invariant(data, 'Expected data to be defined');
+
       invariant(
-        data?.result.pageInfo.totalCount,
+        data.result.pageInfo.totalCount !== null,
```
desfero, line 65: was throwing `invariant` given the initial count is zero and zero is falsy value.

cesarenaldi (approved):

PR #441, opened 2023-07-20
feat: implements `useNotInterested` hook, adds`Profile.observedBy` and `Comment|Post.observedBy`

cesarenaldi:


packages/api-bindings/src/lens/common.graphql
```diff
@@ -390,6 +391,7 @@ fragment CommentBase on Comment {
   referencePolicy @client
   decryptionCriteria @client
   contentInsight @client
+  observedBy @client
```
krzysu, line 394: as I understand, you added this field mainly for us to be able to track who is the observer in a request, to detect data inconsistency, am I right?

krzysu (approved):

packages/api-bindings/src/lens/common.graphql
cesarenaldi: Correct, but cannot exclude usage from consumers perspective... they could use several queries with different observer ID and they might care about how a given publication was fetched.

PR #442, opened 2023-07-21
Use hexadecimal chain ID for switchChain

jarrodwatts:


packages/react/src/wallet/infrastructure/SignerFactory.ts
```diff
@@ -95,7 +95,7 @@ export class SignerFactory implements ISignerFactory {
     try {
       if (signer.provider && signer.provider instanceof JsonRpcProvider) {
         await signer.provider.send('wallet_switchEthereumChain', [
-          { chainId: chainConfig.chainId },
+          { chainId: chainConfig.hexadecimalChainId },
```
cesarenaldi, line 98: Thank you @jarrodwatts for fixing this.

I think we can simplify this by converting the number to hex string on the fly via:
```suggestion
          { chainId: utils.hexValue(chainConfig.chainId) },
```

This is a very low frequency operation that the extra overhead is negligible. The ethers `utils` is already used in this file.

In this case we can revert the config objects changes.

cesarenaldi (changes requested):
Thank @jarrodwatts. I tested you approach on wallet that do support the `chainId` as `number` and works well.

If you can address the comment I'll try to get this around into a `@next` release today.

cesarenaldi:
Tests failing due to lack of env variables for `LensClient` integration tests. Manually tested it locally, all good. Merging...

PR #444, opened 2023-07-21
feat: use CredentialsExpiryController and subscribe to expired event

krzysu:
I was able to trigger the token expiry event by modifying refreshToken stored in local storage and changing its `exp` value. Also setting `AccessTokenStorage.getAccessToken` to always return `''`, to trigger the token refresh.
The logout and storage cleanup proceeded as expected.
Seems everything works fine now!

Screenshot 2023-07-21 at 15 51 27

### DIFF
--- a/packages/react/src/shared.tsx
+++ b/packages/react/src/shared.tsx
@@ -10,7 +10,11 @@ import {
   TransactionQueue,
   TransactionResponders,
 } from '@lens-protocol/domain/use-cases/transactions';
-import { ActiveWallet, TokenAvailability } from '@lens-protocol/domain/use-cases/wallets';
+import {
+  ActiveWallet,
+  TokenAvailability,
+  WalletLogout,
+} from '@lens-protocol/domain/use-cases/wallets';
 import { ILogger, invariant } from '@lens-protocol/shared-kernel';
 import { IStorage, IStorageProvider } from '@lens-protocol/storage';
 import React, { ReactNode, useContext } from 'react';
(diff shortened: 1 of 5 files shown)


packages/react/src/shared.tsx
```diff
@@ -70,24 +78,23 @@ export type Handlers = {
 };
 
 export type SharedDependencies = {
-  appId?: AppId;
   activeProfileGateway: ActiveProfileGateway;
   activeWallet: ActiveWallet;
   apolloClient: SafeApolloClient;
+  appId?: AppId;
   bindings: IBindings;
-  authApi: AuthApi;
```
krzysu, line 78: removed as it was not used anywhere

packages/react/src/shared.tsx
```diff
@@ -218,9 +247,8 @@ export function createSharedDependencies(
     transactionFactory,
     transactionGateway,
     transactionQueue,
-    walletFactory,
-    walletGateway,
-    providerFactory,
+    walletLogin,
+    walletLogout,
```
krzysu, line 251: I needed `walletLogout` in shared to init `CredentialsExpiryController` so I decided to do the same with `walletLogout`, then we don't need to expose `walletFactory` nor `walletGateway`.
I dunno if that's fine but I just thought to try to clean up it a bit. WDYT?

packages/react/src/wallet/adapters/WalletGateway.ts
```diff
@@ -41,7 +41,7 @@ export class WalletGateway
   }
 
   async reset(): Promise<void> {
-    await this.storage.set([]);
+    await this.storage.reset();
```
krzysu, line 44: if we want to cleanup properly on logout, then better to reset it than leave with empty array

cesarenaldi (approved):

PR #450, opened 2023-07-26
fix: use current session when previous session was a session with profile

krzysu:


packages/react/src/lifecycle/__tests__/useCurrentSession.spec.ts
```diff
@@ -110,20 +114,58 @@ describe(`Given the ${useCurrentSession.name} hook`, () => {
...
+    });
+
+    it('should return the new profile data', async () => {
+      const { result } = setupUseCurrentSession({ profile });
+
+      act(() => {
+        updateSession(authenticatedProfile(wallet, profileIdentifier));
+      });
+
+      expect(result.current.loading).toBe(true); // loading while fetching the new profile
```
cesarenaldi, line 157: I think it should stay `false`. Maybe the test setup does not fully simulate this. We should load with a profile and then switch. This because `data` is never primed with the `prevProfile` data.

Despite the test wording this is akin to an existing test.

krzysu: I improved the test to load prevProfile and only then switch session to a new profile, that loading flag is still set to true as we can't return prevSession, neither new session (as they are both SessionType.WithProfile), so the only allowed state is loading: true

If we don't want that, then we need to figure out a different logic

krzysu: but I guess when switching profiles, then it's fine to have loading state in between

cesarenaldi: It shouldn't change. From the point of view of its lifecycle the loading phase is finished.... data can still change independently.

Ultimately we are promoting for `useCurrentSession` to be used as discriminant between logged-out/in and logged-in with and without a profile. We risk consumer's code to unmount entire chunks of app routing.

PR #458, opened 2023-07-27
feat: adds missing `appId` to `Comment` and `Post`

cesarenaldi:


packages/api-bindings/src/lens/common.graphql
```diff
@@ -470,6 +471,7 @@ fragment Comment on Comment {
 fragment Post on Post {
   __typename
   id
+  appId
```
krzysu, line 474: could you add it also in the client sdk?

cesarenaldi: Sure can

PR #460, opened 2023-07-28
Proposal: support attributes and image in non-collectable publications

cesarenaldi:


cesarenaldi:
See https://github.com/lens-protocol/lens-sdk/pull/463

PR #463, opened 2023-07-31
feat: support `attributes` and `image` in non-collectable publications

cesarenaldi:


packages/react/src/transactions/infrastructure/createPublicationMetadata.ts
```diff
@@ -94,42 +101,54 @@ function contentFields(request: CreatePublicationRequest) {
...
+    attributes: mapMetadataAttributes(request.collect.metadata.attributes ?? []),
+    image: request.collect.metadata.image,
+    imageMimeType: request.collect.metadata.imageMimeType,
+  } as const;
+}
+
+function resolveCollectableMetadata(request: CollectablePublicationRequest) {
+  if ('attributes' in request.collect.metadata && 'attributes' in request) {
+    console.warn(
+      'Both "attributes" and "collect.metadata.attributes" are provided. Using "collect.metadata.attributes"',
```
cesarenaldi, line 121: I ended up using existing format over new one as it felt the most backwards compatible way. Early in the SDK development I am aware of some JS integration that did pass these at the top level for lack of better documentation at the time.

With no visibility on those I prefer to avoid breaking examples, demos, docs until we release 2.0.

packages/domain/src/use-cases/publications/CreatePost.ts
```diff
@@ -61,6 +69,27 @@ export type BasePostRequest = {
...
+   */
+  attributes?: MetadataAttribute[];
+  /**
+   * The collect NFT image.
+   *
+   * This is the NFT image visible on marketplaces like OpenSea.
+   *
+   * DO NOT use this as primary storage for publication media. Use the `media` property instead.
+   * In the case your publication has many media consider to use this field as a static representation
+   * of the collect NFT. For example if the publication is an album of audio files this could well be
```
reecejohnson, line 85: nit: 'well' is redundant here 
```suggestion
   * of the collect NFT. For example if the publication is an album of audio files this could be
```

packages/react/src/transactions/infrastructure/createPublicationMetadata.ts
```diff
@@ -139,3 +158,12 @@ export function createPublicationMetadata(
       assertNever(request.collect, `Unexpected collect policy type`);
   }
 }
+
+export function createPublicationMetadata(request: CreatePublicationRequest): PublicationMetadata {
```
reecejohnson, line 162: this is composed nicely, very easy to read!

packages/react/src/transactions/useCreatePost.ts
```diff
@@ -79,6 +81,27 @@ export type CreatePostBaseArgs = {
...
+   */
+  attributes?: MetadataAttribute[];
+  /**
+   * The collect NFT image.
+   *
+   * This is the NFT image visible on marketplaces like OpenSea.
+   *
+   * DO NOT use this as primary storage for publication media. Use the `media` property instead.
+   * In the case your publication has many media consider to use this field as a static representation
+   * of the collect NFT. For example if the publication is an album of audio files this could well be
```
reecejohnson, line 97: nit: same again here with 'well', not necessarily needed 
```suggestion
   * of the collect NFT. For example if the publication is an album of audio files this could well be
```

reecejohnson (approved):
LGTM ✅

PR #465, opened 2023-08-01
feat: check if follower can follow followee before API request

krzysu:


packages/react/src/transactions/useFollow.ts
```diff
@@ -163,6 +163,8 @@ export function useFollow({ followee, follower }: UseFollowArgs): FollowOperatio
         );
       }
 
+      invariant(followee.followStatus?.canFollow, 'You already follow this profile');
```
cesarenaldi, line 166:
```suggestion
      invariant(followee.followStatus?.canFollow, 'You already following this profile. Check the `followee.followStatus.canFollow' to determine if you can call `useFollow`.);
```

cesarenaldi (approved):

PR #467, opened 2023-08-02
feat: skippable read hooks

krzysu:
type-safe skippable hooks (one for read results and one for paginated read results).
I'm not happy with the solution but also don't know any better one.

### DIFF
--- a/packages/react/src/profile/useProfile.ts
+++ b/packages/react/src/profile/useProfile.ts
@@ -11,7 +11,7 @@ import {
   useSourcesFromConfig,
   WithObserverIdOverride,
 } from '../helpers/arguments';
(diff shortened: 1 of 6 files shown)


packages/react/src/profile/useProfile.ts
```diff
@@ -47,38 +56,47 @@ export function useProfile({
...
       ),
     ),
   );
 
+  if (skip) {
+    return {
+      data: undefined,
+      error: undefined,
+      loading: false,
+    } as SkippableReadResult<A, Profile, NotFoundError | UnspecifiedError>;
```
krzysu, line 70: I don't like the need for `as ReturnType` but I didn't figure out how to do it better. Any ideas?

krzysu:
parked till lens v2 react hooks

PR #483, opened 2023-08-10
feat: returns newly created post from `useCreatePost`

cesarenaldi:


packages/shared-kernel/src/Result.ts
cesarenaldi, line 1: These changes are making narrowing of `Success<T, E>` and `Failure<T, E>` easier so that assigning of these to generic `Result<T, E>` works even when non contextual with the `success` (or `failure`) calls.

packages/react/src/transactions/adapters/PublicationCacheManager.ts
```diff
@@ -25,4 +68,19 @@ export class PublicationCacheManager {
       },
     );
   }
+
+  private async request(request: RequestPublicationArgs, fetchPolicy: FetchPolicy) {
+    const { data } = await this.client.query<GetPublicationData, GetPublicationVariables>({
+      query: GetPublicationDocument,
+      variables: {
+        request,
+        observerId: null, //session?.type === SessionType.WithProfile ? session.profile.id : null,
```
krzysu, line 77: shall we use `session.type` here as in the commented out code?

cesarenaldi: good catch... forgot to re-instate this.

packages/react/src/transactions/adapters/CreatePostPresenter.ts
```diff
@@ -0,0 +1,61 @@
...
+    this.deferredResult.resolve(success(post));
+  }
+
+  asResult(): Result<CreatePostAsyncResult, BroadcastingError> {
+    if (this.earlyFailure) {
+      return this.earlyFailure;
+    }
+
+    return success({
+      waitForCompletion: async () => {
```
reecejohnson, line 56: nice!

packages/domain/src/use-cases/transactions/__tests__/PayTransaction.spec.ts
```diff
@@ -51,16 +51,15 @@ describe(`Given the ${PayTransaction.name} interactor`, () => {
...
+        - queue the resulting ${NativeTransaction.name} into the ${TransactionQueue.name}`, async () => {
       const wallet = mockWallet();
       const queue = mockTransactionQueue<AnyTransactionRequestModel>();
       const gateway = mockIPayTransactionGateway({
         request,
         wallet,
         unsignedTransaction,
       });
-      const presenter = mock<IPayTransactionPresenter>();
+      const presenter = mock<IPayTransactionPresenter<AnyTransactionRequestModel>>();
```
reecejohnson, line 62: Unrelated to these changes but the first time i've seen `PayTransaction` and it seems to not read correct. 

Would `PaidTransaction` be more appropriate?

reecejohnson (approved):

packages/domain/src/use-cases/transactions/__tests__/PayTransaction.spec.ts
cesarenaldi: In general use cases are actions that can be performed on a given app. So the verb is usually preferable over nouns.

This one being about "paying for a transaction"... the "paid transaction" could describe a good name for an outcome objects.

Maybe "pay transaction" could be "pay for transactions"? Not sure.

packages/react/src/transactions/adapters/PublicationCacheManager.ts
cesarenaldi: I am not going to add extensive test coverage around this... the whole need of passing current profile id to queries like this one will go away with Lens API v2.

PR #484, opened 2023-08-10
Proposal: forward `Result<T, E>` without need to rewrap

cesarenaldi:
This was something discussed a long time ago and partially available in some 3rd party implementations of the Result monad.

In this PR you can see the change I am suggesting to the `Result` definition, which are backwards compatible as well as some places to showcase the impact.

If the proposal gets approved I will amends the rest of the codebase to avoid rewrapping in `success(...)`, `failure(...)`.

### DIFF
--- a/packages/shared-kernel/src/Result.ts
+++ b/packages/shared-kernel/src/Result.ts
@@ -64,7 +64,7 @@ export interface IEquatableError<T extends string = string, P = Narrow<T>> {
 }
 
 /**
- * A `Result` type represents either `Success` or failure `Failure`.
+ * A `Result` type represents either `Success` or `Failure`.
  *
  * **TL;DR**
  *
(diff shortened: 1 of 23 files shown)


reecejohnson (approved):
this is lovely!

packages/shared-kernel/src/Result.ts
```diff
@@ -10,7 +10,7 @@ import { Narrow } from './ts-helpers/types';
  * @sealed
  * @privateRemarks DO NOT EXPORT, see type export later on
  */
-class Success<T, E> {
+class Success<T, E extends IEquatableError> {
```
desfero, line 13: If we can afford the breaking change and removal of the second generic would simplify to `Success<T>` and `Failure<E>`. In this case you dont' even need `forward` as from the type perspective it should match the value correctly.

cesarenaldi: I think it breaks the relationship with the underlying error/data?
image

cesarenaldi: even worse I get this:
image

cesarenaldi: Ignore me.

cesarenaldi:
Closing this in favour of: https://github.com/lens-protocol/lens-sdk/pull/575

PR #491, opened 2023-08-11
fix: misc issues reported

cesarenaldi:
See individual commits.


krzysu:
so many issues, means people started using our sdk for real! 🥳

PR #497, opened 2023-08-15
Release 1.3-next.10

cesarenaldi:


packages/wagmi/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@lens-protocol/wagmi",
-  "version": "2.1.0-next.9",
+  "version": "3.0.0-next.10",
```
krzysu, line 3: 2.1.0-next.10

krzysu (approved):
but one version bump went wrong so please fix it before releasing

packages/wagmi/package.json
cesarenaldi: Great catch... didn't add it to staged files 🤦

PR #512, opened 2023-08-29
Fix support for ERC1155 gated content

desfero:


packages/gated-content/src/conditions/nft-condition.ts
```diff
@@ -29,30 +29,38 @@ const _handleOwnsAtLeastOneNftFromCollection = (): Partial<LitAccessCondition> =
...
 ): Partial<LitAccessCondition> => ({
   method: LitKnownMethods.BALANCE_OF_BATCH,
-  parameters: [
-    LitKnownParams.USER_ADDRESS.repeat(tokenIds.length).slice(0, -1),
-    tokenIds.join(','),
-  ],
+  parameters: [tokenIds.map(() => LitKnownParams.USER_ADDRESS).join(','), tokenIds.join(',')],
+  returnValueTest: {
+    comparator: LitScalarOperator.GREATER_THAN,
+    value: '0',
```
zannis, line 62: can use const ZERO here just to be aligned with rest

zannis (changes requested):
looking good just one comment

packages/gated-content/src/conditions/__tests__/nft-condition.spec.ts
```diff
@@ -137,14 +137,84 @@ describe(`Given the "${transformNftCondition.name}" function`, () => {
     },
 
     {
-      description: 'a NFT Ownership condition that has invalid contract type',
+      description: 'a ERC20 Ownership condition that has invalid contract type',
       condition: mockNftOwnershipInput({
         chainID: ethereumChainId,
         contractAddress: knownAddress,
         contractType: ContractType.Erc20,
```
cesarenaldi, line 144: I might have wrote this but still I am confused by this. Why "NFT ownership" with `ContractType.Erc20`?

packages/domain/src/entities/Publication.ts
```diff
@@ -29,6 +29,8 @@ export enum Erc20ComparisonOperator {
 /**
  * The active wallet must satisfy the NFT ownership criteria
  */
+// TODO: Introduce a new type that splits NFT ownership criteria into two types: one for ERC721 and one for ERC1155
+// This would allow to force the `tokenIds` field to be always presented for ERC1155
```
cesarenaldi, line 33: Why not removing the `?` from `tokenIds`?

packages/gated-content/src/conditions/__tests__/nft-condition.spec.ts
desfero: Seems like a leftover as this test is also repeated in the `erc20-condition.spec.ts`, removed

packages/domain/src/entities/Publication.ts
desfero: The ERC721 can be called without any sort of `id` to just check if the wallet has any nft from the collection.
I tried to split it  by `type NFTOwnershipCriterion = ERC721OwnershipCriterion | ERC1155OwnershipCriterion` but was causing an issues with `NftOwnershipCriterionSchema` which would require it to be also an union.
But due to the `type` being inside inside `z.discriminatedUnion` i can't repeat the same type twice in the schema.
One approach to consider at some point is to split `DecryptionCriteriaType.NFT_OWNERSHIP` by `DecryptionCriteriaType.ERC721_OWNERSHIP` and `DecryptionCriteriaType.ERC1155Ownership` which i think is more clear from the consumer side as you can at least see that quite likely different contract methods will be called to check the ownership.

cesarenaldi: Interesting, will look into token gated support for metadata v3 soon. Will include your suggestion.

zannis (approved):
lgtm!

PR #514, opened 2023-09-08
feat: token-gated support for Publication Metadata v3 and Lens API v2

cesarenaldi:


packages/client/codegen-api.yml
```diff
@@ -54,7 +54,8 @@ config:
     field: true
 
 schema:
-  - https://api-v2-mumbai.lens.dev/graphql
+  # - https://api-v2-mumbai.lens.dev/graphql
+  - http://localhost:4000/
```
krzysu, line 58: can we revert this line?

packages/client/src/helpers/buildImageTransformsFromConfig.ts
```diff
@@ -0,0 +1,9 @@
+import { MediaTransformsConfig } from '../context';
+
+export function buildImageTransformsFromConfig(config: MediaTransformsConfig = {}) {
```
krzysu, line 3: I think this file is no longer in use

krzysu (approved):

PR #516, opened 2023-09-12
feat: create profile & reactions

reecejohnson:


packages/client/src/LensClient.ts
```diff
@@ -109,6 +110,10 @@ export class LensClient {
     return new Revenue(this.config, this._authentication);
   }
 
+  get reactions(): Reactions {
```
krzysu, line 113: this doesn't make sense, Reactions are a submodule of Publication, so you access them as ie. `client.publication.reactions.add()`

examples/node/scripts/reactions/whoReacted.ts
```diff
@@ -0,0 +1,41 @@
+import {
```
krzysu, line 1: all reaction examples should be inside /scripts/publication/reactions folder I think as it is a submodule of Publication

examples/node/scripts/reactions/addReaction.ts
```diff
@@ -0,0 +1,33 @@
+import { LensClient, PublicationReactionType, development } from "@lens-protocol/client";
```
krzysu, line 1: filenames should be the same as method name, so /scripts/publication/reactions/ and then
- add.ts
- remove.ts
- fetch.ts

packages/client/src/submodules/profile/Profile.ts
```diff
@@ -207,16 +207,11 @@ export class Profile {
     });
   }
 
-  async create(
+  async createWithHandle(
```
krzysu, line 210: is there any way to create a profile without a handle? then I would keep the method name simple and short `profile.create()`

examples/node/scripts/profile/interests.ts
```diff
@@ -0,0 +1,26 @@
+import { LensClient, development } from "@lens-protocol/client";
+
+async function main() {
+  const lensClient = new LensClient({
+    environment: development,
+  });
+
+  // add interests
+  await lensClient.profile.addInterests({
+    interests: ["TECHNOLOGY", "HEALTH_FITNESS__EXERCISE"],
```
krzysu, line 10: the interest values should be an enum I guess, how to keep consistency otherwise? there used to be an API endpoint to get all values for interests, but we could also just hardcode them as enum in the sdk.

packages/client/src/submodules/profile/Profile.ts
reecejohnson: there is a claim profile - but that is separate

intention here was to keep the sdk as close to the GQL naming as possible

i think it makes sense to shorten here though, you're right

examples/node/scripts/profile/interests.ts
reecejohnson: there is the `profileInterestsOptions` endpoint but the interests are returned as strings

shall we create the enum on sdk side or API side?

examples/node/scripts/reactions/addReaction.ts
krzysu: I see, then up to you but please nest it under /publication/ folder to keep the structure

packages/client/src/submodules/publication/index.ts
```diff
@@ -1,3 +1,4 @@
 export * from './Publication';
 export * from './helpers';
 export * from './types';
+export * from './submodules';
```
krzysu, line 4: this is not needed as submodules are internal interest of publication module

packages/client/src/submodules/profile/Profile.ts
```diff
@@ -207,16 +207,11 @@ export class Profile {
...
-    RelaySuccessFragment | CreateProfileWithHandleErrorResultFragment,
-    CredentialsExpiredError | NotAuthenticatedError
-  > {
-    return requireAuthHeaders(this.authentication, async (headers) => {
-      const result = await this.sdk.CreateProfileWithHandle({ request }, headers);
-      return result.data.result;
-    });
+  ): PromiseResult<RelaySuccessFragment | CreateProfileWithHandleErrorResultFragment, never> {
+    const result = await this.sdk.CreateProfileWithHandle({ request });
+    return success(result.data.result);
```
krzysu, line 214: if the result is always a success, then we don't need to wrap it in the Result at all, do we? I'd rather change the return type then.

krzysu: so `Promise<RelaySuccessFragment | CreateProfileWithHandleErrorResultFragment>` ??

packages/client/src/submodules/publication/submodules/reactions/index.ts
```diff
@@ -1 +1,2 @@
+export { PublicationReactionType } from '../../../../graphql/types.generated';
```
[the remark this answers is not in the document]
reecejohnson: okay - we can do this!

packages/client/src/submodules/profile/Profile.ts
reecejohnson: sure!

packages/client/src/submodules/publication/submodules/reactions/index.ts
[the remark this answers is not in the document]
reecejohnson: cool, have remove export from here and added to there too, may be a slight conflict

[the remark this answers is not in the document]
reecejohnson: cool, have remove export from here and added to there too, may be a slight conflict

examples/node/scripts/reactions/addReaction.ts
reecejohnson: done 👍

krzysu (approved):

PR #519, opened 2023-09-13
feat: more publication docs

krzysu:


packages/client/v2.graphql
```diff
@@ -0,0 +1,216 @@
+type Query {
+  challenge(request: ChallengeRequest!): AuthChallengeResult!
```
krzysu, line 2: I added this subset of the schema to have easy reference to all queries and mutations and to spot if anything is missing, I will remove it later

reecejohnson (approved):
looks good 👍

PR #538, opened 2023-09-22
chore: prepare release client 2.0.0-alpha.3

krzysu:


krzysu:
it was already released on Friday

PR #548, opened 2023-09-28
feat: define scalar types

krzysu:


packages/api-bindings/codegen-api.yml
```diff
@@ -44,20 +44,20 @@ generates:
...
+        Nonce: number
         OnchainPublicationId: string
         PoapEventId: string
-        ProfileId: string
-        PublicationId: string
+        ProfileId: ProfileId
+        PublicationId: PublicationId
         Signature: string
         TokenId: string
         TxHash: string
```
reecejohnson, line 54: what's the plan for the one's that remain unbranded?

krzysu: I don't know what to do with the others, I just copied what we have in v1
@cesarenaldi do you want to brand anything more?

cesarenaldi (approved):

packages/api-bindings/codegen-api.yml
krzysu: @reecejohnson there will be more branded types, but we decided with @cesarenaldi that we will add them when needed. this is good enough to start with, as ProfileId, PublicationId and Cursor are what we really need for now.

PR #550, opened 2023-09-28
add examples web app to test v2 hooks

krzysu:


examples/web/README.md
```diff
@@ -0,0 +1,15 @@
+# Lens SDK + wagmi + Bundlr
```
reecejohnson, line 1: maybe note here that the example is using `ethers` as well

people may be looking for a `viem` example so being explicit about which tool here would be useful

examples/web/src/App.tsx
```diff
@@ -0,0 +1,76 @@
...
+    new InjectedConnector({
+      options: {
+        shimDisconnect: false,
+      },
+    }),
+  ],
+});
+
+const lensConfig: LensConfig = {
+  // bindings: wagmiBindings(),
```
reecejohnson, line 40: is this still needed?

reecejohnson (approved):
looks good, two small comments 🙏

PR #556, opened 2023-10-03
react: add missing tests, remove unnecessary arg destructuring, remove limit overwrites 

krzysu:


packages/react/src/profile/__tests__/useProfile.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+  mediaTransformConfigToQueryVariables,
+} from '../../mediaTransforms';
+import { useProfile, UseProfileArgs } from '../useProfile';
+
+function setupTestScenario({ profile, ...args }: UseProfileArgs & { profile: Profile | null }) {
+  const sources = mockSources();
+
+  return renderHookWithMocks(() => useProfile(args), {
+    mocks: {
+      sources,
```
cesarenaldi, line 24: Can we remove `sources`?

krzysu: yes, I'm removing sources in another PR I'm working on right now, with useProfiles

cesarenaldi (approved):

PR #560, opened 2023-10-04
feat: enable typedoc for new react package

reecejohnson:


packages/react/src/publication/usePublications.ts
```diff
@@ -15,41 +15,70 @@ import { PaginatedArgs, PaginatedReadResult, usePaginatedReadResult } from '../h
...
+ *      publicationTypes: [PublicationType.Post]
+ *    }
+ * });
+ * ```
+ * @example
+ * Fetch comment publications
+ * ```tsx
+ * const { data, loading, error } = usePublications({
+ *   where: {
+ *      publicationTypes: [PublicationType.Comment]
```
cesarenaldi, line 37: does it work without profile id at all?

reecejohnson: sorry, just saw this - but yes it does

reecejohnson: added these examples to show how people could replicate 

`useComments`
`usePosts`
`useMirrors`
`useQuotes`

with this hook

cesarenaldi: Worth looking at these examples as most likely use case.

If we are to got with `usePublications` for comments (and not implement `useComments`)... which I think it's an interesting idea BTW... maybe a more likely example is "find all comments on a given `publicationId`.

PR #561, opened 2023-10-04
react: useProfiles hook

krzysu:
- add useProfiles hook
- remove `sources`


packages/react/src/profile/useProfiles.ts
```diff
@@ -0,0 +1,36 @@
...
+ * {@link useProfiles} hook arguments
+ */
+export type UseProfilesArgs = PaginatedArgs<ProfilesRequest>;
+
+/**
+ * `useProfiles` is a paginated hook that lets you fetch profiles based on a set of filters.
+ *
+ * @category Profiles
+ * @group Hooks
+ */
```
reecejohnson, line 21: we should include some examples here as we go

packages/api-bindings/src/lens/__helpers__/queries.ts
```diff
@@ -64,12 +35,14 @@ export function mockPublicationResponse({
...
-      data: mockPublicationData(publication),
+      data: {
+        result: publication,
+      },
     },
   };
 }
 
-export function mockGetPublicationsResponse({
+export function mockPublicationsResponse({
```
reecejohnson, line 45: thanks!

reecejohnson (approved):
looks good - if we can add examples for each hook as we bring it in, will save us doing a bit documentation swoop at the end

PR #564, opened 2023-10-04
feat: add search publications

reecejohnson:


packages/react/src/helpers/reads.ts
```diff
@@ -19,6 +19,8 @@ import { useCallback, useEffect, useRef, useState } from 'react';
 
 import { useSharedDependencies } from '../shared';
 
+export { LimitType };
```
krzysu, line 22: I think all enums and other generated types should be exported from this file, to keep it consistent packages/api-bindings/src/lens/index.ts

reecejohnson: this is slightly different, we're exporting from the react package to be used in clients having installed that package

krzysu: true, you are right, then sth like in v1 I guess, https://github.com/lens-protocol/lens-sdk/blob/main/packages/react/src/index.ts#L120

PR #565, opened 2023-10-05
react: add useProfileFollowers, useProfileFollowing and useMutualFollowers

krzysu:


packages/api-bindings/src/lens/__helpers__/queries/profile.ts
```diff
@@ -0,0 +1,121 @@
...
+  ProfilesData,
+  ProfilesDocument,
+  ProfilesVariables,
+} from '../../graphql/generated';
+import { mockPaginatedResultInfo } from '../fragments';
+
+/**
+ * All paginated profile responses have the same shape
+ */
+function mockAnyPaginatedProfilesResponse<V, Q>({
```
krzysu, line 21: I decided to organize mocks a bit better, to avoid having one single huge file with everything.
I also extracted a reusable response shape into a helper function, used only internally. wdyt?

reecejohnson: Much better, great shout! Thanks Kris

reecejohnson (approved):

packages/react/src/__helpers__/setupTestScenario.ts
```diff
@@ -0,0 +1,23 @@
+import { MockedResponse } from '@apollo/client/testing';
+import { mockLensApolloClient } from '@lens-protocol/api-bindings/mocks';
+import { RenderHookResult } from '@testing-library/react';
+
+import { defaultMediaTransformsConfig } from '../mediaTransforms';
+import { renderHookWithMocks } from './testing-library';
+
+export function setupTestScenario(mocks: MockedResponse[]) {
```
cesarenaldi, line 8: nit: this exact wording (`setupTestScenario`) is used in many places with very local meaning...especially with things not associate with React, not sure if this is intelligible enough... I would consider:

- `setupUITestScenario`
- `setupReactTestScenario`
- `setupHookTestScenario`

PR #569, opened 2023-10-06
react: add useProfileActionHistory and useWhoReactedToPublication

krzysu:


packages/api-bindings/src/apollo/cache/field-policies/createProfileActionHistoryFieldPolicy.ts
```diff
@@ -0,0 +1,5 @@
+import { cursorBasedPagination } from '../utils/cursorBasedPagination';
+
+export function createProfileActionHistoryFieldPolicy() {
+  return cursorBasedPagination([]);
```
krzysu, line 4: I don't know if this is needed if there are no arguments

PR #582, opened 2023-10-09
react: add useCurrencies

krzysu:


packages/react/src/misc/useCurrencies.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  return {
+    ...result,
+    data: (result.data ?? []).map((currency) =>
+      erc20({
+        name: currency.name,
+        decimals: currency.decimals,
+        symbol: currency.symbol,
+        address: currency.contract.address,
+        chainType: ChainType.POLYGON,
```
cesarenaldi, line 82: fine for now... in the future we might have to resolve it from `currency.contract.chainId`... can we possibly leave a comment about this?

cesarenaldi (approved):
Just one comment

PR #591, opened 2023-10-10
v2/hide publication

reecejohnson:
- feat: initial pass at hide publication
- feat: add hide publication hook, adapters, example


packages/react/src/publication/useHidePublication.ts
```diff
@@ -0,0 +1,57 @@
...
+}: UseHidePublicationArgs): HidePublicationOperation {
+  const { data } = useSession();
+
+  const hide = useHidePublicationController();
+
+  return useOperation(async () => {
+    invariant(data?.authenticated, 'Must be authenticated to hide a publication');
+    invariant(
+      publication.by.ownedBy.address === data.address,
+      'Publication not owned by the active wallet. Make sure that publication is owned by the wallet (for .e.g. by calling `isPublicationOwnedByMe`) before trying to hide it?',
```
krzysu, line 48: do we use `isPublicationOwnedByMe` anywhere in v2?

krzysu (approved):
could you use "Squash and Merge" option to not bring unneeded PR history to the integration branch? I think it makes it easier to search the history later

packages/react/src/publication/useHidePublication.ts
reecejohnson: ah nice spot! thanks

PR #597, opened 2023-10-12
feat: implements `useCreatePost` hook

cesarenaldi:


examples/web/src/publications/PublicationsPage.tsx
```diff
@@ -1,6 +1,11 @@
 import { LinkCard } from '../components/LinkCard';
 
 const publicationHooks = [
+  {
+    label: 'CreatePost',
```
krzysu, line 5:
```suggestion
    label: 'useCreatePost',
```

packages/domain/src/use-cases/publications/CreateComment.ts
```diff
@@ -22,27 +22,21 @@ export type CreateCommentRequest = {
...
+  /**
+   * The post reference policy.
+   */
+  reference?: ReferencePolicyConfig;
+};
 
-export class CreateComment extends MomokaOption<CreateCommentRequest> {}
+export class CreateComment extends MomokaCapable<CreateCommentRequest> {
+  override async execute(request: CreateCommentRequest): Promise<void> {
+    if (['actions', 'reference'].some((key) => key in request)) {
```
krzysu, line 37: so that's how we know if post should go onchain or to momoka?

packages/domain/src/use-cases/publications/ReferencePolicyConfig.ts
```diff
@@ -32,7 +32,9 @@ export type DegreesOfSeparationReferencePolicyConfig = {
      */
     quotesRestricted: boolean;
     /**
-     * You can set the degree to follow someone elses graph, if you leave blank it use the authenticate Profile ID
+     * You can set the degree to follow someone elses graph.
+     *
+     * @defaultValue the authenticate Profile ID
```
krzysu, line 37:
```suggestion
     * @defaultValue the authenticated Profile ID
```

packages/react/src/transactions/adapters/publications/CreateMomokaPostGateway.ts
```diff
@@ -0,0 +1,131 @@
...
+
+    return UnsignedProtocolCall.create({
+      id: result.id,
+      request,
+      typedData: omitTypename(result.typedData),
+      fallback: this.createRequestFallback(request, result),
+    });
+  }
+
+  private async broadcast(
```
krzysu, line 69: I see that's how to use lens profile manager

packages/react/src/transactions/adapters/useCreatePostController.ts
```diff
@@ -0,0 +1,69 @@
...
+    const onChainPost = new SubsidizeOnChain(
+      activeWallet,
+      transactionGateway,
+      onChainGateway,
+      onChainRelayer,
+      transactionQueue,
+      presenter,
+    );
+
+    const delegableOnChainPost = new DelegableSigning(
```
krzysu, line 39: is this lens manager flow?

packages/react/src/transactions/useCreatePost.ts
```diff
@@ -0,0 +1,354 @@
...
+ *
+ * // second: upload it using the upload function you used to pass to `useCreatePost`:
+ * const uri = await uploadToIpfs(metadata);
+ *
+ * // finally, invoke the `execute` function:
+ * const result = await execute({
+ *   metadata: uri,
+ * })
+ *
+ * // continue as usual
```
krzysu, line 323: this is all super cool, but how is this maintainable? 
wouldn't it be easier to put long form docs like this one in a markdown file? 
I don't know if that would work with typedoc, but we can output typedoc to markdown and combine with long-form docs in markdown and render them as webpages (maybe with docusaurus).
just something to think about for the future.

krzysu (approved):
very nice, let's merge it soon so we can all build on top of it

packages/domain/src/use-cases/publications/CreateComment.ts
cesarenaldi: Yeah, i double checked this with Josh who said we should abstract this away from end users and use momoka whenever it makes sense.

packages/react/src/transactions/adapters/useCreatePostController.ts
cesarenaldi: correct... same as with dispatcher... the signing is delegated. No domain change in these regards, just implementation details.

packages/react/src/transactions/useCreatePost.ts
cesarenaldi: I think we need to write it regardless. Checking in with the code is the first step. Cross link to other resources is also important.

TSDoc thicks both boxes.

I do agree, we should not exclude a mixed approach with MD based tools and TSDoc. For now we need to move, so this is better than nothing.

PR #600, opened 2023-10-13
feat: set profile metadata v2

reecejohnson:


packages/react/src/transactions/useSetProfileMetadata.ts
```diff
@@ -0,0 +1,58 @@
...
+/**
+ * Set a profile's metadata.
+ *
+ * You MUST be authenticated via {@link useLogin} to use this hook.
+ *
+ * @category Profiles
+ * @group Hooks
+ * @param args - {@link UseSetProfileMetadataArgs}
+ */
+export function useSetProfileMetadata({
```
krzysu, line 33: I think we talked about calling it `useUpdateProfileMetadata` to match other similar hooks name
also in other hooks we use useDeferredTask but honestly I don't understand the difference.

from user perspective the difference is in `isPending` vs `loading` returned, so better to have it consistent among similar hooks.

cc @cesarenaldi what shall we use?

cesarenaldi: `useDeferredTask` is a generalization of `useOperation` that can be used also for lazy read hooks.

We should not longer use `useOperation`.

packages/domain/src/use-cases/profile/UpdateProfileDetails.ts
```diff
@@ -1,18 +1,10 @@
...
-  coverPicture?: string | null;
-  delegate: boolean;
+  metadataURI: string;
   kind: TransactionKind.UPDATE_PROFILE_DETAILS;
-  name: string;
-  profileId: ProfileId;
+  delegate: boolean;
 };
 
 export class UpdateProfileDetails extends DelegableSigning<UpdateProfileDetailsRequest> {}
```
cesarenaldi, line 10: Should we rename this `SetProfileMetadata`?

reecejohnson: sure

PR #603, opened 2023-10-13
react: replace useOperation with useDeferredTask

krzysu:


packages/react/src/publication/useHidePublication.ts
```diff
@@ -2,46 +2,44 @@ import { AnyPublication } from '@lens-protocol/api-bindings';
...
  * function HidePublication({ publication }: { publication: AnyPublication }) {
- *  const { execute: hide, isPending } = useHidePublication({ publication });
+ *   const { execute: hide, loading } = useHidePublication({ publication });
  *
- *  return (
- *    <button onClick={hide} disabled={isPending}>
- *      Hide
- *    </button>
- * );
- *```
```
krzysu, line 35: as you notice, this example snippet is broken and there is no way for us to auto-detect it, so the more content we write inside comment blocks the more content we don't really control

cesarenaldi: This is true regardless of where we write examples. Here you spotted it 👍

packages/react/src/publication/useHidePublication.ts
```diff
@@ -2,46 +2,44 @@ import { AnyPublication } from '@lens-protocol/api-bindings';
...
+ *   return (
+ *     <button onClick={hide} disabled={loading}>
+ *       Hide
+ *     </button>
+ *   );
+ * }
+ * ```
+ *
+ * @category Publications
+ * @group Hooks
```
cesarenaldi, line 33:
```suggestion
 * @group Hooks
 * @param args - {@link UseHidePublicationArgs}
```

packages/react/src/publication/useReportPublication.ts
```diff
@@ -91,29 +84,30 @@ export type ReportPublicationOperation = Operation<void, never, [ReportPublicati
...
+ *       <button type="submit" disabled={loading}>
  *         Report
  *       </button>
  *     </form>
  *   );
  * }
  * ```
+ *
+ * @category Publications
+ * @group Hooks
```
cesarenaldi, line 100:
```suggestion
 * @group Hooks
 * @param args - {@link UseReportPublicationArgs}
```

PR #609, opened 2023-10-16
feat: implements `useOpenAction` hook

cesarenaldi:


packages/react/src/profile/useLazyProfile.ts
```diff
@@ -16,20 +16,20 @@ import { useDeferredTask, UseDeferredTask } from '../helpers/tasks';
 export type FetchProfileArgs = OneOf<ProfileRequest>;
 
 /**
- * Lazy version of {@link useProfile} hook.
+ * `useLazyProfile` is a lazy version of {@link useLazyProfile} React Hook.
```
krzysu, line 19:
```suggestion
 * `useLazyProfile` is a lazy version of {@link useProfile} React Hook.
```

packages/react/src/shared.tsx
```diff
@@ -179,8 +188,8 @@ export type SharedDependencies = {
   momokaRelayer: MomokaRelayer;
   onChainRelayer: OnChainRelayer;
   profileCacheManager: IProfileCacheManager;
-  providerFactory: ProviderFactory;
   publicationCacheManager: PublicationCacheManager;
+  tokenAvailability: TokenAvailability;
```
krzysu, line 192: as you added this to shared, could you clean up the extra instance in useFollowController, https://github.com/lens-protocol/lens-sdk/blob/9dd33b0370283d332402d2313b189b5883a8b3d5/packages/react/src/transactions/adapters/useFollowController.ts#L82

PR #620, opened 2023-10-18
feat(react): rework hook args to be passed to hook callback

reecejohnson:
- refactor: move useReportPublication args to callback
- refactor: move useHidePublication args to callback


examples/web/src/profiles/UseSetProfileMetadata.tsx
```diff
@@ -24,6 +24,11 @@ function UpdateProfileForm({ activeProfile }: UpdateProfileFormProps) {
     const location = formData.get('location') as string | undefined;
     const website = formData.get('website') as string | undefined;
 
+    if (!name || !bio || !location || !website) {
+      toast.error('Please fill all fields');
+      return;
+    }
```
cesarenaldi, line 30: This cannot really happen with HTML5 validation in place.

Do we need to be this strict? I relaxed this example cause it was a little annoying to use.

reecejohnson: It's very unlikely but people can manipulate the HTML to remove the requires attrs. 

This was more for type safety and avoiding adding an empty location/website attr, that seems undesirable

PR #638, opened 2023-10-24
feat(react): implement `useBlockProfiles` & `useUnblockProfiles`

reecejohnson:


examples/web/src/profiles/ProfilesPage.tsx
```diff
@@ -3,81 +3,91 @@ import { LinkCard } from '../components/LinkCard';
...
   {
     label: 'useUpdateFollowPolicy',
-    description: `Update the follow policy for the logged-in Profile.`,
+    description: 'Update the follow policy for the logged-in Profile.',
     path: '/profiles/useUpdateFollowPolicy',
   },
+  {
+    label: 'useUnblockProfiles',
+    description: `Unblock profiles.`,
+    path: '/profiles/useUnblockProfiles',
```
krzysu, line 77: do we need this separately from the 'useBlockProfiles & useUnblockProfiles' item below in this list?

examples/web/src/App.tsx
```diff
@@ -144,7 +145,9 @@ export function App() {
                       <Route path="useProfileActionHistory" element={<UseProfileActionHistory />} />
                       <Route path="useSetProfileMetadata" element={<UseSetProfileMetadata />} />
                       <Route path="useUpdateFollowPolicy" element={<UseUpdateFollowPolicy />} />
+                      <Route path="useUnblockProfiles" element={<UseBlockAndUnblockProfiles />} />
```
krzysu, line 148: do we need two routes to the same example?

packages/react/src/transactions/useSetProfileMetadata.ts
```diff
@@ -87,12 +87,8 @@ export function useSetProfileMetadata(): UseDeferredTask<
...
   return useDeferredTask(async ({ metadataURI }: UseSetProfileMetadataArgs) => {
     invariant(
-      session?.authenticated,
-      'You must be authenticated to set profile metadata. Use `useLogin` hook to authenticate.',
-    );
-    invariant(
-      session.type === SessionType.WithProfile,
-      'You must have a profile to set profile metadata.',
+      session?.type === SessionType.WithProfile,
+      'You must be authenticated with a profile to block a profile. Use `useLogin` hook to authenticate.',
```
krzysu, line 91: I think this message is in the wrong hook here, did you intend to change the logic in `useSetProfileMetadata` hook?

krzysu (approved):
just a few comments, the rest looks good! 👍

packages/react/src/transactions/useSetProfileMetadata.ts
reecejohnson: ah good spot, thanks!

PR #645, opened 2023-10-25
update dependencies - expected range

krzysu:


krzysu:
@cesarenaldi @reecejohnson I'd really like to merge this PR, I checked that codegen works fine, also examples work fine, just next-js one is breaking but the same happens to me on develop branch, so not related to the update of deps

examples/next-js/package.json
```diff
@@ -11,21 +11,21 @@
...
-    "eslint-config-next": "14.0.1"
+    "@types/node": "^18.18.11",
+    "@types/react": "^18.2.38",
+    "@types/react-dom": "^18.2.16",
+    "autoprefixer": "^10.4.16",
+    "eslint": "^8.54.0",
+    "eslint-config-next": "14.0.1",
+    "postcss": "^8.4.31",
+    "tailwindcss": "^3.3.5",
+    "typescript": "^5.3.2"
```
krzysu, line 29: this is all result of `pnpm -r update`

cesarenaldi: I reckon we should stick to TS 5.2 for a little longer. 5.3 has only 2 weeks.

pnpm-lock.yaml
cesarenaldi, line 1: We need to dedupe this after the update. In the RN branch I found several issue arising from non-deduped deps.

PR #653, opened 2023-10-27
react: fix profile cache manager

krzysu:


packages/react/src/environments.ts
```diff
@@ -108,35 +108,3 @@ export const development: EnvironmentConfig = {
...
- *
- * This is the environment to be used when you develop and you need to experiment with custom collect/follow modules.
- * Although the Lens contract are also deployed on Mumbai this is a separate deployment so expect different test data, profiles, users, etc.
- *
- * - Endpoint: https://api-sandbox-mumbai.lens.dev
- * - Chain IDs: 80001 (Mumbai), 5 (Goerli)
- * - Profile handle namespace: `test/`
- * - Environment specific timings
- */
-export const sandbox: EnvironmentConfig = {
```
krzysu, line 123: there won't be sandbox in v2

packages/react/src/profile/infrastructure/ProfileCacheManager.ts
```diff
@@ -52,30 +46,6 @@ export class ProfileCacheManager implements IProfileCacheManager {
     });
   }
 
-  // TODO no longer need to accept a profile id, only authenticated profile can be used
-  updateProfile(id: string, updateFn: (current: Profile) => Profile): void {
```
krzysu, line 56: I removed this to promote optimistic implementations using type policies, it was not in use anyway

packages/react/src/environments.ts
cesarenaldi: no?

packages/react/src/profile/infrastructure/ProfileCacheManager.ts
cesarenaldi: 😢

cesarenaldi (approved):

PR #659, opened 2023-10-31
react: add inbox hooks

krzysu:


packages/react-web/package.json
```diff
@@ -74,6 +75,7 @@
   },
   "peerDependencies": {
     "@lens-protocol/metadata": "^1.0.0",
+    "@xmtp/react-sdk": "^3.0.0",
```
cesarenaldi, line 78: Are we in a position to make this also optional?

krzysu: they are, see below

```
"peerDependenciesMeta": {
    "@xmtp/react-sdk": {
      "optional": true
    }
  },
```

cesarenaldi: can you please double check the import path? basically only importing the relevant hook should end up importing the XMTP SDK

krzysu: yes, I checked, all xmtp imports are in the inbox folder, which is even exported as `@lens-protocol/react-web/inbox`. I also tested example web app without xmtp peer dep and of course without inbox hooks, it all works fine.

krzysu:
@cesarenaldi the issue with xmtp messages constant rerender was solved, see last commit, this PR is then ready from my side.

examples/web/src/inbox/components/ConversationComposer.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { never } from '../../utils';
+
+type ConversationComposerProps = {
+  peerProfile: Profile;
+};
+
+export function ConversationComposer({ peerProfile }: ConversationComposerProps) {
+  const { startConversation, isLoading, error } = useStartLensConversation({
+    peerProfile,
+  });
```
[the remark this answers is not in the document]
krzysu: everywhere else in the sdks we use objects as arguments to make them open for future additions without breaking changes, so the same thought is here

PR #664, opened 2023-11-02
fix: add network to wallet when not present

reecejohnson:
Original fix was implemented in the wagmi bindings, which worked.

But this was only needed as we were trying to get a signer for a specific chainId. If we omit the chain id when retrieving the signer, there is no error thrown in the bindings and the SignerFactory implementation works as expected.


packages/wagmi/src/index.ts
```diff
@@ -49,14 +92,10 @@ export function bindings(): IBindings {
...
-          } catch (err) {
-            // best effort to switch network, if it fails, we just ignore it
-            if (!(err instanceof SwitchChainNotSupportedError)) {
-              throw err;
-            }
-          }
+          const walletClient = await getWalletClient();
+          invariant(walletClient, 'Wallet client not found');
+          await switchChain({ walletClient, chainId });
+          return signerFromWalletClient({ walletClient, chainId });
```
cesarenaldi, line 98: There is logic already doing this in the `SignerFactory`, rather than replicating it a this level, can we see why it does not seem to reach the underlying `window.ethereum` provider?

cesarenaldi: (through these bindings)

reecejohnson: thanks, much simplifier fix in the end to avoid the error being thrown in these bindings

packages/react/src/wallet/infrastructure/SignerFactory.ts
```diff
@@ -30,7 +30,7 @@ export class SignerFactory implements ISignerFactory {
     chainType,
   }: CreateSignerConfig): PromiseResult<RequiredSigner, WalletConnectionError> {
     const chainId = chainType ? this.chains[chainType].chainId : undefined;
-    const signer = await this.bindings.getSigner({ chainId });
+    const signer = await this.bindings.getSigner({});
```
cesarenaldi, line 33: can you please clarify what this change achieve?

cesarenaldi: I don't know if it's safe to make this assumption on all potential bindings. You mentioned an error... which probably is the cause of the issue... do you have more details on the error? Can the wagmi bindings do a best effort approach and return the desired signer or fallback to chain-less Id if not available?

PR #670, opened 2023-11-03
react: upgrade credentials storage

krzysu:


packages/react/src/authentication/adapters/CredentialsStorageSchema.ts
```diff
@@ -0,0 +1,51 @@
...
+  role: 'refresh';
+  iat: number;
+  exp: number;
+};
+
+function isV1Content(decodedJwt: unknown): decodedJwt is V1JwtPayload {
+  return isObject(decodedJwt) && 'role' in decodedJwt && decodedJwt.role === 'refresh';
+}
+
+export class CredentialsStorageSchema extends BaseStorageSchema<typeof AuthData> {
```
cesarenaldi, line 23: `AuthData` is already available as a type in scope at line 10.

```suggestion
export class CredentialsStorageSchema extends BaseStorageSchema<AuthData> {
```

packages/react/src/authentication/adapters/CredentialsStorageSchema.ts
```diff
@@ -0,0 +1,51 @@
...
+  version = 2;
+
+  constructor(key: string) {
+    super(key, AuthData);
+  }
+
+  protected migrate(storageItem: IStorageItem<AuthData>): TypeOf<typeof AuthData> {
```
cesarenaldi, line 30: Same as above:
```suggestion
  protected migrate(storageItem: IStorageItem<AuthData>): TypeOf<AuthData> {
```

PR #679, opened 2023-11-07
feat: add useProfilesManaged hook

krzysu:


packages/api-bindings/src/lens/graphql/wallet.graphql
```diff
@@ -10,13 +10,18 @@ query OwnedHandles($for: EvmAddress!, $limit: LimitType, $cursor: Cursor) {
...
+  $limit: LimitType
+  $cursor: Cursor
   $profileCoverSize: ImageTransform = {}
   $profilePictureSize: ImageTransform = {}
   $activityOn: [AppId!]
   $fxRateFor: SupportedFiatType = USD
 ) {
-  result: profilesManaged(request: $request) {
+  result: profilesManaged(
+    request: { for: $for, includeOwned: $includeOwned, limit: $limit, cursor: $cursor }
```
cesarenaldi, line 23: ah, perfect, didn't know this flag existed.

cesarenaldi: I think the example code in (`LoginPage.tsx` and tsdoc of `useLogin` should mention this hook rather than `useProfiles`.

packages/react/src/wallet/useCanClaimHandle.ts
```diff
@@ -15,6 +17,14 @@ export type { ClaimableProfilesResult, ReservedClaimable } from '@lens-protocol/
  * @group Hooks
  */
 export function useCanClaimHandle(): ReadResult<ClaimableProfilesResult> {
+  const { data: session } = useSession();
+
+  invariant(session?.authenticated, 'You must be authenticated.');
+  invariant(
+    session.type === SessionType.JustWallet,
+    'You must be authenticated with a WalletOnlySession',
+  );
```
cesarenaldi, line 26: Was it failing otherwise?

cesarenaldi: Asking cause this throws in the render path, which at times could be annoying. It can stay if it needs to be.

packages/react/src/wallet/useProfilesManaged.ts
```diff
@@ -0,0 +1,36 @@
...
+  Profile,
+} from '@lens-protocol/api-bindings';
+
+import { useLensApolloClient } from '../helpers/arguments';
+import { PaginatedArgs, PaginatedReadResult, usePaginatedReadResult } from '../helpers/reads';
+
+/**
+ * {@link useProfilesManaged} hook arguments
+ */
+export type UseProfilesManagedArgs = PaginatedArgs<ProfilesManagedRequest>;
```
cesarenaldi, line 13: `ProfilesManagedRequest` includes too many low level features including `cursor` with the odd codegen typings `?: InputMaybe<...>`.

Can we list what relevant?

packages/react/src/wallet/useProfilesManaged.ts
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * @category Wallet
+ * @group Hooks
+ *
+ * @example
+ * ```tsx
+ * const { data, loading, error } = useProfilesManaged({
+ *   for: '0x1234567890123456789012345678901234567890',
+ * });
+ * ```
```
cesarenaldi, line 26: Can we add also show how to use it to list owne profiles too?

packages/react/src/wallet/useCanClaimHandle.ts
krzysu: if you have a wrong accessToken then the API returns Unauthorised error, so yes, I think it's better to fail early on the client

packages/react/src/wallet/useProfilesManaged.ts
krzysu: hm, `PaginatedArgs<>` removes the cursor and we use the same approach in every other paginated hook, what is different here?

packages/react/src/wallet/useCanClaimHandle.ts
cesarenaldi: ok

packages/react/src/wallet/useProfilesManaged.ts
cesarenaldi: I forgot 😅 ... all good!

.changeset/tame-kiwis-own.md
```diff
@@ -0,0 +1,7 @@
+---
+"@lens-protocol/api-bindings": patch
+"@lens-protocol/react": patch
+"@lens-protocol/react-web": patch
+---
+
+Added `useProfilesManaged` hook
```
cesarenaldi, line 7: nit: `**feat:** ...` reads more clear in the changelogs.

packages/react/src/wallet/useProfilesManaged.ts
krzysu: done

PR #682, opened 2023-11-07
fix: useDefferedTask called property when void return type

reecejohnson:


packages/react/src/helpers/__tests__/tasks.spec.ts
cesarenaldi, line 1: First of all, thank you for embarking into writing this test suite! It's something I initially tried to skip with the type definition you see in place, but now seems important to have.

In order to verify the accuracy of the tests I had to extract  the descriptions so that I could make my mind on the covered scenarios:
```
Given the useDeferredTask hook
  when the hook is idle
    should return state inline with type of "DeferredTaskIdle"

  when it is the first execution of the hook
    and the execution of the hook is loading
      should return the state inline with type of "DeferredTaskFirstCall"

    and the execution of the hook callback is successful
      and the hook callback returns data
        should return the state inline with type of "DeferredTaskSuccess"
       
      and the hook callback returns void
        should return the state inline with type of "DeferredTaskSuccess"

    and the execution of the hook callback returns a failure
      should return the state inline with type of "DeferredTaskFailed"

    and the execution of the hook callback throws an unexpected error
      should return the state inline with type of "DeferredTaskIdle"
        
  when it is the n-th execution of the hook
    and the execution of the hook callback is loading
      should return the state inline with type of "DeferredTaskNthCall"

    and the execution of the hook callback is successful
      should return the state inline with type of "DeferredTaskSuccess" with updated data

    and the execution of the hook callback returns a failure
      should return the state inline with type of "DeferredTaskFailed"

    and the execution of the hook callback throws an unexpected error
      should return the state inline with type of "DeferredTaskIdle"
```

cesarenaldi: One thing I often do is limit the number of nesting whenever possible and aim for a clear separation between:
- preconditions (i.e. the _given(s)_)
- the event (i.e. the **single** __when__)
- the expectations (i.e. the __then(s)__)

I exploded the above as it would be reported to a developer if each individual test would fail:

```
Given the useDeferredTask hook
  when the hook is idle
    should return state inline with type of "DeferredTaskIdle"

  when it is the first execution of the hook
    and the execution of the hook is loading
      should return the state inline with type of "DeferredTaskFirstCall"

  when it is the first execution of the hook
    and the execution of the hook callback is successful
      and the hook callback returns data
        should return the state inline with type of "DeferredTaskSuccess"
       
  when it is the first execution of the hook
    and the execution of the hook callback is successful
      and the hook callback returns void
        should return the state inline with type of "DeferredTaskSuccess"

  when it is the first execution of the hook
    and the execution of the hook callback returns a failure
      should return the state inline with type of "DeferredTaskFailed"
        
  when it is the first execution of the hook
    and the execution of the hook callback throws an unexpected error
      should return the state inline with type of "DeferredTaskIdle"
        
  when it is the n-th execution of the hook
    and the execution of the hook callback is loading
      should return the state inline with type of "DeferredTaskNthCall"

  when it is the n-th execution of the hook
    and the execution of the hook callback is successful
      should return the state inline with type of "DeferredTaskSuccess" with updated data
        
  when it is the n-th execution of the hook
    and the execution of the hook callback returns a failure
      should return the state inline with type of "DeferredTaskFailed"
        
  when it is the n-th execution of the hook
    and the execution of the hook callback throws an unexpected error
      should return the state inline with type of "DeferredTaskIdle"
```

cesarenaldi: The first thing that capture my attention is that there are several scenarios that seems to have same givens and same event, which usually is symptom of duplication.

But it's not the case. Instead the whens have a chained `and...` that specify their actual behaviour.

This is usually the case that some preconditions are not formalized as such but mixed with the event.

Given the importance of this test, I would aim at making it easy to scan. This time I knew enough about the problem and I dedicated more time than usual to read it carefully, but it might not be the case for a future PR in this area. If there is any friction in reading these tests, maybe on another busy day me (or anybody else) could skip through something that is actually important.

So jumping to the conclusions. I am suggesting a flatter organization and jargon that is reminiscent of the language we would use to explain this to another developer in a conversation or in documentation. Given the hook is called `useDeferredTask` I am reusing the word "task" and the concept of "success", "failure" to explain what happens.

```
Given the useDeferredTask hook
  when rendered for the first time
    should return state inline with type of "DeferredTaskIdle"

Given the useDeferredTask hook
  when the task is in progress
    should return the state inline with type of "DeferredTaskFirstCall"

Given the useDeferredTask hook
  when the task succeeds
    should return the state inline with type of "DeferredTaskSuccess"
    should support void as success data
      
Given the useDeferredTask hook
  when the task fails
    should return the state inline with type of "DeferredTaskFailed"

Given the useDeferredTask hook
  when the task throws an unexpected error
    should return the state inline with type of "DeferredTaskIdle"
        
Given the useDeferredTask hook
  and the hook is executed again
    when the task is in progress
      should return the state inline with type of "DeferredTaskNthCall"

Given the useDeferredTask hook
  and the hook is executed again
    when the task succeeds
      should return the state inline with type of "DeferredTaskSuccess" with updated data

Given the useDeferredTask hook
  and the hook is executed again
    when the task fails
      should return the state inline with type of "DeferredTaskFailed"

Given the useDeferredTask hook
  and the hook is executed again
    when the task throws an unexpected error
      should return the state inline with type of "DeferredTaskIdle"
```

which in a test report renders like this:
```
Given the useDeferredTask hook
  when rendered for the first time
    should return state inline with type of "DeferredTaskIdle"

  when the task is in progress
    should return the state inline with type of "DeferredTaskFirstCall"

  when the task succeeds
    should return the state inline with type of "DeferredTaskSuccess"
    should support void as success data

  when the task fails
    should return the state inline with type of "DeferredTaskFailed"

  when the task throws an unexpected error
    should return the state inline with type of "DeferredTaskIdle"
        
  and the hook is executed again
    when the task is in progress
      should return the state inline with type of "DeferredTaskNthCall"

    when the task succeeds
      should return the state inline with type of "DeferredTaskSuccess" with updated data

    when the task fails
      should return the state inline with type of "DeferredTaskFailed"

    when the task throws an unexpected error
      should return the state inline with type of "DeferredTaskIdle"
```

PR #687, opened 2023-11-09
feat: add useBlockedProfiles hook

krzysu:


packages/react/src/publication/useBookmarks.ts
```diff
@@ -0,0 +1,56 @@
...
+        },
+      }),
+    ),
+  );
+}
+
+/**
+ * @deprecated Use {@link useBookmarks} instead.
+ */
+export const useMyBookmarks = useBookmarks;
```
krzysu, line 51: I renamed this hook as it is the only one that uses "my" in the name, making it not consistent with the other naming, wdyt?

cesarenaldi: sounds good w/ me

packages/react/src/__helpers__/setupHookTestScenarioWithSession.ts
```diff
@@ -0,0 +1,36 @@
...
+        request: { forProfileId: sessionProfile.id },
+      },
+      result: sessionProfile,
+    }),
+
+    ...mocks,
+  ]);
+
+  const { result: sessionResult } = setupResult.renderHook(() => useSession());
+  await waitFor(() => expect(sessionResult.current.loading).toBeFalsy());
```
cesarenaldi, line 33: Can you please clarify this `waitFor`? Is `loading` anything else than `false` at the beginning?

packages/react/src/discovery/useFeed.ts
```diff
@@ -42,6 +44,14 @@ export type UseFeedArgs = OmitCursor<FeedRequest>;
  * ```
  */
 export function useFeed({ where }: UseFeedArgs): PaginatedReadResult<FeedItem[]> {
+  const { data: session } = useSession();
+
+  invariant(session?.authenticated, 'You must be authenticated.');
+  invariant(
+    session.type === SessionType.WithProfile,
+    'You must be authenticated with a profile to use this query. Use `useLogin` hook to authenticate.',
+  );
```
cesarenaldi, line 53: 👍

packages/react/src/publication/__tests__/useNotInterestedToggle.spec.ts
```diff
@@ -1,136 +1,69 @@
...
-        expect(scenario.updatedPublicationFragment).toMatchObject({
-          operations: {
-            isNotInterested: expected,
-          },
-        });
-      },
-    );
+      expect(publicationResult.current.data.operations.isNotInterested).toBe(false);
 
-    it('should update the `publication.mirrorOf.notInterested` in case the `publication` is a Mirror', async () => {
```
cesarenaldi, line 112: Is this now a very long test?

Can it be broken down?

packages/react/src/__helpers__/setupHookTestScenarioWithSession.ts
```diff
@@ -0,0 +1,36 @@
+import { MockedResponse } from '@apollo/client/testing';
+import { Profile, updateSessionData } from '@lens-protocol/api-bindings';
+import { mockProfileFragment, mockProfileResponse } from '@lens-protocol/api-bindings/mocks';
+import { waitFor } from '@testing-library/react';
+
+import { SessionType, useSession } from '../authentication';
+import { setupHookTestScenario } from './setupHookTestScenario';
+
+export async function setupHookTestScenarioWithSession(
```
cesarenaldi, line 9: Nice idea.

cesarenaldi (commented):
None of my comment are blocker to release the feature. If any work needs to be done... can be in a follow up PR.

packages/react/src/__helpers__/setupHookTestScenarioWithSession.ts
krzysu: useSession internally is loading a profile so we need to wait for it before allowing any other hook to use session to check for the authentication requirements.

packages/react/src/publication/__tests__/useNotInterestedToggle.spec.ts
krzysu: I did break it, but I didn't notice any time saving in test execution https://github.com/lens-protocol/lens-sdk/pull/687/commits/c554b32718b38e42f02db6fc425dd5a92903a9f7

cesarenaldi: My point was mostly in readability/maintainability. The need to add comments seems a symptom.

PR #690, opened 2023-11-10
feat: add invite hooks

krzysu:
- useInvitedProfiles
- useWasWalletInvited + useLazyWasWalletInvited
- useInviteWallets

### DIFF
--- a/packages/domain/src/use-cases/wallets/InviteWallets.ts
+++ b/packages/domain/src/use-cases/wallets/InviteWallets.ts
@@ -1,0 +1,53 @@
+import { EvmAddress, failure, success } from '@lens-protocol/shared-kernel';
+
+import { IGenericResultPresenter } from '../transactions';
+
+export type Invite = {
+  address: EvmAddress;
+};
+
+export type InviteWalletsRequest = {
+  wallets: EvmAddress[];
+};
+
+export class WalletAlreadyInvitedError extends Error {
+  name = 'WalletAlreadyInvitedError' as const;
+
+  constructor(public readonly invalid: EvmAddress[]) {
+    super(`Wallets already invited: ${invalid.join(', ')}`);
+  }
+}
+
+export interface IInviteWalletsFactory {
+  create(data: InviteWalletsRequest): Promise<Invite[]>;
+}
+
+export interface IInviteWalletsGateway {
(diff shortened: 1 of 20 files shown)


packages/react/src/misc/adapters/InviteWalletsGateway.ts
```diff
@@ -0,0 +1,34 @@
...
+        mutation: InviteDocument,
+        variables: {
+          request: {
+            invites: request.wallets,
+          },
+        },
+      });
+
+      return success();
+    } catch (error) {
```
krzysu, line 29: we don't really use this pattern but I don't see how else to catch API errors and turn them into user meaningful errors.

packages/react/src/misc/adapters/InviteWalletsGateway.ts
```diff
@@ -0,0 +1,34 @@
...
+      assertError(error);
+      return failure(new WalletAlreadyInvitedError(error.message));
```
krzysu, line 31: this could be any other api error too.

packages/react/src/misc/useWasWalletInvited.ts
```diff
@@ -0,0 +1,91 @@
...
+  const [fetch] = useProfileAlreadyInvitedLazyQuery(useLensApolloClient());
+
+  return useDeferredTask(async (args): PromiseResult<boolean, NotFoundError | UnspecifiedError> => {
+    const { data, error } = await fetch({ variables: { request: args } });
+
+    if (error) {
+      return failure(new UnspecifiedError(error));
+    }
+
+    if (!data) {
```
cesarenaldi, line 85: this should be an invariant and not a `NotFoundError` i think

packages/domain/src/use-cases/wallets/InviteWallets.ts
```diff
@@ -0,0 +1,53 @@
+import { EvmAddress, failure, success } from '@lens-protocol/shared-kernel';
+
+import { IGenericResultPresenter } from '../transactions';
+
+export type Invite = {
+  address: EvmAddress;
+};
```
cesarenaldi, line 7: Can this live in `entities` folder?

PR #693, opened 2023-11-14
fix: export api enums to use in hook args

krzysu:


packages/react/src/index.ts
```diff
@@ -84,14 +84,55 @@ export type {
...
 } from './helpers/reads';
-export { LimitType } from './helpers/reads';
 export * from './helpers/tasks';
 
 /**
  * GQL common types
  */
 export type { App } from '@lens-protocol/api-bindings';
-export { TriStateValue } from '@lens-protocol/api-bindings';
+
```
krzysu, line 93: could you help to identify which of these should NOT be exported? I just copied the list from the client SDK and removed a few I know are internal for react sdk

packages/react/src/index.ts
```diff
@@ -84,14 +84,46 @@ export type {
...
+// GQL enums
+export {
+  ChangeProfileManagerActionType,
```
cesarenaldi, line 96: Remove: `useUpdateProfileManagers` provides an easier DX

PR #698, opened 2023-11-15
feat: allows self-funded useOpenAction

cesarenaldi:


packages/react/src/transactions/useOpenAction.ts
```diff
@@ -53,6 +53,18 @@ export type OpenActionArgs = {
    * The publication to perform the Open Action on.
    */
   publication: AnyPublication;
+  /**
+   * Whether the transaction gas costs should be sponsored by the Lens API or
+   * should be paid by the authenticated wallet.
+   *
+   * Although the request is marked as sponsored there are scenarios where the
+   * sponsorship will not be denied. See {@link BroadcastingError} with:
```
reecejohnson, line 61: should this read like: 

```suggestion
   * Although the request is marked as sponsored there are scenarios where the
   * sponsorship will be denied. See {@link BroadcastingError} with:
```

reecejohnson (approved):
Nice! 

`sponsored` is a great name and `signless` makes a lot more sense that `delegate`

PR #700, opened 2023-11-15
feat: add use client directive to react-web

reecejohnson:
After some digging, I found that preconstruct enabled the ability to add the "use client" directive to modules in: 
- https://github.com/preconstruct/preconstruct/pull/526

This adds the directive and updates preconstruct to 2.4.0


examples/web-wagmi/stats.html
cesarenaldi, line 1: I did the same, this file was accidentally checked-in after `examples/web-wagmi` got removed and with it the `gitignore` file within it.

.changeset/friendly-planes-run.md
```diff
@@ -0,0 +1,5 @@
+---
+"@lens-protocol/react-web": patch
+---
+
+adds use client directive to entry point
```
cesarenaldi, line 5: Can we use `**fix:**`, `**chore:**`, `**feat:**` prefix so that changelog is easy to read and share in release notes.

cesarenaldi: I usually use more or less the same copy of the commit (or PR title if contains more than one commit).

reecejohnson: sure

Issue #691, opened 2023-11-10
In the useLogin react hook, when the user cancel the signing message modal, an unhandled promise is being thrown instead of being handled by the SDK


pradel:
**Describe the bug**

In the useLogin hook, the SDK does not handle the signing message error properly. When the user rejects the sign message in their wallet, the error is not reported here but thrown as an unhandled promise rejection.

**To Reproduce**

1. go to https://lens-sdk-example-web-wagmi.vercel.app/login
2. click to login and select the profile
3. get the sign message popup from your wallet and cancel the request
4. see the unhandled promise in the console

**Expected behaviour**

The error should be catched and returned by the `isFailure` function

**If using the React bindings, what is the essence of your React app?**
- [x] React Web App
- [ ] React Native 

**Version of the packages you are using**

"@lens-protocol/react-web": "2.0.0-alpha.19"

reecejohnson:
thanks for raising this @pradel 

i was able to reproduce this, thanks for the detailed reproduction steps.

we are taking a look into the cause now.

i will keep update this issue once resolved 👍

reecejohnson:
@pradel see mentioned PR for small fix that better handles this error

this is not related directly to our React SDK but more so an issue with our example app using `wagmi` and `viem`

we do already handle the case where a user rejects the request in the login flow 

this error was coming from before the login flow was executed, in the connection of the wallet to example app

i've added an `onError` callback to this example to better handle this case

pradel:
@reecejohnson from my tests the error is coming from the login flow (when you sign a message), not the connect flow. It was working properly with the V1 of the SDK but not with the V2 so maybe a regression?

reecejohnson:
@pradel i see, can you provide some information around which wallet you're using? and a screen recording if possible

cesarenaldi:
@pradel any update on this?

pradel:
The issue is still happening with the latest version, I am using Taho Wallet and it can be reproduced in the lens example consistently https://lens-sdk-example-web-wagmi.vercel.app/login

pradel:
@cesarenaldi @reecejohnson check this is the video showing the issue I am facing.

https://github.com/lens-protocol/lens-sdk/assets/5749437/baf617b5-16f4-4e66-ac82-fb6c359abd86

cesarenaldi:
[...]

PR #701, opened 2023-11-16
fix: handle wagmi connection rejection in example web app

reecejohnson:
Add handling for errors that are thrown by the `useConnect` hook from `wagmi` in the example web app.


reecejohnson:
related to #691

cesarenaldi (approved):

PR #704, opened 2023-11-20
docs: add add app id docs to publication creation hooks

reecejohnson:


packages/react/src/transactions/useCreateMirror.ts
```diff
@@ -130,6 +135,29 @@ export type CreateMirrorArgs = {
  * };
  * ```
  *
+ * ## Creating an app-specific mirror
+ *
+ * You can create a mirror that is specific to an app by defining the `appId` when creating the mirror metadata. For a mirror, this is optional as if not provided the metadata of the original publication will be used.
```
reecejohnson, line 140: Is this correct way to describe what happens when the metadata is not provided @cesarenaldi?

PR #707, opened 2023-11-21
feat: advanced contract condition

cesarenaldi:


packages/gated-content/src/conditions/advanced-contract-condition.ts
```diff
@@ -0,0 +1,150 @@
...
+
+    // assert view function
+    invariant(fn.stateMutability === 'view', 'unsupported');
+
+    // assert single output
+    invariant(Array.isArray(fn.outputs) && fn.outputs.length === 1, 'unsupported');
+
+    // assert output is boolean or uint
+    invariant(
+      fn.outputs[0] && (fn.outputs[0].type === 'bool' || fn.outputs[0].type === 'uint256'),
```
zannis, line 52: might break with diff types of uints, suggesting this check instead 
```suggestion
      fn.outputs[0] && (fn.outputs[0].type === 'bool' || /^uint(8|16|32|64|128|256)$/.test(fn.outputs[0].type),
```

packages/gated-content/src/conditions/advanced-contract-condition.ts
```diff
@@ -0,0 +1,150 @@
...
+        }
+      } else if (input.baseType.includes('int')) {
+        BigNumber.from(param);
+      } else if (input.baseType === 'bool') {
+        invariant(
+          param === 'true' || param === 'false',
+          `param ${input.name} is invalid, must be a boolean)`,
+        );
+      } else if (input.baseType === 'bytes') {
+        invariant(param.startsWith('0x'), `param ${input.name} is invalid, must be a hex string)`);
```
zannis, line 103: a bit stronger checks 
```suggestion
        invariant(/^0x[0-9A-Fa-f]+$/.test(param), `param ${input.name} is invalid, must be a hex string)`);
```

packages/gated-content/src/conditions/advanced-contract-condition.ts
```diff
@@ -0,0 +1,150 @@
...
+    if (['bool', 'string', 'bytes', 'address', 'array', 'tuple'].includes(fn.outputs[0].baseType)) {
+      invariant(
+        condition.comparison === ConditionComparisonOperator.EQUAL ||
+          condition.comparison === ConditionComparisonOperator.NOT_EQUAL,
+        `comparison ${condition.comparison} is invalid for function return type ${fn.outputs[0].type}`,
+      );
+    }
+
+    // for uint results we allow all comparisons but we check the provided value
+    if (fn.outputs[0].baseType.includes('int')) {
```
zannis, line 143:
```suggestion
    if (/^uint(8|16|32|64|128|256)$/.test(fn.outputs[0].baseType)) {
```

zannis (changes requested):
left a few code change suggestions, couple stronger validation checks 🍻

PR #720, opened 2023-11-23
chore: update more deps, relax versions

krzysu:


krzysu:
this is just a follow-up on updating deps based on hints from renovate

PR #723, opened 2023-11-23
react: add inbox hooks

krzysu:
I'm testing the CI build as it fails on the other PR but I can't figure out why

### DIFF
--- a/packages/react-web/src/inbox/useEnhanceConversation.ts
+++ b/packages/react-web/src/inbox/useEnhanceConversation.ts
@@ -1,0 +1,105 @@
+import {
+  ProfileId,
+  Profile,
+  ReadResult,
+  useLazyProfile,
+  NotFoundError,
+  UnspecifiedError,
+} from '@lens-protocol/react';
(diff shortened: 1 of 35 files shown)


packages/react-web/src/inbox/useEnhanceConversation.ts
```diff
@@ -0,0 +1,105 @@
...
+    [conversation, profile],
+  );
+
+  const { data: peerProfile, error, loading, execute } = useLazyProfile();
+
+  useEffect(() => {
+    if (peerProfileId) {
+      void execute({ forProfileId: peerProfileId });
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
cesarenaldi, line 62: is this because of `execute`?

krzysu: yes, in general there is issue with hook being rendered multiple times, skipping `execute` helped a bit

krzysu:
the build is still failing so I'll close this one, seems it's not the CI issue

PR #742, opened 2023-11-28
chore: prepare release react 2.0.0-alpha.23

reecejohnson:


packages/react/CHANGELOG.md
```diff
@@ -1,5 +1,38 @@
...
+- 493895b8: **feat:** self-funded `useFollow`
+- fdd0073d: **feat:** support for self-funded `useCreateQuote`
+- 5d243a83: **feat:** support for self-funded `useCreateMirror`
+- b37f6f4e: **feat:** self-funded `useSetProfileMetadata`
+- 2e351a8c: fixes called property of hooks to update correctly when hook returns void
+- 2698fc65: **feat:** support for self-funded `useCreatePost`
+- 6a25dc02: **feat:** support self-funded `useOpenAction`
+- 40abddd9: **feat:** support for self-funded `useCreateComment`
+
+### Patch Changes
```
krzysu, line 16: as it's still alpha, the split for minor/patch doesn't really make much sense and it's easier to read if all points were in a single list, could we merge them (just in the changelog, better to not touch changeset files at this point).

packages/react-native/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@lens-protocol/react-native",
-  "version": "0.0.1",
+  "version": "0.1.0-alpha.0",
```
krzysu, line 3: is it ok to start react-native from 0.1.0 or from 0.0.1 ?

reecejohnson: good spot, thanks

PR #743, opened 2023-11-28
client: ensure interoperability with react SDK, add example

krzysu:


examples/web/src/misc/LensClientInteroperability.tsx
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  setItem(key: string, value: string) {
+    window.localStorage.setItem(key, value);
+  }
+
+  removeItem(key: string) {
+    window.localStorage.removeItem(key);
+  }
+}
```
cesarenaldi, line 19: Having the `LensProvider` using its own storage implementation and having this one accessing `localStorage` seems "uneven". Requires consumers to know something that until now was an internal.

Can we do something more encapsulated? Can we just use the same exact adapter instance that `@lens-protocol/react-web` use somehow?

PR #753, opened 2023-11-30
react: improve logout activities

krzysu:
a few things in this PR:
- react: clear transaction storage on logout
- react & client: remove migration path from credentials storage
- react: revoke credentials on logout 
- react: logout on any refresh token API request error


packages/storage/src/CredentialsStorageSchema.ts
```diff
@@ -20,13 +20,6 @@ export class CredentialsStorageSchema extends BaseStorageSchema<PersistedCredent
...
   ): PersistedCredentials {
-    const storageVersion = storageItem.metadata.version;
-
-    if (this.version > storageVersion) {
-      return this.parseData({
-        refreshToken: '',
-      });
-    }
     return this.parseData(storageItem.data);
   }
```
cesarenaldi, line 24: We can omit `migrate` I think.

packages/domain/src/use-cases/authentication/Logout.ts
```diff
@@ -24,17 +41,25 @@ export interface ILogoutPresenter {
...
+    if (!credentials) {
+      never('User is not authenticated');
+    }
+
+    await this.sessionGateway.revoke({ authorizationId: credentials.authorizationId });
     await this.walletGateway.reset();
     await this.conversationsGateway.reset();
-
+    await this.transactionGateway.reset();
     await this.credentialsGateway.invalidate();
```
cesarenaldi, line 63: I don't know if intentional or not, but with the `IRevokeSessionGateway` you are implicitly promoting "session" as a domain object.

Last time I worked on this area I thoughts that "credentials" became a poor abstraction over time (due to the changes we had to made to it). Session is a good replacement.

Replacing Credentials with Session is a bigger work for the scope of this ticket. So for now I would limit the revoke of credentials as implementation detail of `IResettableCredentialsGateway#invalidate`.

packages/storage/src/CredentialsStorageSchema.ts
krzysu: in the default migration method, it will throw if versions won't match, is that ok?

```
if (this.version !== storageVersion) {
      throw new NoMigrationPathError(this.key, storageVersion, this.version);
    }
```

PR #759, opened 2023-12-04
fix: logic that infers when to create Momoka publications

cesarenaldi:


reecejohnson (approved):

packages/react/src/transactions/adapters/schemas/publications.ts
```diff
@@ -108,10 +108,8 @@ export const CreatePostRequestSchema: z.ZodType<CreatePostRequest, z.ZodTypeDef,
     signless: z.boolean(),
     sponsored: z.boolean(),
     metadata: UriSchema,
-    reference: ReferencePolicyConfigSchema.optional(),
-    actions: OpenActionConfigSchema.array()
-      .min(1, 'You must provide at least one open action')
-      .optional(),
+    reference: ReferencePolicyConfigSchema.catch({ type: ReferencePolicyType.ANYONE }),
+    actions: OpenActionConfigSchema.array().catch([]),
```
krzysu, line 112: @cesarenaldi I've noticed now that `catch` is hiding any validation errors that could happen, ie. I have a misconfiguration of paid collect, I have amount defined but no recipient, and that should fail early, but with `catch` it is silenced and publication is created without any actions (defaults to [])

cesarenaldi: Good catch. I think we should use [`.default()`](https://zod.dev/?id=default).

PR #763, opened 2023-12-04
chore: add keywords to public packages

krzysu:


packages/react/package.json
```diff
@@ -21,6 +21,17 @@
...
+    "lens-api",
+    "lens-protocol-sdk",
+    "lens-protocol",
+    "lens",
+    "react",
+    "sdk",
+    "social",
+    "web3-social",
+    "web3"
+  ],
```
cesarenaldi, line 34: Maybe we should leave this one out. it might become internal as soon as the RN one is stable.

PR #780, opened 2023-12-11
feat: introduce debug mode, allow to set internal logger level

krzysu:


packages/shared-kernel/src/logger/ILogger.ts
```diff
@@ -2,8 +2,18 @@
  * Logger interface
  */
 export interface ILogger {
+  level: LoggerLevel;
```
krzysu, line 5: @cesarenaldi what do you think about extending ILogger like this and using this to determine debug mode deep in the code? see https://github.com/lens-protocol/lens-sdk/pull/780/files#diff-568da022b5078bcf9590f1377af950be03777482416db3e0d009de7b1fd32f25R52

packages/react/src/transactions/adapters/AbstractContractCallGateway.ts
```diff
@@ -41,21 +48,29 @@ export abstract class AbstractContractCallGateway<TRequest extends AnyTransactio
 
     const { contractAddress, encodedData } = await this.createEncodedData(request);
 
+    // skip gas estimation if debug mode is enabled
+    if (this.logger.level === LoggerLevel.Debug) {
```
cesarenaldi, line 52: I would rather pass `debug` flag here rather than a second hand information (i.e. the log level based on `debug` value), WDTY?

krzysu: the thing is that you need to pass this extra param all the way down through the class hierarchy, and it is used only at the bottom, is there any better way to do it?

packages/react/src/transactions/adapters/AbstractContractCallGateway.ts
```diff
@@ -41,21 +48,29 @@ export abstract class AbstractContractCallGateway<TRequest extends AnyTransactio
...
+      const transactionRequest = {
+        to: contractAddress,
+        from: wallet.address,
+        data: encodedData,
+      };
+
+      return new UnsignedContractCallTransaction(request, transactionRequest);
```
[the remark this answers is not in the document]
krzysu: I haven't tested it yet but I assumed ethers knows how to handle it. also, wallets are very capable of estimating the gas price afaik.

packages/react/src/transactions/adapters/AbstractContractCallGateway.ts
cesarenaldi: I understand, but this is no different than the effort already done here. Instead of passing the `logger` you would pass the `config` (or `config.debug` flag).

krzysu: ok, I changed to pass down `config` instead of `logger`

krzysu:
@cesarenaldi take a look at my latest commit please https://github.com/lens-protocol/lens-sdk/pull/780/commits/f7a3bd6b6e44ebe6865bdfa75afb7ad4d5f1ff80

PR #782, opened 2023-12-14
Use @ethersproject/address in place of ethers/lib/utils

DefiCake:
This is a small contribution. This issue is keeping me from using ethers v6 in conjunction with @lens-protocol/react-native. Because `getAddress` is getting routed through `ethers/lib/utils`, it is trying to resolve to the ethers v6 dependency instead of the v5 that the SDK needs.

I couldn't run the tests locally, I think I need some env configuration. Here is hoping that the change is small enough to not break anything.

EDIT. Some further investigation revealed that the direct use of the ethers library should also be removed, so I pushed additional changes

### DIFF
--- a/.changeset/quiet-garlics-live.md
+++ b/.changeset/quiet-garlics-live.md
@@ -1,0 +1,5 @@
+---
+"@lens-protocol/react": patch
+---
+
+Use @ethersproject/address instead of ethers/lib/utils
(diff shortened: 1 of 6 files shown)


krzysu:
hi @DefiCake I see what you mean, I can clean up this PR and take it from here

krzysu:
see this PR for complete removal of dependency on ethers5 https://github.com/lens-protocol/lens-sdk/pull/787

DefiCake:
I will close this PR in favor of yours @krzysu

PR #784, opened 2023-12-14
feat: allow decryptable content with wallet-only auth

reecejohnson:


packages/client/src/authentication/Authentication.ts
```diff
@@ -176,6 +176,22 @@ export class Authentication implements IAuthentication {
     return handler(profileId);
   }
 
+  async requireOnlyWalletAuthentication<
```
krzysu, line 179: this is more like `requireAnyAuthentication` where the other method is `requireProfileAuthentication` yes?

reecejohnson: good point, thanks

cesarenaldi (approved):

PR #787, opened 2023-12-15
chore: remove all ethers5 peer deps

krzysu:


packages/gated-content/package.json
```diff
@@ -52,7 +52,8 @@
     "@lit-protocol/encryption": "2.1.62",
     "@lit-protocol/node-client": "^2.1.62",
     "@lit-protocol/types": "2.1.62",
-    "siwe": "^1.1.6",
+    "@spruceid/siwe-parser": "^2.0.2",
```
cesarenaldi, line 55: Can you please provide some context why this is needed?

krzysu: I think I encountered this issue (see link) while upgrading siwe to remove its deps on ethers@5, but I see it worked fine without it too https://github.com/spruceid/siwe/issues/176

PR #804, opened 2024-01-23
client: allow to authenticate with an existing refresh token

krzysu:


packages/client/src/authentication/Authentication.ts
```diff
@@ -35,6 +35,11 @@ export class Authentication implements IAuthentication {
     this.credentials = new CredentialsStorage(context.storage, context.environment.name);
   }
 
+  async fromRefreshToken(refreshToken: string): Promise<void> {
+    const credentials = new Credentials(undefined, refreshToken);
+    await this.credentials.set(credentials);
```
krzysu, line 40: @cesarenaldi I made it as simple as this.
shall we validate if the input is a valid jwt token and if it is a lens refresh token or leave it out to the consumer? if not valid then the API will fail on the refresh step anyway and it will happen the first time you call any client method. wdyt?

packages/client/package.json
```diff
@@ -97,7 +97,7 @@
     "typescript": "5.2.2"
   },
   "engines": {
-    "node": "^18.15.0"
+    "node": ">=18"
```
cesarenaldi, line 100: What about a range up to current stable (i.e. 20)?

packages/client/src/authentication/Authentication.ts
cesarenaldi: I think a simple approach works for now. I have a question about the design choice.

Why you went for a method compared to a configuration option in the constructor?

cesarenaldi:
Hey @bigint, this change should enable you to use the `LensClient` for gated publication starting from the credentials you already have.

packages/client/src/authentication/Authentication.ts
krzysu: first that `client.authentication` module contains all methods related to authentication, so it's easier to find it here if you are checking what is possible, 
2nd, you might have a client instance from another place of your app and only authenticate later when you have a token.

krzysu: it's probably not clear but each form of authentication will overwrite the previous one

packages/client/package.json
krzysu: done

packages/client/src/authentication/Authentication.ts
cesarenaldi: I see and understand the rationale. The only thing is that the method name might not be intelligible. the "fromSomething" kinda make sense as static factory method (e.g. `Car.fromParts(...)`) but not sure if works well as instance method of a submodule.

cesarenaldi:
> Amazing @cesarenaldi when this is going live?

This week.

packages/client/src/authentication/Authentication.ts
krzysu: I'm open to better ideas 😄

packages/client/package.json
```diff
@@ -97,7 +97,7 @@
     "typescript": "5.2.2"
   },
   "engines": {
-    "node": "^18.15.0"
+    "node": ">=18 <21"
```
cesarenaldi, line 100: Can you please add another changeset with this so it surfaces in the changelog?

krzysu: done

packages/client/src/authentication/Authentication.ts
krzysu: I renamed it to

```ts
/**
* Authenticate with an existing, valid refresh token.
*/
authenticateWith({ refreshToken }: { refreshToken: string }): Promise<void>;
```

cesarenaldi:
> > Amazing @cesarenaldi when this is going live?
> 
> This week.

@bigint sorry, we had some last minute things to deal with which disrupted out SDK release. Aiming to get this out early next week.

cesarenaldi:
@bigint this is now released.

PR #805, opened 2024-01-23
chore: make Authentication required in all submodules

krzysu:


cesarenaldi:
LGTM, what was the rationale before this change?

packages/client/src/helpers/requireAuthHeaders.ts
```diff
@@ -1,19 +1,14 @@
...
 
 type Handler<Val> = (headers: Record<string, string>) => Promise<Val>;
 
 export async function requireAuthHeaders<Val>(
-  authentication: Authentication | undefined,
+  authentication: Authentication,
   handler: Handler<Val>,
 ): PromiseResult<Val, CredentialsExpiredError | NotAuthenticatedError> {
-  // TODO revisit this might be actually defined all the time
-  if (!authentication) {
```
krzysu, line 13: @cesarenaldi the rationale was this comment here and the fact that I had to use authentication in nfts.fetch to determine query params (in another PR) so I decided to just clean it up already everywhere

PR #807, opened 2024-01-25
fix: fix typos

rex4539:


rex4539:
@cesarenaldi Could you take a look?

packages/api-bindings/src/lens/utils/open-actions.ts
```diff
@@ -0,0 +1,209 @@
+/* eslint-disable no-case-declarations */
```
cesarenaldi, line 1: I suspect this file was accidentally re-introduced while resolving a conflict.

cesarenaldi (changes requested):
@rex4539 Thank you for taking the time.

In general all `packages/*-v1` are going to be deleted soon. No harm in your changes there, just letting you know.

Left a comment on a not expected file.

packages/api-bindings/src/lens/utils/open-actions.ts
rex4539: Removed.

rex4539:
All should be good now.

cesarenaldi (approved):

PR #810, opened 2024-01-29
fix: useOpenAction supports sponsoredApproved, signlessApproved flags

cesarenaldi:


packages/react/src/transactions/useFollow.ts
```diff
@@ -24,82 +24,69 @@ import {
...
+    session?.authenticated,
+    'You must be authenticated to use this operation. Use `useLogin` hook to authenticate.',
+  );
+  invariant(
+    session.type === SessionType.WithProfile,
+    'You must have a profile to use this operation.',
+  );
 
-      case FollowPolicyType.UNKNOWN:
-        const result = await fetchModule({ implementation: followPolicy.contractAddress });
```
krzysu, line 73: no fetching module data for unknown module anymore?

packages/react/src/transactions/useFollow.ts
```diff
@@ -24,82 +24,69 @@ import {
...
-        return {
-          kind: TransactionKind.FOLLOW_PROFILE,
-          profileId: args.profile.id,
-          signless: session.profile.signless,
-          sponsored: args.sponsored ?? true,
-        };
+/**
+ * @internal
+ */
+export function createFollowRequest(args: FollowArgs, session?: Session): FollowRequest {
```
krzysu, line 37: does it need to be exported? I don't see it used anywhere outside of this file

packages/react/src/transactions/useBlockProfiles.ts
```diff
@@ -34,9 +34,10 @@ export type BlockProfileArgs = {
    * - {@link BroadcastingErrorReason.RATE_LIMITED} - the profile reached the rate limit
    * - {@link BroadcastingErrorReason.APP_NOT_ALLOWED} - the app is not whitelisted for gasless transactions
    *
-   * @defaultValue true, the request will be attempted to be sponsored by the Lens API.
+   * If not specified, or `true`, the hook will attempt a Signless Experience when possible;
+   * otherwise, it will fall back to a signed experience.
    */
-  sponsored?: boolean;
+  sponsored?: false;
```
krzysu, line 40: is `false` here intentional?

packages/react/src/transactions/useFollow.ts
cesarenaldi: It is not needed when following, the information I was looking for (i.e. `sponsoredApproved` and `signlessApproved`) is repeated in the `UnknownFollowModuleSettings` for convenience.

On the other hand it's going to be needed when configuring a Profile's Follow Module and when creating a new publication.

packages/react/src/transactions/useFollow.ts
cesarenaldi: no more, cleared some of this in the following PR.

packages/react/src/transactions/useBlockProfiles.ts
cesarenaldi: no, good catch.

krzysu (approved):

PR #816, opened 2024-01-30
feat: add useReportProfile hook

krzysu:


packages/client/src/gated/LensClient.ts
```diff
@@ -30,7 +30,7 @@ export type LensClientConfig = base.LensClientConfig & {
 /**
  * Lens Protocol Client with token-gated content support.
  *
- * It provides access to all the base {@link Base.LensClient} modules and the {@link Gated} module.
+ * It provides access to all the core {@link Core.LensClient} modules and the {@link Gated} module.
```
krzysu, line 33: fixing some typedoc linking issues

packages/domain/src/use-cases/transactions/SupportedTransactionRequest.ts
```diff
@@ -1,6 +1,6 @@
-import { JustProtocolRequest } from '../../entities';
-import { ClaimHandleRequest } from '../profile';
+import { JustProtocolRequest } from '../../entities/Transactions';
 import { BlockProfilesRequest } from '../profile/BlockProfiles';
+import { ClaimHandleRequest } from '../profile/ClaimHandle';
```
krzysu, line 3: fixing circular deps after adding report profile

packages/domain/src/entities/PublicationReport.ts
```diff
@@ -1,6 +1,4 @@
-// No entity was required so far for this use case.
-
-export enum ReportReason {
+export enum PublicationReportReason {
```
cesarenaldi, line 1: We need to be careful with this change and provide an alias for backwards compatibility. At least in the immediate.

krzysu: it is here https://github.com/lens-protocol/lens-sdk/pull/816/files#diff-0ce407683ae3d6eabed3d60f01c2638a14a06f03e80931119ee70845bcd8def1R9-R16

packages/domain/src/use-cases/transactions/SupportedTransactionRequest.ts
```diff
@@ -1,6 +1,6 @@
-import { JustProtocolRequest } from '../../entities';
-import { ClaimHandleRequest } from '../profile';
+import { JustProtocolRequest } from '../../entities/Transactions';
```
cesarenaldi, line 1: Is this also for circular dep? There shouldn't be

krzysu: you are right, reverting this change

PR #817, opened 2024-01-31
feat: resolveReference policy helper

cesarenaldi:


packages/api-bindings/src/lens/FollowModule.ts
```diff
@@ -17,6 +13,46 @@ export type ResolveFollowPolicy = {
...
+  verified: boolean;
+};
+
+export type OpenFollowPolicy = {
+  type: FollowPolicyType.ANYONE;
+};
+
+export type FollowPolicy =
+  | ChargeFollowPolicy
+  | NoFeeFollowPolicy
```
krzysu, line 52: it should be like this, no?
```suggestion
  | NoFollowPolicy
```

krzysu (approved):

PR #836, opened 2024-02-09
client: introduce common query params

krzysu:


packages/client/src/helpers/commonQueryVariables.ts
```diff
@@ -0,0 +1,41 @@
...
+export function commonQueryVariables({ params }: LensContext): CommonQueryVariables {
+  return {
+    publicationImageTransform: params.image?.medium,
+    publicationOperationsActedArgs: {},
+    publicationStatsInput: { metadata: { publishedOn: params.statsFor } },
+    publicationStatsCountOpenActionArgs: {},
+    profileCoverTransform: params.profile?.cover,
+    profilePictureTransform: params.profile?.thumbnail,
+    profileStatsArg: { forApps: params.statsFor },
+    profileStatsCountOpenActionArgs: {},
```
krzysu, line 35: some of them are still not possible to configure but at least all are defined in this single place

cesarenaldi (approved):

PR #845, opened 2024-02-14
react: allow to disable sponsored tx on the config level

krzysu:


packages/domain/src/use-cases/transactions/SponsorshipReady.ts
```diff
@@ -4,9 +4,6 @@ type WithSponsoredFlag<T extends ProtocolTransactionRequestModel> = T & {
   sponsored: boolean;
 };
 
-export type SubsidizableProtocolTransactionRequestModel =
-  WithSponsoredFlag<ProtocolTransactionRequestModel>;
```
krzysu, line 8: it was not used anywhere

packages/react/src/transactions/publications/useCreateMirror.ts
```diff
@@ -240,11 +242,13 @@ export function useCreateMirror(): UseDeferredTask<
...
-      sponsored: args.sponsored ?? true,
-      ...args,
-    });
+    const request = configureRequest(
+      createMirrorRequest({
+        signless: session.profile.signless,
+        sponsored: args.sponsored ?? true,
+        ...args,
+      }),
+    );
```
cesarenaldi, line 251: Given we do have places where we do:
```
const { config } = useSharedDependencies();
```
wondering if it's more straightforward to also retrieve `config` in these hooks and make the fallback logic apparent here. Rather than creating an request object and rely on `configureRequest` to override it.

packages/react/src/transactions/publications/useCreatePostRequest.ts
```diff
@@ -16,8 +18,16 @@ function extractUnknownOpenActionModuleAddresses(desired: CreatePostRequest): Ev
...
 export function useCreatePostRequest() {
   const { execute: fetch } = useLazyModuleMetadata();
+  const { config } = useSharedDependencies();
+
   return async (input: UnknownObject): Promise<CreatePostRequest> => {
     const desired = createPostRequest(input);
+
+    // if sponsored is disabled globally, return here
+    if (config.sponsored === false) {
+      return overwriteSponsored(desired, config);
```
cesarenaldi, line 28: I would make the decision explicit here, not sure if `overwriteSponsored` level of indirection is needed.

packages/react/src/transactions/publications/useCreateMirror.ts
krzysu: in some places, with simple logic, it could be apparent, but in complicated cases where the sponsored flag is set based on multiple conditions, adding one extra condition to the mix will only make it less obvious, and more error prone, than just overwriting it at the end, if needed at all (if defined in the config).

krzysu: I added overwrites to all hooks that would require it, none uses `useSharedDependencies` beside `useCreatePostRequest` that has the most complicated logic and external calls to build the Request.

cesarenaldi: Fair enough

cesarenaldi (approved):

PR #847, opened 2024-02-14
fix: allows to define Origin header from RN integrations

cesarenaldi:


packages/api-bindings/src/apollo/links/AuthLink.ts
```diff
@@ -126,7 +131,8 @@ export function createAuthLink(accessTokenStorage: IAccessTokenStorage) {
         // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
         headers: {
           ...('headers' in prevContext && prevContext.headers),
-          authorization: `Bearer ${token}`,
+          Authorization: `Bearer ${token}`,
+          Origin: origin,
```
krzysu, line 135: what if it's not defined? maybe something like

```
...origin && { Origin: origin }
```

PR #857, opened 2024-02-19
fix: pagination issue with several React hooks

cesarenaldi:
What to review in this PR:
- ApolloClient upgrade
- `config.ts` changes
- one non-paginated hook: `usePublication`
- one lazy non-paginated hook: `useLazyProfile`
- one paginated hook: `useProfileFollowers`
- one lazy list hook: `useLazyPublications`
- adds missing `useCurrencies` paginated field policy
- some of the "odd ones": `useInvitedProfiles`, `useRevenueFromFollow` (just `$fxRateFor` var used)
- removed unused queries in `modules.graphql`
- removed code from previous variable injection approach (start from `createTypePolicies.ts`)


packages/api-bindings/package.json
```diff
@@ -74,7 +74,7 @@
     "typescript": "5.2.2"
   },
   "peerDependencies": {
-    "@apollo/client": "~3.8.7",
+    "@apollo/client": "^3.8.10",
```
krzysu, line 77: is this version number correct?

packages/react/src/shared.tsx
```diff
@@ -79,7 +72,7 @@ export function createSharedDependencies(userConfig: BaseConfig): SharedDependen
 
   // apollo client
   const apolloClient = createLensApolloClient({
-    queryParams: config.params,
+    connectToDevTools: config.debug,
```
krzysu, line 75: nice!

krzysu (approved):
reviewed, tested, LGTM 👍

packages/api-bindings/package.json
cesarenaldi: Not affecting the final result, but prefer to fix it. Good catch.

PR #860, opened 2024-02-21
fix: export URI/URL, add type policy for metadata

krzysu:


packages/react/src/publication/infrastructure/PublicationCacheManager.ts
```diff
@@ -75,6 +75,7 @@ export class PublicationCacheManager implements IPublicationCacheManager {
           },
           ...this.variables,
         },
+        returnPartialData: true,
```
krzysu, line 78: looks like this is not needed anymore, @cesarenaldi your other PR (https://github.com/lens-protocol/lens-sdk/pull/857) solved the issue.
but shall we keep returnPartialData here? in case of cache manager maybe it's not a problem to return partial data? makes it less likely to return nothing

packages/api-bindings/src/apollo/cache/type-policies/createPrimaryPublicationTypePolicy.ts
```diff
@@ -4,6 +4,7 @@ export function createPrimaryPublicationTypePolicy() {
   return {
     fields: {
       stats: oneToOneRelationship(),
+      metadata: oneToOneRelationship(),
```
krzysu, line 7: not sure if this is required

PR #861, opened 2024-02-21
chore: remove all what was deprecated

krzysu:


packages/api-bindings/codegen-api.yml
```diff
@@ -56,7 +56,7 @@ generates:
         TxId: string
         UnixTimestamp: number
         URI: string
-        URL: Url
+        URL: string
```
cesarenaldi, line 59: Should `URI` and `URL` now be mapped to the `URI` and `URL` branded types in shared-kernel?

krzysu: right, will do!

krzysu: done

packages/react/src/index.ts
```diff
@@ -57,7 +57,6 @@ export type {
   PromiseResult,
   Result,
   Success,
-  Url,
```
cesarenaldi, line 60: Verify if export `URI` and `URL` is needed here or already done elsewhere.

packages/react/src/transactions/adapters/schemas/common.ts
```diff
@@ -13,22 +13,20 @@ import {
...
 
 export const EvmAddressSchema: z.ZodType<EvmAddress, z.ZodTypeDef, string> = z
   .string()
   .min(42)
   .transform((value) => value);
 
-export const DataSchema: z.ZodType<Data, z.ZodTypeDef, string> = z
-  .string()
-  .transform((value) => value as Data);
+export const DataSchema: z.ZodType<Data, z.ZodTypeDef, string> = z.string().transform(data);
```
cesarenaldi, line 29: 👍

cesarenaldi (approved):

packages/react/src/index.ts
krzysu: they are now only used internally in our packages, don't need to export them at all

PR #867, opened 2024-02-27
feat: useRecommendProfileToggle hook

krzysu:


packages/react/src/profile/infrastructure/ProfileCacheManager.ts
```diff
@@ -49,6 +50,33 @@ export class ProfileCacheManager implements IProfileCacheManager {
     });
   }
 
+  update(
+    profileId: ProfileId,
+    updateFn: <TProfile extends Profile>(current: TProfile) => TProfile,
+  ) {
+    this.client.cache.updateQuery<ProfileData, ProfileVariables>(
```
krzysu, line 57: @cesarenaldi I had to introduce cache update in ProfileCacheManager to optimistically flip the new peerToPeerRecommendedByMe flag, but it doesn't work (again) :(
the apollo error says "Missing field 'result' while writing result {}" what could it be?

packages/domain/src/use-cases/publications/ToggleProperty.ts
```diff
@@ -1,43 +1,41 @@
-import { PublicationId } from '../../entities';
-
-export type TogglePropertyRequest = {
-  publicationId: PublicationId;
+export type TogglePropertyRequest<I> = {
+  id: I;
```
cesarenaldi, line 2: If you make the `id: string` you won't need `I` generic, which simplifies a lot of changes in this PR.

krzysu: done

packages/react/src/profile/infrastructure/ProfileCacheManager.ts
krzysu: replaced with `updateFragment`

PR #868, opened 2024-02-29
react: add useProfilePrice hook

krzysu:


packages/react/src/misc/adapters/useProfilePriceController.ts
```diff
@@ -0,0 +1,37 @@
...
+  PromiseResult,
+  assertError,
+  failure,
+  matic,
+  success,
+} from '@lens-protocol/shared-kernel';
+
+import { useSharedDependencies } from '../../shared';
+
+export function useProfilePriceController() {
```
krzysu, line 17: I decided to keep all logic in the controller without splitting into gateway/presenter, wdyt?

packages/react/src/misc/adapters/useProfilePriceController.ts
```diff
@@ -0,0 +1,37 @@
...
+      const provider = await providerFactory.createProvider({ chainType: ChainType.POLYGON });
+      const contract = permissionlessCreator(
+        config.environment.contracts.permissionlessCreator,
+        provider,
+      );
+
+      const priceBN = await contract.getProfileWithHandleCreationPrice();
+      const result = formatUnits(priceBN, matic().decimals);
+
+      return success(Amount.matic(result));
```
krzysu, line 31: I have some experience recently with using Amount and different types of assets and I don't find it easy to work with, as you can see in the example above.

packages/gated-content/package.json
```diff
@@ -116,12 +116,5 @@
       "web/index.ts"
     ],
     "exports": true
-  },
-  "pnpm": {
```
krzysu, line 120: pnpm was complaining about this, it should be defined at the root of the monorepo

cesarenaldi: I think the intent was to not have this warning screaming at people using `pnpm` to install these packages.

cesarenaldi: Actually I don't know if this is effective if done on what, from the consumer perspective, is a dependency `package.json`

packages/react/src/misc/adapters/useProfilePriceController.ts
```diff
@@ -0,0 +1,23 @@
+import { UnspecifiedError } from '@lens-protocol/api-bindings';
+import { GetProfilePrice, ProfilePrices } from '@lens-protocol/domain/use-cases/profile';
+import { PromiseResult } from '@lens-protocol/shared-kernel';
+
+import { useSharedDependencies } from '../../shared';
+import { PromiseResultPresenter } from '../../transactions/adapters/PromiseResultPresenter';
+import { ProfilePriceGateway } from './ProfilePriceGateway';
+
+export type ProfilePriceResult = ProfilePrices;
```
cesarenaldi, line 9: Do we need to alias it? The `Result` suffix could lead to confusion given the term is widely used in the context of `Result<T, E>`.

packages/react/src/misc/useProfilePrice.ts
```diff
@@ -0,0 +1,46 @@
...
+        setError(result.error);
+        return;
+      }
+
+      setPrices(result.value);
+    };
+
+    void fetchData();
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
cesarenaldi, line 43: Should we just make it lazy at this point? `data` comes back as `ProfilePriceResult` as well as part of the lazy callback?

packages/react/src/misc/adapters/useProfilePriceController.ts
cesarenaldi: `ProfilePrices` seems good to go out in the wild, it's DTO anyway.

packages/react/src/misc/useProfilePrice.ts
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * @example
+ * ```tsx
+ * const { data, error, loading } = useProfilePrice();
+ * ```
+ *
+ * @category Misc
+ * @group Hooks
+ */
+export function useProfilePrice(): ReadResult<ProfilePriceResult> {
```
cesarenaldi, line 24: Should we consolidate the use of plural/singular? Sometimes we say `ProfilePrices`, here it's singular.

krzysu: I made it plural as we expect to add more currencies soon

packages/gated-content/package.json
krzysu: I removed this as pnpm complains, when running install with:

```
packages/gated-content                   |  WARN  The field "pnpm" was found in /lens-sdk/packages/gated-content/package.json. 
This will not take effect. You should configure "pnpm" at the root of the workspace instead.
```

packages/react/src/misc/useProfilePrice.ts
krzysu: made it lazy, added also error handling

examples/web/src/profiles/UseCreateProfile.tsx
```diff
@@ -1,7 +1,26 @@
...
 import { RequireConnectedWallet } from '../components/auth';
+import { formatAmount } from '../utils/formatAmount';
+
+function ProfilePrice() {
+  const { data, execute, loading, error } = useProfilePrices();
+
+  useEffect(() => {
+    void execute();
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
krzysu, line 14: the only problem is here, if `execute` is provided in useEffect deps, then the app is stuck in a never-ending loop.

cesarenaldi: Can we brainstorm some options here? Seems a flaw in the way `useDeferredTask` is implemented.

krzysu: even if not perfect, I'd merge this branch as an example of onchain read hook, and keep it low profile, wdyt?

cesarenaldi:
Parking this one for now.

PR #873, opened 2024-03-04
feat: create-next-app template with react-web sdk and connectkit

krzysu:
test with:
```
pnpm create next-app -e https://github.com/lens-protocol/lens-sdk/tree/nextjs-template/examples/lens-next-app
```


examples/lens-next-app/src/components/ConnectWalletButton.tsx
```diff
@@ -0,0 +1,12 @@
+import { ConnectKitButton } from "connectkit";
+import { Button } from "./Button";
+
+export const ConnectWalletButton = () => {
```
cesarenaldi, line 4: Let's keep a consistent style on these components definition. It seems NextJS docs favour Function Declarations in their examples over Arrow Function Expressions, I would stick to that style.

examples/lens-next-app/src/app/favicon.ico
cesarenaldi, line 1: nit: we could use lens icon.

examples/lens-next-app/src/components/ConnectWalletButton.tsx
krzysu: true, I missed it, fixed now!

examples/lens-next-app/src/app/favicon.ico
krzysu: done!

PR #879, opened 2024-03-12
react: resolveCollectPolicy and resolveFollowPolicy with conversion rate

krzysu:


packages/api-bindings/src/lens/utils/CollectModuleSettings.ts
```diff
@@ -52,3 +55,219 @@ export function findCollectModuleSettings(
...
+  fee: CollectFee;
+};
+
+export type CollectPolicy = FreeCollectPolicy | PaidCollectPolicy | MultirecipientCollectPolicy;
+
+export enum CollectPolicyType {
+  FREE_COLLECT = 'FREE_COLLECT',
+  PAID_COLLECT = 'PAID_COLLECT',
+  MULTIRECIPIENT_COLLECT = 'MULTIRECIPIENT_COLLECT',
+}
```
cesarenaldi, line 103: Should we use `OpenActionType` values?

krzysu: I simplified it so we don't need type at all

examples/web/src/discovery/UseExplorePublications.tsx
```diff
@@ -1,9 +1,46 @@
...
+            policy.fee.amount,
+            policy.fee.rate,
+          )})`}</div>
+          {policy.fee.referralFee > 0 && <div>{`Referral fee: ${policy.fee.referralFee}%`}</div>}
+          {!isMultirecipientCollectFee(policy.fee) ? (
+            <div>{`Recipient: ${policy.fee.recipient}`}</div>
+          ) : (
+            <div>{`Recipients: ${policy.fee.recipients.map((r) => r.recipient).join(', ')}`}</div>
+          )}
+        </>
```
krzysu, line 39: see how easy is to display all collect info - all that is important from the UI perspective

cesarenaldi (approved):

PR #882, opened 2024-03-18
fix: client profile metadata request

krzysu:


packages/client/src/submodules/feed/graphql/feed.graphql
```diff
@@ -41,7 +41,7 @@ query Feed(
   $profilePictureTransform: ImageTransform!
   $profileStatsArg: ProfileStatsArg
   $profileStatsCountOpenActionArgs: ProfileStatsCountOpenActionArgs
-  $profileMetadataSource: AppId = null
```
krzysu, line 44: the issue was defaulting to null, and API expects undefined, seems null was still processed as a valid AppId returning always fallback metadata.

krzysu: does it make sense?

krzysu: I tested it with the same scenario as reported, having a profile with metadata, updating it, and requesting the profile again right after the update is indexed, and the issue is fixed with the code change above.

cesarenaldi: I don't doubt it it's like you described it. Odd for the API to make a distinction between `null` and `undefined` as these are often interchangeable in GQL.

PR #886, opened 2024-03-20
chore: react - export missing types

krzysu:


packages/react/src/publication/useNotInterestedToggle.ts
```diff
@@ -37,7 +35,7 @@ export type NotInterestedOperation = UseDeferredTask<void, never, UseNotInterest
  * @category Publications
  * @group Hooks
  */
-export function useNotInterestedToggle(): NotInterestedOperation {
+export function useNotInterestedToggle(): UseDeferredTask<void, never, UseNotInterestedToggleArgs> {
```
cesarenaldi, line 38: No objection per se, just curious about the rationale for removing this alias.

packages/react/src/publication/index.ts
```diff
@@ -19,33 +19,80 @@ export * from './useWhoReactedToPublication';
...
+  ThreeDMetadataV3,
+  ThreeDMetadataV3Asset,
+  TransactionMetadataV3,
   UnknownOpenActionModuleSettings,
   UnknownReferenceModuleSettings,
+  Video,
+  VideoMetadataV3,
+  WhoActedOnPublicationWhere,
+  WhoReactedPublicationWhere,
 } from '@lens-protocol/api-bindings';
```
cesarenaldi, line 96: Good job!

packages/react/src/publication/useNotInterestedToggle.ts
krzysu: I just found them unnecessary, and not consistent with other hooks, it's a breaking change though

cesarenaldi: Understood. I wouldn't worry too much about a breaking change. I think it's very unlikely anybody referenced these types in code.

PR #906, opened 2024-04-09
chore: re-export ClientError as GraphQLClientError

UncleBill:
## Use case

```ts
import { GraphQLClientError } from '@lens-protocol/client'
try {
  await client.follow(...)
  await doSomeThingElse()
} catch (error) {
  if (error instanceof Error) {
     if (error instanceof GraphQLClientError) {
        // handle client error.
     }
     // handle other error.
  }
}
```

### DIFF
--- a/.changeset/olive-spiders-pull.md
+++ b/.changeset/olive-spiders-pull.md
@@ -1,0 +1,5 @@
+---
+"@lens-protocol/client": patch
+---
+
+**feat:** re-exported GraphQLClientError from graphql-request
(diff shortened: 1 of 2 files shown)


krzysu:
hey @UncleBill thanks for contributing! 

could you also add a changset? you do it like

```
$ pnpm changeset
```
pick @lens-protocol/client and patch version, write message sth like:

```
**feat:** re-exported GraphQLClientError from graphql-request
```

add the file to your PR, if the build is green we will merge it. 
you can also run `pnpm lint` to make sure there are no linter/prettier issues

swkatmask:
@krzysu Changeset added.

UncleBill:
@krzysu changeset added.

krzysu (approved):

PR #915, opened 2024-04-16
react: useAddProfileInterests and useRemoveProfileInterests

krzysu:


packages/react/src/profile/useAddProfileInterests.ts
```diff
@@ -0,0 +1,61 @@
...
+  const { add } = useProfileInterestsController();
+
+  return useDeferredTask(async (request) => {
+    invariant(
+      session?.type === SessionType.WithProfile,
+      'You must be authenticated with a profile to use this hook. Use `useLogin` hook to authenticate.',
+    );
+
+    await add({
+      profileId: session.profile.id,
```
cesarenaldi, line 55: Can one add interest to another profile other than the logged-in one?

krzysu: no, but it is used in the presenter for optimistic update. also I didn't want to bring interests enum to the domain so I used only profileId in the domain request.
profileId is used only internally here, not exposed to the SDK user.

PR #927, opened 2024-04-29
fix: react native origin header

krzysu:


packages/api-bindings/src/apollo/index.ts
```diff
@@ -39,14 +38,20 @@ export function createLensApolloClient({
...
 
 export type AuthApolloClientConfig = {
-  uri: string;
   logger: ILogger;
+  origin?: string;
+  uri: string;
 };
 
-export function createAuthApolloClient({ uri, logger }: AuthApolloClientConfig) {
+export function createAuthApolloClient({ logger, origin, uri }: AuthApolloClientConfig) {
```
cesarenaldi, line 46: Why only for this ApolloClient and not the authenticated one?

krzysu: I thought the origin overwrite is used only in the auth challenge and is not used later. or is it?
is the API doing anything else based on origin?

cesarenaldi (approved):

PR #930, opened 2024-05-02
feat: suspense support for usePublication hook

krzysu:


packages/react/src/publication/usePublication.ts
```diff
@@ -26,66 +49,81 @@ export type UsePublicationArgs = OneOf<PublicationRequest>;
...
+  });
 
-  if (error) {
-    return {
-      data: undefined,
-      error,
-      loading: false,
-    };
-  }
+  // if (request.forId && !isValidPublicationId(request.forId)) {
```
krzysu, line 114: there is a problem, that with wrong publicationId the API returns error, but it's not handled anywhere so it's thrown in the example app, would it make sense to do simple publicationId validation, the same as on the API side? the other thing is, it is not possible before useSuspendableQuery hook and later is too late as the API error is already thrown. any idea?

krzysu: or that use case shouldn't really exist? I made this UI and it breaks badly when using a wrong publication Id https://lens-sdk-example-web-wagmi-q0w3lfali-avaraxyz.vercel.app/publications/usePublication

cesarenaldi: to me this is a case of wrong usage, a wrong Id (from a format perspective) to me should throw.

cesarenaldi: I rebased my branch on top of this one. I polished this aspect there. Will merge both now.

cesarenaldi (approved):

PR #935, opened 2024-05-06
feat: support Shared Revenue collect module

cesarenaldi:


packages/api-bindings/src/lens/utils/CollectModuleSettings.ts
```diff
@@ -141,64 +190,89 @@ export function resolveCollectPolicy(collectable: PrimaryPublication): CollectPo
...
+          endsAt: module.endsAt,
+          fee: fee,
+        };
+      }
+      return {
+        ...shared,
+        collectNft: module.collectNft,
+        collectLimit: module.collectLimit,
+        endsAt: module.endsAt,
+        creatorClient: module.creatorClient ?? undefined,
```
krzysu, line 274: so creatorClient is defined only if no fee?

packages/api-bindings/src/lens/utils/CollectModuleSettings.ts
```diff
@@ -103,10 +132,30 @@ function isMultirecipientCollectModuleSettings(
...
+function buildCollectFee(
+  module: gql.LegacyAaveFeeCollectModuleSettings | gql.LegacyErc4626FeeCollectModuleSettings,
+): CollectFee;
+function buildCollectFee(
+  module:
+    | gql.LegacyLimitedFeeCollectModuleSettings
+    | gql.LegacyLimitedTimedFeeCollectModuleSettings,
+): CollectFee;
+function buildCollectFee(
+  module: Extract<CollectModuleSettings, { amount: gql.Amount }>,
```
krzysu, line 157: does it read better than `module: CollectModuleSettings` ? or is the type safety better this way?

packages/api-bindings/src/lens/utils/CollectModuleSettings.ts
cesarenaldi: correct

packages/api-bindings/src/lens/utils/CollectModuleSettings.ts
cesarenaldi: I changed it so that works only with modules that can have a fee.

This was more type safe at the call site.

I also provided the overloads so that call site is even more accurate as I experienced a typo while I was working and wasn't flagging the call.

Finally the implementation signature just supports all possible overrides.

PR #961, opened 2024-07-10
feat: Privy next.js template

iPaulPro:
Adds a new example `lens-next-privy-app`, forked from `lens-next-app`, which replaces ConnectKit with Privy for wallet authentication.

It only requires supplying the Privy App ID, in place of the Wallet Connect Project ID used in the original example, to run.

Creating a draft because we're waiting on a fix from Privy in the `@privy-io/react-auth` library. I will also update the readme with a link to the Privy docs for Lens integration when it's published.


iPaulPro:
Tested @privy-io/react-auth v1.74.1 and it no longer has the signing prompt issue. This is ready for review now!

pnpm-lock.yaml
```diff
@@ -5,7 +5,6 @@ settings:
   excludeLinksFromLockfile: false
 
 overrides:
-  ethereumjs-abi: https://registry.npmjs.org/ethereumjs-abi/-/ethereumjs-abi-0.6.8.tgz
```
cesarenaldi, line 8: @iPaulPro do you recall any change that could have removed this? It's still present in the `package.json` so unclear to me why it gets removed from the lock file.

iPaulPro: The lock file was very out of date and this is just removing an override, not the library. If you run `pnpm install` on the current `develop` branch the same change is made.

PR #983, opened 2024-11-29
Implement follow/unfollow

juangm:


packages/client/src/actions/follow.ts
```diff
@@ -0,0 +1,51 @@
...
+
+import type { SessionClient } from '../clients';
+import type { UnauthenticatedError, UnexpectedError } from '../errors';
+
+/**
+ * Follow an Account
+ *
+ * ```ts
+ * const result = await follow(sessionClient,
+ *  { account: EvmAddress('0x90c8c68d0Abfb40D4fCD72316A65e42161520BC3') }
```
cesarenaldi, line 18:
```suggestion
 *  { account: evmAddress('0x90c8c68d0Abfb40D4fCD72316A65e42161520BC3') }
```

it's an helper function.

cesarenaldi: Formatting it's a bit odd, is your IDE formatting these for you?

cesarenaldi (approved):

PR #996, opened 2024-12-04
Implement follow queries

juangm:


packages/client/src/actions/follow.ts
```diff
@@ -49,3 +64,85 @@ export function unfollow(
...
+ *   target: evmAddress('0xe5439696f4057aF073c0FB2dc6e5e755392922e1'),
+ * });
+ * ```
+ *
+ * @param client - The session client for the authenticated Account.
+ * @param request - The query request.
+ * @returns List of accounts following.
+ */
+export function fetchFollowersYouKnow(
+  client: SessionClient,
```
cesarenaldi, line 123: Is it necessary to be authenticated?

cesarenaldi (approved):
I left a question, but not sure. Address it if need to and then feel free to merge.

PR #997, opened 2024-12-04
Implement username/usernames queries

juangm:


packages/graphql/src/fragments/namespace.ts
```diff
@@ -0,0 +1,27 @@
...
+      address
+      namespace
+      createdAt
+      metadata {
+        __typename
+        description
+        id
+      }
+      owner
+      # TODO: Implement rules
```
cesarenaldi, line 16: I don't think `gql.tada` likes GQL comments, does it?

cesarenaldi (commented):
There is a failing test that seems something new.

packages/graphql/src/fragments/namespace.ts
cesarenaldi: over on the types generated from it, if you get `never` it failed.

packages/client/src/actions/timeline.ts
```diff
@@ -32,7 +32,7 @@ export function fetchTimeline(
 }
 
 /**
- * Fetch fetchTimelineHighlights from an account.
+ * Fetch Timeline Highlights from an account.
```
cesarenaldi, line 35:
```suggestion
 * Fetch Timeline Highlights for an account.
```

packages/client/src/actions/username.ts
```diff
@@ -0,0 +1,47 @@
...
+ */
+export function fetchUsername(
+  client: AnyClient,
+  request: UsernameRequest,
+): ResultAsync<Username | null, UnexpectedError> {
+  return client.query(UsernameQuery, { request });
+}
+
+/**
+ * Fetch usernames owned by an account.
```
cesarenaldi, line 30:
```suggestion
 * Fetch usernames owned by an address.
```

cesarenaldi (approved):

PR #1000, opened 2024-12-05
Implement notifications query

juangm:


packages/client/src/actions/health.ts
```diff
@@ -4,8 +4,7 @@ import type { AnyClient } from '../clients';
 import type { UnexpectedError } from '../errors';
 
 /**
- * Health checks.
- *
+ * Health check endpoint.
```
cesarenaldi, line 7: nit: not an 'endpoint', it's a query. `/graphql` is an endpoint.

cesarenaldi (approved):

PR #1002, opened 2024-12-06
Implement missing mutations for account

juangm:


packages/client/src/actions/account.ts
```diff
@@ -298,3 +310,102 @@ export function unmuteAccount(
...
   return client.mutation(UnmuteAccountMutation, { request });
 }
+
+/**
+ * Report an account.
+ *
+ * ```ts
+ * const result = await reportAccount(sessionClient, {
+ *   account: evmAddress("0xe5439696f4057aF073c0FB2dc6e5e755392922e1"),
+ *   reason: "REPETITIVE_SPAM",
```
cesarenaldi, line 320: We should create an enum in graphql package for these (see `PageSize` and the way it's bound to scalars).

cesarenaldi (approved):

PR #1008, opened 2024-12-10
feat: introduces accountFragment in client config

cesarenaldi:


packages/client/src/clients.ts
```diff
@@ -469,4 +464,15 @@ export type { SessionClient };
 /**
  * Any client that can be used to interact with the Lens GraphQL API.
  */
-export type AnyClient = PublicClient | SessionClient;
+// TODO remove default
```
cesarenaldi, line 467: To be decided once we have rolled out the solution to all relevant queries. It might stay.

PR #1035, opened 2024-12-18
Upgrade index.ts

Cherkaso8:
Added error processing in case the user cancels the request to connect the wallet:

Announce "let address: Address;" only once in front of the try block to avoid conflict


cesarenaldi:
@Cherkaso8 thank you for taking the time. But the examples serve the purpose of explaining Lens SDK specific features, not be 100% accurate on 3rd party aspects. There's many things we could have done to make the example bombproof (e.g., switch to network, add network, etc.) but it would add unnecessary noise and defeated the purpose of explaining something.

Going to close this PR.

PR #1060, opened 2025-01-21
Continue adding fetch methods to the SDK

juangm:


packages/react/src/account/useAccountManagers.ts
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * Fetch Account Managers.
+ *
+ * ```tsx
+ * const { data, loading } = useAccountManagers();
+ * ```
+ */
+export function useAccountManagers(
+  args: AccountManagersArgs,
```
cesarenaldi, line 30: I didn't check this in detail, but wherever the request is optional, and from the example `useAccountManagers()` it seems so, this param should optional.

Otherwise this will only work with an awkward empty object `useAccountManagers({})` if no specific args are needed.

packages/react/src/account/useAccountManagers.ts
```diff
@@ -0,0 +1,42 @@
...
+): ReadResult<Paginated<AccountManager>>;
+
+export function useAccountManagers({
+  suspense = false,
+  ...request
+}: AccountManagersArgs & { suspense?: boolean }): SuspendableResult<Paginated<AccountManager>> {
```
cesarenaldi, line 36: I believe in the case of optional args for non-suspense signature we need to also provide a default value here:

```suggestion
}: AccountManagersArgs & { suspense?: boolean } = {}): SuspendableResult<Paginated<AccountManager>> {
```

This is not visible from the outside, it's just needed to TS to support all signatures.

cesarenaldi (approved):

PR #1085, opened 2025-02-21
feat: frames 1.1

defispartan:


packages/graphql/src/enums.ts
```diff
@@ -1137,3 +1137,17 @@ export enum SimpleCollectValidationFailedReason {
   NotEnoughBalance = 'NOT_ENOUGH_BALANCE',
   NotAFollower = 'NOT_A_FOLLOWER',
 }
+
+export enum FrameVerifySignatureResult {
```
cesarenaldi, line 1141: Can you please bind this new enum to the corresponding type in [here](https://github.com/lens-protocol/lens-sdk/blob/next/packages/graphql/src/graphql.ts#L111-L219)? This results in GQL inferred types that reference this enum.

packages/client/src/actions/frames.ts
cesarenaldi, line 1: The TSDocs examples have the Lens SDK v2 signature and jargon (e.g. 'LensClient'). Can we please update them according to the new implementation?

packages/graphql/src/enums.ts
defispartan: Added the binding

After I added it though [this line](https://github.com/defispartan/lens-sdk/blob/37e5f1e63c673003f5bcf128b488f6ad9bfb2795/packages/client/src/actions/frames.ts#L101) now returns a type error

Is there an example of another API endpoint that directly returns an enum? I tried searching to see how the types were implemented in the SDK but I couldn't find one

packages/client/src/actions/frames.ts
defispartan: Updated

packages/client/src/actions/frames.ts
```diff
@@ -0,0 +1,102 @@
...
+/**
+ * Create Frame action typed data to be signed by user wallet
+ *
+ * @param request - The request object
+ * @returns Typed data for Frame request
+ * @experimental This function might change in the future release
+ *
+ * @example
+ * ```ts
+ * const result = await createFrameTypedData(client, {
```
cesarenaldi, line 27: We use this naming convention to re-inforce in the example what it can be used.
```suggestion
 * const result = await createFrameTypedData(anyClient, {
```

packages/client/src/actions/frames.ts
```diff
@@ -0,0 +1,102 @@
...
+ *
+ * ⚠️ Requires authenticated LensClient.
+ *
+ * @param request - The request object
+ * @returns Signature result
+ * @experimental This function might change in the future release
+ *
+ * @example
+ * ```ts
+ * const result = await signFrameAction(client, {
```
cesarenaldi, line 59:
```suggestion
 * const result = await signFrameAction(sessionClient, {
```
Similarly, just reinforcing what's needed.

packages/client/src/actions/frames.ts
```diff
@@ -0,0 +1,102 @@
...
+/**
+ * Verify Frame signature
+ *
+ * @param request - The request object
+ * @returns Verification result
+ * @experimental This function might change in the future release
+ *
+ * @example
+ * ```ts
+ * const result = await verifyFrameSignature(client, {
```
cesarenaldi, line 88:
```suggestion
 * const result = await verifyFrameSignature(anyClient, {
```

packages/client/src/actions/frames.ts
```diff
@@ -0,0 +1,102 @@
...
+  client: AnyClient,
+  request: CreateFrameTypedDataRequest,
+): ResultAsync<CreateFrameEip712TypedDataFragment, UnexpectedError> {
+  return client.query(CreateFrameTypedDataQuery, { request });
+}
+
+/**
+ * Sign Frame action with Lens Manager if enabled
+ *
+ * ⚠️ Requires authenticated LensClient.
```
cesarenaldi, line 51:
```suggestion
 * ⚠️ Requires authenticated SessionClient.
```

packages/graphql/src/enums.ts
```diff
@@ -1137,3 +1137,17 @@ export enum SimpleCollectValidationFailedReason {
...
+  postDoesntExist = 'POST_DOESNT_EXIST',
+  identityUnauthorized = 'IDENTITY_UNAUTHORIZED',
+  identityTokenNotValid = 'IDENTITY_TOKEN_NOT_VALID',
+  typedDataTypesIncorrectFields = 'TYPED_DATA_TYPES_INCORRECT_FIELDS',
+  typedDataDomainIncorrect = 'TYPED_DATA_DOMAIN_INCORRECT',
+  typedDataAccountNotMatchingIdentityToken = 'TYPED_DATA_ACCOUNT_NOT_MATCHING_IDENTITY_TOKEN',
+  identityCannotUseAccount = 'IDENTITY_CANNOT_USE_ACCOUNT',
+  signatureNotValid = 'SIGNATURE_NOT_VALID',
+  deadlineExpired = 'DEADLINE_EXPIRED',
+  signerAddressCannotUseAccount = 'SIGNER_ADDRESS_CANNOT_USE_ACCOUNT',
```
cesarenaldi, line 1152: Can we please PascalCase these?

packages/graphql/src/enums.ts
cesarenaldi: Returning an enum should not be different than returning a scalar.

The closest thing I can think of is `Query. appServerApiKey` which returns a string.

Since you told the type system that the `FrameVerifySignatureResult` binds to the provided enum, you should not need to do lookup (with type assertion) at all.

```diff
export function verifyFrameSignature(
  client: AnyClient,
  request: VerifyFrameSignatureRequest,
): ResultAsync<FrameVerifySignatureResult, UnexpectedError> {
  return client
    .query(VerifyFrameSignatureQuery, { request })
-   .map((result) => FrameVerifySignatureResult[result as keyof typeof FrameVerifySignatureResult]);
}
```

cesarenaldi (changes requested):

packages/graphql/src/enums.ts
defispartan: That worked, thanks!

packages/graphql/src/enums.ts
defispartan: Updated

cesarenaldi (approved):

PR #1111, opened 2025-03-19
Add support for unknownPostMetadata

juangm:


zannis (approved):

packages/graphql/schema.graphql
```diff
@@ -3022,6 +3022,8 @@ type JoinGroupResponse {
 
 union JoinGroupResult = JoinGroupResponse | GroupOperationValidationFailed | SponsoredTransactionRequest | SelfFundedTransactionRequest | TransactionWillFail
 
+scalar JsonString
```
cesarenaldi, line 3025: Can we defined this under `packages/types/src/misc.ts` as a branded string, and map it in the types scalars?

PR #1116, opened 2025-03-25
Update schema and small fix

juangm:


packages/graphql/src/post.ts
```diff
@@ -370,7 +370,7 @@ export const PostReactionStatusFragment = graphql(
     postId
     account
     result
-    app
+    app @include(if: true)
```
cesarenaldi, line 373: Can you clarify what's the intent here?

packages/client/src/actions/account.ts
```diff
@@ -214,7 +214,7 @@ export function fetchAccountsAvailable(
  */
 export function fetchAccountsBlocked(
   client: SessionClient,
-  request: AccountsBlockedRequest,
+  request: AccountsBlockedRequest = {},
```
cesarenaldi, line 217: 👍

PR #1117, opened 2025-03-25
Update notification fragments

juangm:


packages/graphql/src/notifications.ts
```diff
@@ -1,12 +1,27 @@
...
+  SimpleCollectPostActionExecutedFragment,
+  TippingPostActionExecutedFragment,
+  UnknownPostActionExecutedFragment,
+} from './post';
 
 const FollowNotificationFragment = graphql(
   `fragment FollowNotification on FollowNotification {
     __typename
     id
     followers {
```
cesarenaldi, line 23: I don't know why we didn't but let's create a fragment for this.

packages/graphql/src/notifications.ts
```diff
@@ -22,10 +37,12 @@ const ReactionNotificationFragment = graphql(
     __typename
     id
     reactions {
```
cesarenaldi, line 39: same here, we need a fragment

packages/graphql/src/notifications.ts
```diff
@@ -22,10 +37,12 @@ const ReactionNotificationFragment = graphql(
...
+      __typename
       account {
         ...Account
       }
       reactions {
```
cesarenaldi, line 44: + fragment

packages/graphql/src/notifications.ts
```diff
@@ -101,8 +119,137 @@ const MentionNotificationFragment = graphql(
...
+>;
+
+export const PostActionExecutedNotificationFragment = graphql(
+  `fragment PostActionExecutedNotification on PostActionExecutedNotification {
+    __typename
+    id
+    post {
+      ...AnyPost
+    }
+    actions{
```
cesarenaldi, line 215: We need a fragment here

cesarenaldi (changes requested):

cesarenaldi:
can you please review the notifications, I didn't realize we are missing so many fragments.

packages/graphql/src/notifications.ts
juangm: Getting error when trying to do it.
```
The inferred type of this node exceeds the maximum length the compiler will serialize. An explicit type annotation is needed.ts(7056)
```

packages/graphql/src/notifications.ts
```diff
@@ -1,40 +1,74 @@
...
+
+const NotificationAccountPostReactionFragment = graphql(
+  `fragment NotificationAccountPostReaction on NotificationAccountPostReaction {
+    __typename
+    account {
         ...Account
       }
-      followedAt
+      reactions {
+        ...PostReaction
```
cesarenaldi, line 51: indentation is off.

cesarenaldi (approved):

PR #1121, opened 2025-03-27
Update schema for account creation with owner and new field in UsernameSpace fragment

juangm:


packages/graphql/src/fragments/primitives.ts
```diff
@@ -401,6 +402,25 @@ const UsernameNamespaceStatsFragment = graphql(
...
+    __typename
+    bannerImage
+    collaborators
+    description
+    externalLink
+    featuredImage
+    image
+    name
+    schema
+    signature
```
cesarenaldi, line 416: this is a mistake on the API side, can you please omit it from the fragment and raise a low priority API ticket for Zannis?

cesarenaldi (approved):

PR #1162, opened 2025-05-27
chore: update schema with native token support

juangm:


packages/graphql/src/fragments/post.ts
```diff
@@ -281,17 +297,26 @@ export type OperationValidationOutcome =
   | PostOperationValidationUnknown
   | PostOperationValidationFailed;
 
-export const PostTipFragment = graphql(
+export type PostTip = {
+  __typename: 'PostTip';
+  amount: Erc20Amount;
```
cesarenaldi, line 302: @juangm missing the deprecation warning.

desfero (approved):

PR #1172, opened 2025-06-11
React hooks for fetching data from ML

juangm:


packages/react/src/ml/useAccountRecommendations.ts
```diff
@@ -0,0 +1,44 @@
+import type { Account, AccountRecommendationsRequest, Paginated } from '@lens-protocol/graphql';
+import { MlAccountRecommendationsQuery } from '@lens-protocol/graphql';
+
+import type { ReadResult, Suspendable, SuspendableResult, SuspenseResult } from '../helpers';
+import { useSuspendableQuery } from '../helpers';
+
+export type AccountRecommendationsArgs = AccountRecommendationsRequest;
+
+/**
+ * Fetch account recommendations from ML.
```
cesarenaldi, line 10: We can do better with this descriptions.

PR #1177, opened 2025-06-24
feat: add hook for setAccountmetadata

juangm:


packages/react/src/account/useSetAccountMetadata.ts
```diff
@@ -0,0 +1,40 @@
...
+  | SigningError
+  | ValidationError
+  | TransactionIndexingError
+  | UnauthenticatedError
+  | UnexpectedError
+> {
+  return useAuthenticatedAsyncTask((sessionClient, request) =>
+    setAccountMetadata(sessionClient, request)
+      .andThen(args.handler)
+      .andThen(sessionClient.waitForTransaction),
```
cesarenaldi, line 38: It's pretty difficult to do any sort of optimistic updates since we just receive the metadata URI from the caller. But we could do an effort for updating the cache at the end. We can re-fetch the account with `fetchAccount`, in doing so the cache will update and any rendered hook (e.g. `useAccount`) returning the same Account should re-render.

Also, unless we have a use case, let's not return what we have (e.g. `TxHash`). Sometimes is better to return Void and potentially change it later, than return something we don't know if useful or not and then be stuck in not being able to change that something. Changing Void into something is not a breaking change, but changing `TxHash` into something else, is a breaking change.

PR #1179, opened 2025-06-26
Add useAccountStats, useAccountFeedStats and usePostTags hooks

juangm:


packages/client/src/actions/misc.ts
```diff
@@ -25,7 +25,13 @@ export function health(
 }
 
 /**
- * Fetch an Access Control details.
+ * Fetch an account's Access Control details.
```
cesarenaldi, line 28: this is not 'account'

packages/react/src/authentication/useAccessControl.ts
cesarenaldi, line 1: Not needed for now on social side. let's remove this

PR #1180, opened 2025-06-26
Add useCreateAccount and useCreateAccountWithUsername hooks

juangm:


packages/react/src/account/useCreateAccountWithUsername.ts
```diff
@@ -0,0 +1,49 @@
...
+  | UnauthenticatedError
+  | UnexpectedError
+> {
+  return useAuthenticatedAsyncTask((sessionClient, request) =>
+    createAccountWithUsername(sessionClient, request)
+      .andThen(args.handler)
+      .andThen(sessionClient.waitForTransaction)
+      .andThen((txHash) => fetchAccount(sessionClient, { txHash }))
+      .map(nonNullable)
+      .map(expectTypename('Account')),
```
cesarenaldi, line 47: Same here, this is not a union. Not needed.

packages/react/src/account/useCreateAccountWithUsername.ts
```diff
@@ -0,0 +1,49 @@
...
+import { expectTypename, nonNullable } from '@lens-protocol/types';
+
+import { type UseAsyncTask, useAuthenticatedAsyncTask } from '../helpers';
+
+export type UseCreateAccountWithUsernameArgs = {
+  handler: OperationHandler;
+};
+
+/**
+ * Creates an account on Lens.
```
cesarenaldi, line 26: In general, let's always use Account (upper cased A) when we refer to the Lens Account smart contract (like in docs).

PR #1183, opened 2025-06-27
Ass useAppUsers, useBalancesBulk and useCanCreateUsername hooks

juangm:


packages/react/src/account/useBalancesBulk.ts
```diff
@@ -0,0 +1,54 @@
...
+  Suspendable,
+  SuspendableResult,
+  SuspenseResult,
+} from '../helpers';
+import { useSuspendableQuery } from '../helpers';
+
+export type BalancesBulkArgs = BalancesBulkRequest;
+
+/**
+ * Fetch a finite number of balances for the specific address.
```
cesarenaldi, line 15:
```suggestion
 * Fetch a finite number of balances for the specified address.
```

packages/react/src/app/useAppUsers.ts
```diff
@@ -0,0 +1,56 @@
...
+export type AppUsersArgs = AppUsersRequest;
+
+/**
+ * Fetch list of users using an App.
+ *
+ * This signature supports React Suspense:
+ *
+ * ```tsx
+ * const { data } = useAppUsers({
+ *   app: evmAddress('0x123...'),
```
cesarenaldi, line 25: Replace `...` with proper symbol.

packages/react/src/username/useCanCreateUsername.ts
```diff
@@ -0,0 +1,45 @@
...
+          canCreateUsername: false,
+          reason: JSON.stringify(result.extraChecksRequired),
+        };
+      }
+      return {
+        canCreateUsername: false,
+        reason: result.reason,
+      };
+    }),
+  );
```
cesarenaldi, line 44: Remove `.map` as agreed in call.

PR #1184, opened 2025-06-27
Add useFollowStatus, useGroupBannedAccounts and useGroupMembershipRequests

juangm:


packages/react/src/follow/useFollowStatus.ts
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * ```tsx
+ * const { data } = useFollowStatus({
+ *   pairs: [
+ *     {
+ *       account: evmAddress('0x…'),
+ *       follower: evmAddress('0x…'),
+ *     }
+ *   ],
+ * }, suspense: true });
```
cesarenaldi, line 30: this syntax is broken.

cesarenaldi, line 30:
```suggestion
 *   suspense: true,
 * });
```

cesarenaldi (approved):

PR #1185, opened 2025-06-27
Add useWhoExecutedActionOnAccount, useWhoExecutedActionOnPost and useWhoReferencedPost

juangm:


packages/react/src/account/useWhoExecutedActionOnAccount.ts
```diff
@@ -0,0 +1,59 @@
...
+
+import type {
+  ReadResult,
+  Suspendable,
+  SuspendableResult,
+  SuspenseResult,
+} from '../helpers';
+import { useSuspendableQuery } from '../helpers';
+
+export type WhoExecutedActionOnAccountArgs = WhoExecutedActionOnAccountRequest;
```
cesarenaldi, line 16:
```suggestion
export type UseWhoExecutedActionOnAccountArgs = WhoExecutedActionOnAccountRequest;
```

PR #1188, opened 2025-07-07
chore: update schema and adds group.e2e.ts tests

cesarenaldi:


packages/client/src/actions/group.e2e.ts
```diff
@@ -0,0 +1,60 @@
...
+  name: 'web3natives',
+});
+
+describe('Given a logged-in user', () => {
+  describe('When joining a Group with SimplePaymentRule in native token denomination', () => {
+    let groupAddress: EvmAddress = '0x9904A9E96dfB83409Cd0D89DF15588eDBb99B322';
+
+    beforeAll(async () => {
+      const result = await loginAsBuilder().andThen((sessionClient) =>
+        createGroup(sessionClient, {
```
juangm, line 24: do we want to create a group each run of the tests? if so then we should remove the default address (test will fail anyway if the beforeAll throws error.

packages/client/src/actions/group.e2e.ts
```diff
@@ -0,0 +1,60 @@
...
+
+    it('Then the user can join the Group by fulfilling the payment rule with native tokens from the Lens Account', async () => {
+      const result = await loginAsAccountOwner().andThen((sessionClient) =>
+        joinGroup(sessionClient, {
+          group: groupAddress,
+        })
+          .andThen(handleOperationWith(wallet))
+          .andThen(sessionClient.waitForTransaction),
+      );
+      assertOk(result);
```
juangm, line 57: I would also check the user is part of the member of the group.

juangm (approved):

PR #1190, opened 2025-07-07
Invalidate cache when bookmark a post

juangm:


cesarenaldi (commented):
Let's add the optimistic cache update part in this same PR. The reset alone won't sort any effect.

PR #1195, opened 2025-07-09
fix: include schema into bundle

juangm:


cesarenaldi:
incorporated in: https://github.com/lens-protocol/lens-sdk/pull/1196

PR #1199, opened 2025-07-10
feat: add optimistic cache to follow/unfollow

juangm:


juangm:
Missing proper id on `LoggedInAccountOperations` from the backend.

packages/react/src/authentication/useMe.ts
cesarenaldi, line 1: Let's call this `useMeDetails` (akin to what we did with the action function). `useMe` sounds unclear.

packages/client/src/cache.ts
```diff
@@ -1,5 +1,14 @@
+// biome-ignore-all lint/suspicious/noExplicitAny: simplicity
```
cesarenaldi, line 1: let's avoid blanket turn off of linting features.

